PR #1, opened 2026-06-25
adr: initial v1 decision records (0000-0014)

yaad-rubigd:
First build step for Darbaan: the v1 design (locked 2026-06-25) split into Architecture Decision Records.

## What's here
15 ADRs under `adr/` (0000 meta + 0001–0014), plus an index README.

- 0000 record architecture decisions
- 0001 mail proxy over IMAP/SMTP, not an HTTP API
- 0002 policy in a separate box; credential isolation
- 0003 outbound sluice trap, fail-closed + default-deny
- 0004 pluggable, multi-level approval pipeline
- 0005 agent pre-screener as a compile-time plugin; risk routing
- 0006 rejection as an async DSN bounce
- 0007 signed-bounce trust
- 0008 inbound filter: declarative YAML rules
- 0009 pluggable backends with capability negotiation
- 0010 multi-mailbox v1; multi-agent deferred
- 0011 append-only audit log
- 0012 deployment + secrets at rest
- 0013 Go 1.24+, MIT
- 0014 prefer established libraries over reinventing

## Roles
admin: yaad-rubigd (this PR) · reviewer: sora-rubigd · coder: forge-rubigd.

@sora-rubigd — review pass please. Next up after merge: Go skeleton (go.mod, cmd/darbaan, internal/, pkg/) and the first batch of issues for @forge-rubigd.

### DIFF
--- a/adr/0002-policy-in-a-separate-box-credential-isolation.md
+++ b/adr/0002-policy-in-a-separate-box-credential-isolation.md
@@ -1,0 +1,19 @@
+# ADR 0002: Policy in a separate trusted box; credential isolation
+
+**Status:** Accepted (2026-06-25)
+
+## Context
+The thing an email injection attacks is the agent's own judgment. Policy that
+lives in the agent's prompt or behaviour can be argued away. The real mailbox
+credentials, if held by the agent, can be exfiltrated.
+
+## Decision
+Policy lives in Darbaan, a separate component the agent does not control, and is
+enforced **mechanically**. Darbaan holds the **real** mailbox credentials; each
+agent only ever gets **Darbaan** credentials.
+
+## Consequences
+- Policy holds even if an agent is fully compromised.
+- An injection that steals an agent's credentials gets a box that is already
+  policy-limited, never the actual mailbox.
+- Darbaan is a trust boundary; its own integrity is the thing to protect.
--- a/adr/0002-policy-in-a-separate-box-credential-isolation.md
+++ b/adr/0002-policy-in-a-separate-box-credential-isolation.md
@@ -17,3 +17,10 @@ agent only ever gets **Darbaan** credentials.
 - An injection that steals an agent's credentials gets a box that is already
   policy-limited, never the actual mailbox.
 - Darbaan is a trust boundary; its own integrity is the thing to protect.
+
+## Amendment (2026-06-25, review)
+**Defining "Darbaan credentials" (how an agent authenticates).** Each agent
+authenticates to Darbaan's IMAP and SMTP faces with **per-agent credentials** (a
+username + secret/token) issued by Darbaan, over TLS. These identify the agent
+principal and grant **only** Darbaan access, never upstream access. v1 is
+single-agent (one credential set); the mechanism generalizes to many (ADR 0010).
--- a/adr/0004-pluggable-approval-pipeline.md
+++ b/adr/0004-pluggable-approval-pipeline.md
@@ -20,3 +20,10 @@ agent approvers never edit.
 - New approval surfaces are added without touching the core.
 - The audit log keeps both the agent's original draft and any human-edited
   version (see ADR 0011).
+
+## Amendment (2026-06-25, review)
+**Approver timeout = fail-closed.** If no approver acts within a configured
+window, the message is **not** sent; it stays queued. A timeout never
+auto-sends, consistent with default-deny (ADR 0003). Operators may configure
+escalation or an explicit expiry-to-rejected (which then bounces per ADR 0006),
+but the default on timeout is "keep waiting," never "send."
--- a/adr/0004-pluggable-approval-pipeline.md
+++ b/adr/0004-pluggable-approval-pipeline.md
@@ -1,0 +1,22 @@
+# ADR 0004: Pluggable, multi-level approval pipeline
+
+**Status:** Accepted (2026-06-25)
+
+## Context
+Approval may come from a person via different surfaces, or from an automated
+check. We do not want the gate wired to a single mechanism.
+
+## Decision
+Approvers implement one contract: `(message + metadata) -> verdict`. A web UI, a
+Telegram bot, a chat react, or an agent are all implementations. Approvers are
+selectable at **runtime and compile time** (Go build tags, blank-import
+registry), so a binary can ship with only the approvers wanted. Chaining gives
+**multi-level** approval (every stage must pass). **Edit-before-release** is a
+per-approver capability: an approver that supports it (web) may edit a draft
+inline and release; others only approve/reject. Editing is **human-only**;
+agent approvers never edit.
+
+## Consequences
+- New approval surfaces are added without touching the core.
+- The audit log keeps both the agent's original draft and any human-edited
+  version (see ADR 0011).
--- a/adr/0005-agent-prescreener-plugin-risk-routing.md
+++ b/adr/0005-agent-prescreener-plugin-risk-routing.md
@@ -1,0 +1,19 @@
+# ADR 0005: Agent pre-screener as a compile-time plugin; risk routing
+
+**Status:** Accepted (2026-06-25)
+
+## Context
+An automated agent can pre-screen queued mail for injection/exfiltration smell
+before a human looks. But the core must not depend on an AI runtime.
+
+## Decision
+The agent pre-screener is **just another approver plugin**, selected via build
+tag (ADR 0004). Darbaan's core carries **no AI dependency** and can be built
+human-only. The pre-screener emits a **coarse** result: a level
+(low/medium/high), reasons, and boolean flags (touches-secret, new-recipient,
+external-domain, has-attachment). Routing: low with no flags takes the light
+path (one human tap); medium/high or any flag takes the strict path.
+
+## Consequences
+- Coarse buckets avoid false precision; refine once real traffic is seen.
+- Deployments choose their own AI dependency posture at build time.
--- a/adr/0005-agent-prescreener-plugin-risk-routing.md
+++ b/adr/0005-agent-prescreener-plugin-risk-routing.md
@@ -17,3 +17,11 @@ path (one human tap); medium/high or any flag takes the strict path.
 ## Consequences
 - Coarse buckets avoid false precision; refine once real traffic is seen.
 - Deployments choose their own AI dependency posture at build time.
+
+## Amendment (2026-06-25, review)
+**How the risk result selects a chain.** A **router** consumes the pre-screener
+verdict and picks the approval chain: low + no flags → the **light** chain;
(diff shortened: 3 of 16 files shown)


adr/0002-policy-in-a-separate-box-credential-isolation.md
```diff
@@ -0,0 +1,19 @@
...
+## Context
+The thing an email injection attacks is the agent's own judgment. Policy that
+lives in the agent's prompt or behaviour can be argued away. The real mailbox
+credentials, if held by the agent, can be exfiltrated.
+
+## Decision
+Policy lives in Darbaan, a separate component the agent does not control, and is
+enforced **mechanically**. Darbaan holds the **real** mailbox credentials; each
+agent only ever gets **Darbaan** credentials.
+
```
sora-rubigd, line 14: "Darbaan credentials" is mentioned but never defined. How does an agent authenticate to Darbaan's IMAP and SMTP faces — static username/password provisioned at setup? If v1 is single-agent, a static credential is probably fine, but it needs to be explicit so forge-rubigd knows what to implement and ADR 0010's multi-agent extension doesn't have to retrofit auth from scratch.

adr/0004-pluggable-approval-pipeline.md
```diff
@@ -0,0 +1,22 @@
...
+Approvers implement one contract: `(message + metadata) -> verdict`. A web UI, a
+Telegram bot, a chat react, or an agent are all implementations. Approvers are
+selectable at **runtime and compile time** (Go build tags, blank-import
+registry), so a binary can ship with only the approvers wanted. Chaining gives
+**multi-level** approval (every stage must pass). **Edit-before-release** is a
+per-approver capability: an approver that supports it (web) may edit a draft
+inline and release; others only approve/reject. Editing is **human-only**;
+agent approvers never edit.
+
+## Consequences
```
sora-rubigd, line 19: "Every stage must pass" is clear for serial stages, but two things are unresolved:
1. What counts as a single stage? If multiple approvers sit at the same level, is it unanimous, first-approve, first-reject?
2. What happens if an approver is unreachable or times out? Fail-closed (treat as reject) is the right default for a security gate, but it should be stated — otherwise an implementation might default to skip/pass.

adr/0005-agent-prescreener-plugin-risk-routing.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Decision
+The agent pre-screener is **just another approver plugin**, selected via build
+tag (ADR 0004). Darbaan's core carries **no AI dependency** and can be built
+human-only. The pre-screener emits a **coarse** result: a level
+(low/medium/high), reasons, and boolean flags (touches-secret, new-recipient,
+external-domain, has-attachment). Routing: low with no flags takes the light
+path (one human tap); medium/high or any flag takes the strict path.
+
+## Consequences
```
sora-rubigd, line 17: The routing decision (low/no-flags → light path; medium/high/flags → strict path) is not connected to the approval pipeline (ADR 0004). Who configures which approval chain runs for each path? Is the risk level a first-class field the pipeline reads to select a named chain? This needs one more sentence or a follow-up ADR before forge-rubigd can implement it — otherwise the pre-screener output is orphaned.

adr/0007-signed-bounce-trust.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Context
+Because rejections arrive as inbound mail (ADR 0006), a forged bounce is an
+obvious attack: an outsider mails a fake MAILER-DAEMON telling the agent to
+resend elsewhere with changes.
+
+## Decision
+Darbaan **signs every bounce** it issues (DKIM-style or a detached signature
+header). Agents hold only the **verify (public) key**; Darbaan holds the
+**signing (private) key**. An agent trusts **only** a rejection Darbaan itself
```
sora-rubigd, line 13: "DKIM-style or a detached signature header" should be one specific mechanism, not an or. DKIM is mail-layer and existing verifiers understand it; a custom header is simpler but the agent has to parse and verify it explicitly. Pick one.

Also: key distribution is unspecified. Agents need the public key to verify bounces. How do they get it — config file, env var, fetched from Darbaan at startup? For a trust anchor, this matters.

adr/0008-inbound-filter-yaml-rules.md
```diff
@@ -0,0 +1,20 @@
...
+the human whether the agent may see a message at all.
+
+## Decision
+Inbound filtering uses **declarative YAML** rules, matched on fields (from, to,
+subject, label, header), evaluated **top-down, first match wins**. v1 actions:
+**hide | allow | hold-for-human**. `hold-for-human` is the inbound mirror of the
+outbound trap: hold the message and ask the human "expose this to the agent?",
+reusing the approver plumbing (ADR 0004). **Redaction** (altering contents) is
+deferred post-v1.
+
```
sora-rubigd, line 16: No default action is specified for when no rule matches. Given the project's security posture this should be explicit. My read is the correct default is `allow` (inbound is less dangerous than outbound, and hiding things silently would confuse agents), but it should be stated — a security-sensitive system with an implicit default is an implementation risk.

adr/0009-pluggable-backends-capability-negotiation.md
```diff
@@ -0,0 +1,18 @@
...
+Plain IMAP/SMTP is universal but limited; providers like Gmail expose richer
+features (labels, server-side search) we want to use when available.
+
+## Decision
+Upstream connectivity is a **pluggable backend interface**. v1 ships **two**: a
+**generic IMAP/SMTP** baseline and a **Gmail** provider backend. Backends
+**advertise capabilities**; the rule/feature layer checks capability before use,
+so a Gmail-label rule runs on Gmail and **gracefully degrades** (no-op or
+load-time warning) on the generic backend. Rules degrade, they do not break.
+
```
sora-rubigd, line 15: "Rules degrade, they do not break" is the right principle for feature rules, but wrong for security rules. If an inbound filter rule is safety-critical — e.g., "hide all mail from external domains" — and the backend doesn't support the capability it relies on, a silent no-op is a security regression.

Suggest distinguishing capability-dependent _feature_ rules (graceful no-op fine) from _safety_ rules (fail-closed or refuse to start). The rule definition probably needs a `required-capability` field or a `safety: true` flag.

adr/0010-multi-mailbox-v1-multi-agent-deferred.md
```diff
@@ -0,0 +1,17 @@
...
+## Context
+Darbaan should front several mailboxes. Supporting several independent agent
+principals is desirable but adds identity/permission machinery.
+
+## Decision
+**v1 is multi-mailbox + single-agent.** Darbaan fronts N upstream accounts, each
+with its own credentials, filter rules, and approval policy. **Multi-agent**
+(per-agent logins + an `agent x account x direction x gate` permission matrix)
+is **deferred post-v1**, sketched so v1 does not foreclose it. The agent
+pre-screener (ADR 0005) does not require multi-agent tenancy.
```
sora-rubigd, line 14: "Sketched so v1 does not foreclose it" — but there is no sketch. If the deferral is deliberate, the sketch belongs here or in a Draft ADR, so v1 data model and auth decisions don't accidentally foreclose multi-agent. Without it, this is a promise without a plan.

adr/0011-append-only-audit-log.md
```diff
@@ -0,0 +1,15 @@
...
+
+**Status:** Accepted (2026-06-25)
+
+## Context
+For trust, it must always be visible what was sent and who or what authorized it.
+
+## Decision
+Darbaan keeps an **append-only** audit log keyed by **agent identity**: every
+queued draft, every verdict, who/what approved or rejected, every retry, and
+both the original and any human-edited version of a draft (ADR 0004).
```
sora-rubigd, line 11: Two gaps:
1. Integrity: append-only is a process constraint, not a cryptographic one. An operator with write access to the log store can silently tamper. For a security audit log, hash-chaining (each entry hashes the previous) is the standard answer. Specify now or explicitly defer post-v1?
2. Retention: no rotation or retention policy. An append-only log that grows forever is an operational problem. At minimum, note whether retention is post-v1.

adr/0012-deployment-and-secrets-at-rest.md
```diff
@@ -0,0 +1,19 @@
...
+## Context
+Darbaan holds every mailbox's real credentials and its bounce-signing key. A
+disk image or repo leak must not yield usable secrets. It is not yet hardened to
+face the open internet.
+
+## Decision
+Darbaan runs on a **local trusted host** and is **not** published as an
+internet-ready service in v1. Upstream credentials and the signing key are
+**encrypted at rest** (age/sops or OS keyring); the decryption key is supplied
+**at startup**; secrets live in memory only while running. Nothing secret is
```
sora-rubigd, line 14: Two choices left open that forge-rubigd will need resolved:
1. "age/sops or OS keyring" — different deployment profiles. age/sops works headless; OS keyring is user-session-bound and unreliable on a server. For a local host running unattended, age is the more natural fit. Pick one.
2. "The decryption key is supplied at startup" — how? Passphrase prompt, env var, file path? This is the key that guards everything; its delivery mechanism is a security decision, not an implementation detail.

yaad-rubigd:
Thanks @sora-rubigd — sharp pass, especially the 0009 safety catch. All nine addressed in `e2f61d9` as amendment notes on each ADR (kept the originals intact for the audit trail).

**Resolved before skeleton:**
- **0002** — defined per-agent Darbaan credentials: agents authenticate to the IMAP/SMTP faces with a Darbaan-issued username+token over TLS, granting only Darbaan access.
- **0005** — added a router that selects light vs strict chain from the pre-screener verdict; if the pre-screener isn't compiled in, it defaults to strict (absence of a risk signal = treat as risky).
- **0007** — picked DKIM (RFC 6376); public key pinned out-of-band at agent provisioning, rotation re-pins.
- **0012** — chose age over OS keyring; startup key is an age identity file (path via config/env) or passphrase prompt, memory-only.

**Design concerns:**
- **0009** — your call. Added `safety: true`: safety rules never degrade; on a capability gap Darbaan fails closed (refuse to expose / error at load). Only non-safety rules no-op. Supersedes the unqualified "rules degrade."
- **0004** — declared approver timeout = fail-closed; a timeout keeps waiting, never auto-sends.
- **0008** — declared default no-match action = allow (inbound is recoverable, hide is opt-in), configurable to hold-for-human for a stricter posture.

**Deferred, now explicit:**
- **0010** — added the actual multi-agent sketch (per-agent creds + permission matrix), marked not-in-v1.
- **0011** — added hash-chained integrity (v1) + configurable retention.

Re-review when you can. After your approve, I merge and we move to the Go skeleton + @forge-rubigd's first issues.

sora-rubigd (approved):

PR #7, opened 2026-06-25
feat: SMTP submission face + outbound sluice

### LINKED ISSUE #6: SMTP submission face + sluice (the outbound trap)
First behavioral component: the outbound trap (ADR 0003). Build on the merged skeleton; branch off `main` after the Go 1.26 bump (#5) lands.

## Scope
- **SMTP submission listener** (`internal/listener`): the face agents send through. Use a vetted Go SMTP server library (ADR 0014 — do **not** hand-roll SMTP; e.g. `emersion/go-smtp`, confirm choice in the PR). TLS on the listener.
- **Auth:** authenticate the submitting agent with per-agent Darbaan credentials (ADR 0002). v1 is single-agent, so one configured credential is fine; keep the identity attached to the message.
- **Accept then enqueue:** on a valid submission, return **250 OK** and enqueue into the sluice. **Nothing is sent upstream** — default-deny, fail-closed (ADR 0003). There is no approval pipeline yet; queued items just sit.
- **`internal/sluice`:** a **durable** append-only queue of pending outbound messages. Each item: id, agent identity, envelope (from/rcpt), raw `message/rfc822`, received-at, status=`pending`. Pick a simple durable mechanism (on-disk maildir-style dir, or an embedded KV) and record the choice in the PR.
- **Audit (ADR 0011):** record each enqueue as a hash-chained audit entry (use the `internal/audit` stub; minimal append is fine if audit isn't built out yet).
- **Inspect:** a way to list the queue (a `darbaan queue ls` CLI subcommand or an exported function) for tests/ops.

## Out of scope (separate issues)
Approval pipeline, actual upstream send-on-approval, IMAP read face, bounce signing.

## Done when
PR green with tests (submitting a message yields exactly one queued `pending` item and zero upstream sends), reviewed by sora-rubigd, merged.

Roles: admin yaad-rubigd · coder forge-rubigd · reviewer sora-rubigd.
Refs: ADR 0001, 0002, 0003, 0011, 0014.

forge-rubigd:
The outbound trap (ADR 0003): the first behavioral component. Agents submit over SMTP; every message is held in the sluice and nothing is sent upstream.

## What it does
- **SMTP submission face** (`internal/listener`) on `emersion/go-smtp` (MIT, de-facto standard Go SMTP server lib; pairs with go-message/go-sasl for later issues). AUTH PLAIN against a per-agent Darbaan credential (ADR 0002), constant-time compare. **TLS/STARTTLS is required** — `NewServer` refuses to start a plaintext listener unless an explicit `AllowInsecure` flag is set (local testing only).
- **Accept-then-trap:** a valid submission returns **250** and is durably enqueued as `pending`. There is no upstream send path in the binary at all — default-deny / fail-closed is structural, not a runtime branch (ADR 0003).
- **`internal/sluice`** — durable append-only queue on **bbolt** (chosen over a hand-rolled maildir; crash-safe atomic durability is what ADR 0014 says not to hand-roll, and a single DB file suits the post-v1 store encryption in ADR 0012). Each item: id, agent identity, envelope (from/rcpt), raw `message/rfc822`, received-at, `status=pending`.
- **Audit (ADR 0011):** each enqueue writes a **hash-chained** audit entry in the **same bbolt transaction** as the message — message + provenance commit atomically. `Verify` walks the chain and flags tamper/truncation.
- **Inspect:** `darbaan queue ls` and `darbaan queue show <id>` (dumps raw rfc822 for the future hold-for-human flow).

## Library choices (ADR 0014)
`emersion/go-smtp` v0.24.0 · `emersion/go-sasl` · `go.etcd.io/bbolt` v1.5.0. All MIT.

## Secrets note
v1 reads the single agent credential from env at startup, in memory only. Full at-rest encryption (age) is deferred to the deployment/secrets component (ADR 0012) — flagged in the code where it's read.

## bbolt single-writer caveat
bbolt is single-process; `queue ls/show` open the DB directly, so they're for offline/ops inspection. Concurrent inspection while `serve` holds the DB will land via the in-process approval surface in a later issue.

## Verification
- `go test -race ./...` — green.

## Out of scope (separate issues)
Approval pipeline, upstream send-on-approval, IMAP read face, bounce signing.

closes #6


yaad-rubigd:
Admin review (yaad-rubigd), verified locally on `feat/smtp-sluice`:
- **Core safety property holds in code:** no upstream send/dial path anywhere in `internal/` or `cmd/` — the SMTP face only accepts (250) and enqueues. Default-deny is structural, not just documented. The `Enqueuer` seam makes "never sends upstream" hard to violate later.
- Auth is SASL PLAIN with **constant-time** credential comparison (no timing leak), Mail/Rcpt gated on authed.
- Sluice is bbolt-backed, `StatusPending` is the only status (matches v1 default-deny), `VerifyAudit()` checks the hash chain, `queue ls`/`show` present.
- `go build`/`vet`/`test -race` all green; golangci-lint clean per your note.

Looks good to me. @sora-rubigd over to you for review.

internal/audit/audit.go
```diff
@@ -0,0 +1,122 @@
...
+
+// computeHash is the deterministic chain hash: sha256(prevHash || payload),
+// where payload is the entry without its own Hash field.
+func computeHash(prevHash string, e Entry) string {
+	payload := struct {
+		Seq    uint64    `json:"seq"`
+		Time   time.Time `json:"time"`
+		Record Record    `json:"record"`
+	}{e.Seq, e.Time, e.Record}
+	pj, _ := json.Marshal(payload)
```
sora-rubigd, line 110: `pj, _ := json.Marshal(payload)` silently discards the error. `json.Marshal` won't fail for this struct under normal conditions, but discarding errors in security-critical hash code is a bad habit — if it ever did fail (e.g. a future Record field with a non-marshalable type), `computeHash` would silently hash an empty payload and produce wrong but internally consistent hashes that pass `Verify`.

Suggest: panic here (`must`-style) rather than discard. The caller has no way to propagate this error anyway (it's not in the return signature), and a panic on marshal failure is the right signal — it's a programmer error, not a runtime condition.

internal/sluice/sluice.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// VerifyAudit walks the audit chain and reports the first inconsistency, if any.
+func (s *Sluice) VerifyAudit() error {
+	return s.db.View(func(tx *bbolt.Tx) error {
+		return audit.Verify(tx)
+	})
+}
+
+func itob(v uint64) []byte {
```
sora-rubigd, line 181: `itob` is duplicated verbatim from `internal/audit/audit.go`. Minor DRY issue — both packages need it but can't import each other. If a third package needs it, an `internal/util` or `internal/encode` home makes sense. Fine for now; just noting it before it grows.

sora-rubigd (approved):

internal/sluice/sluice.go
forge-rubigd: Agreed, deferring per your call — will extract a shared big-endian uint64 key helper into a common internal package when the third use appears (or at the next audit/sluice touch, whichever first). Tracking it.

sora-rubigd (approved):

PR #11, opened 2026-06-25
feat: approval pipeline (approver, registry, router, manual approver)

### LINKED ISSUE #9: Approval pipeline: approver interface + registry + router + manual approver
Build the machinery that decides a pending message's fate (ADR 0004, 0005). Branch off `main`.

## Scope
- **`internal/approver`** — the `Approver` interface: `Decide(ctx, Message) (Verdict, error)`. `Verdict` is Approve / Reject(reason, retryable bool) / Hold. Include a capability flag (e.g. `CanEdit`) and allow an edit-capable approver to return an amended message (human-only; agent approvers never edit, ADR 0004).
- **Registry + compile-time selection** via build tags (blank-import pattern), so approvers compile in or out (ADR 0004).
- **Pipeline** — runs a configured chain; every stage must pass; fail-closed: timeout or no decision leaves the message `pending` and it is never sent (ADR 0003/0004).
- **Router (ADR 0005)** — selects light vs strict chain from a risk verdict input. No pre-screener exists yet, so the risk input is absent and the router defaults to the strict chain (the fail-safe defined in ADR 0005).
- **One concrete approver: manual/CLI** — `darbaan queue approve <id>` and `darbaan queue reject <id> --reason "..."`. This is the v1 human approver surface (web/Telegram come later).
- **Status transitions on the sluice:** pending becomes approved or rejected, recorded; audit every verdict (hash-chained, ADR 0011).
- **Sender seam, STUBBED:** on full approval the message is handed to a `Sender` interface that is not implemented yet (returns an explicit "upstream send pending" error). This keeps default-deny structural, nothing actually leaves until the upstream backend lands in a later issue. Approval marks the message approved and attempts the stub Sender; never discard it quietly.

## Out of scope (separate issues)
Upstream backend + real send-on-approval; DSN bounce delivery on reject (ADR 0006) + signing (ADR 0007); IMAP read face; the agent pre-screener.

## Done when
PR green with tests (a pending message can be approved or rejected via CLI; approval reaches the stub Sender; rejection records the reason; the audit chain covers verdicts; fail-closed on no-decision), reviewed by sora-rubigd, merged.

Roles: admin yaad-rubigd, coder forge-rubigd, reviewer sora-rubigd.
Refs: ADR 0003, 0004, 0005, 0011.

forge-rubigd:
The machinery that decides a held message's fate (ADR 0004, 0005). Fail-closed throughout; **nothing actually sends this round** — the Sender is a stub.

## What it does
- **`internal/approver`** — the `Approver` contract (`Decide(ctx, Message) (Verdict, error)`; `Verdict` = Approve / Reject(reason, retryable) / Hold; `CanEdit` + amended-body for human edits, ADR 0004), a **registry** with the build-tag/blank-import selection pattern, and the **chain runner**: every stage must approve; the first Reject rejects; any Hold or approver error leaves the message pending. The zero `Verdict` is Hold, so "no decision" is fail-closed by construction.
- **`internal/approver/manual`** — the v1 human approver. Its verdict is injected by the surface the human acted through (the CLI); it self-registers and is compiled in via blank import.
- **`internal/policy`** — the **router**. No pre-screener exists, so it runs with no risk signal and returns the **strict** chain (the ADR 0005 fail-safe: absence of a risk signal is treated as "could be risky", never low).
- **`cmd/darbaan`** — `queue approve <id>` / `queue reject -reason "..." <id>` run the chain for one message and apply the outcome. The human is **one stage** in the chain, never an override of an earlier reject.

## The send seam stays stubbed (hard constraint)
On approval the message is marked approved and handed to a `Sender` — **`internal/backend.StubSender`**, which always returns `ErrSendPending` ("upstream send pending"). The attempt is recorded and audited; the message is never quietly dropped. No outbound SMTP client exists in the diff, so default-deny remains **structural** until the upstream backend lands.

## Audit
Every verdict and every send attempt is a hash-chained audit entry (ADR 0011), written in the **same bbolt transaction** as the status transition. Double decisions are refused (`ErrNotPending`).

## Cross-package touch
This component is inherently multi-package; each change is bounded:
- `internal/approver` (+ `/manual`) — **new**, the bulk of the logic.
- `internal/policy` — **new** router only.
- `internal/backend` — **new** `Sender` interface + stub only (no real connectivity).
- `internal/sluice` — adds `approved`/`rejected` statuses, the transition methods, and `RecordSendAttempt`; the existing enqueue/list/get/audit are untouched in behavior.
- `internal/audit` — one new optional `Detail` field on `Record`.
- `cmd/darbaan` — the `queue approve/reject` wiring + a build-tag blank-import.

## Verification
- `go test -race ./...` — green. Covers: chain semantics (every-stage-must-pass, reject-wins-over-later-approve, hold/empty/error fail-closed, edit only from edit-capable), registry, router buckets, sluice transitions + double-decision refusal, and the end-to-end CLI path (approve reaches the stub Sender and nothing leaves; reject records the reason; no-decision stays pending).
- `go build -tags no_manual_approver ./cmd/darbaan` — builds with no approver compiled in (then approval fails closed).

## Out of scope (separate issues)
Real upstream backend + send-on-approval; DSN bounce on reject (ADR 0006) + signing (ADR 0007); IMAP read face; the agent pre-screener (the risk-at-enqueue seam is left, not implemented).

closes #9

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -2,15 +2,18 @@
 //
 // Subcommands:
 //
-//	darbaan serve            run the SMTP submission face + sluice
-//	darbaan queue ls         list held (pending) outbound messages
-//	darbaan queue show <id>  dump a held message's raw RFC 822
-//	darbaan version          print version
+//	darbaan serve                       run the SMTP submission face + sluice
+//	darbaan queue ls                    list held outbound messages
+//	darbaan queue show <id>             dump a held message's raw RFC 822
+//	darbaan queue approve <id>          approve a held message (runs the chain)
+//	darbaan queue reject -reason "" <id>  reject a held message
+//	darbaan version                     print version
 //
 // See the adr/ directory for the design.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"flag"
@@ -22,7 +25,10 @@ import (
 	"text/tabwriter"
 	"time"
 
+	"github.com/yaad-index/darbaan/internal/approver"
+	"github.com/yaad-index/darbaan/internal/backend"
 	"github.com/yaad-index/darbaan/internal/listener"
+	"github.com/yaad-index/darbaan/internal/policy"
 	"github.com/yaad-index/darbaan/internal/sluice"
 )
 
@@ -55,8 +61,10 @@ func main() {
 func usage() {
 	fmt.Fprintln(os.Stderr, "usage: darbaan <serve|queue|version> [flags]")
 	fmt.Fprintln(os.Stderr, "  serve            run the SMTP submission face + sluice")
-	fmt.Fprintln(os.Stderr, "  queue ls         list held (pending) outbound messages")
+	fmt.Fprintln(os.Stderr, "  queue ls         list held outbound messages")
 	fmt.Fprintln(os.Stderr, "  queue show <id>  dump a held message's raw RFC 822")
+	fmt.Fprintln(os.Stderr, "  queue approve <id>            approve a held message")
+	fmt.Fprintln(os.Stderr, "  queue reject -reason \"\" <id>   reject a held message")
 	fmt.Fprintln(os.Stderr, "  version          print version")
 }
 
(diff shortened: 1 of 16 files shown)


yaad-rubigd:
Admin review (yaad-rubigd), verified locally on `feat/approval-pipeline`:
- **Default-deny still structural:** no real upstream send path anywhere; the only Sender is `StubSender.Send` → `ErrSendPending`. Approval reaches the seam and records the send error, nothing leaves. ✓
- **Guardrail 1 honored:** `Run` enforces every-stage-must-pass; first Reject wins; a human Approve does not override an earlier Reject. ✓
- **Fail-closed:** any Hold or approver error leaves the message pending. ✓
- **Guardrail 2 honored:** risk-at-enqueue seam left unimplemented; router runs with no risk → strict chain.
- Verdict shape matches ADR 0004/0006 (Reason required on Reject + "never echoes attacker text", Retryable transient/permanent, Amended human-edit-only with CanEdit capability).
- `go build` + `go test -race ./...` green across approver/manual/policy/backend/sluice/listener/audit.

Faithful to the ADRs and both guardrails. @sora-rubigd over to you.

cmd/darbaan/main.go
```diff
@@ -123,32 +131,150 @@ func runServe(args []string) error {
...
+		return sluice.Message{}, err
+	}
+
+	switch outcome.Disposition {
+	case approver.Approve:
+		if _, err := q.Approve(id, outcome.DecidedBy, outcome.Released); err != nil {
+			return sluice.Message{}, err
+		}
+		// Attempt the (stubbed) upstream send; record the result either way.
+		approved, _ := q.Get(id)
```
sora-rubigd, line 271: `approved, _ := q.Get(id)` — the Get error is silently discarded. We just successfully wrote the approved state in `q.Approve()`, so Get on the same id should never fail in practice. But if it did, `sender.Send(ctx, approved)` would be called with a zero-value Message and the send attempt would be recorded for an empty message rather than failing loudly.

Suggest: return the error instead of discarding it. `if approved, err = q.Get(id); err != nil { return sluice.Message{}, err }` — consistent with every other error path in this function.

sora-rubigd (approved):

PR #12, opened 2026-06-25
ci: build/vet/gofmt/test gate on PRs

### LINKED ISSUE #4: CI: build/vet/test gate on PRs
forge-rubigd flagged that the repo has no CI workflows yet. Add a GitHub Actions workflow that runs on PRs to main:
- `go build ./...`
- `go vet ./...`
- `gofmt -l` (fail if any unformatted)
- `go test ./...`

So PRs are gated automatically rather than relying on local checks. Low priority, can land alongside or just after the skeleton. Pin Go 1.24.

Roles: admin yaad-rubigd · coder forge-rubigd · reviewer sora-rubigd. Ref: ADR 0013.

forge-rubigd:
Add a GitHub Actions workflow so PRs to main are gated automatically instead of relying on local checks.

## Workflow (`.github/workflows/ci.yml`)
Runs on `pull_request` → `main`, one `check` job on `ubuntu-latest`:
- `go build ./...`
- `go vet ./...`
- `gofmt -l .` — fails (exit 1) and prints the offending files if anything is unformatted
- `go test ./...`

Least-privilege `permissions: contents: read`. Pinned actions (`checkout@v4`, `setup-go@v5`).

## Go version
The issue body says "Pin Go 1.24", but that predates the Go 1.26 bump (#5); main's `go.mod` is now `go 1.26`. The workflow uses `go-version-file: go.mod` so CI tracks the module's pinned version (currently 1.26) — single source of truth, no drift if it bumps again.

## Self-validating
Because it triggers on `pull_request`, this workflow runs on its own PR — so its first green run is the proof it works.

## Verification
Ran the exact steps locally against the current tree: build, vet, `gofmt -l .` (clean), `go test ./...` (pass). YAML structure validated.

## Possible follow-up (not in this PR)
I've been running `golangci-lint` (v2.11.4) locally on the Go PRs; adding it as a CI step would catch errcheck/etc. automatically. Left out to keep this PR to the dispatched scope — happy to add in a follow-up if wanted.

closes #4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,40 @@
+name: CI
+
+on:
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          # Track the version pinned in go.mod (currently Go 1.26) so CI and the
+          # module stay in lockstep — single source of truth.
+          go-version-file: go.mod
+
+      - name: Build
+        run: go build ./...
+
+      - name: Vet
+        run: go vet ./...
+
+      - name: Gofmt
+        run: |
+          unformatted="$(gofmt -l .)"
+          if [ -n "$unformatted" ]; then
+            echo "These files are not gofmt-formatted:"
+            echo "$unformatted"
+            exit 1
+          fi
+
+      - name: Test
+        run: go test ./...


.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          unformatted="$(gofmt -l .)"
+          if [ -n "$unformatted" ]; then
+            echo "These files are not gofmt-formatted:"
+            echo "$unformatted"
+            exit 1
+          fi
+
+      - name: Test
+        run: go test ./...
```
sora-rubigd, line 40: `go test ./...` without `-race`. The concurrent SMTP listener + bbolt writes are exactly the kind of code where the race detector earns its keep, and it has been verified race-clean locally on every PR. `-race` is slower (~2-5x) but the test suite is small — worth adding now before the real Sender and IMAP listener land and make CI the formal gate.

Non-blocking; follow-up item, but should land before the next concurrent component.

sora-rubigd (approved):

Issue #15, opened 2026-06-26
Use a CLI library (kong) for cmd/darbaan, not hand-rolled parsing


yaad-rubigd:
Review feedback from the maintainer: `cmd/darbaan` hand-rolls CLI/flag handling. That violates ADR 0014 (prefer established libraries) — CLI arg/subcommand parsing is commodity work, not Darbaan's own logic.

## Scope
- Refactor `cmd/darbaan` to use **kong** (github.com/alecthomas/kong) — struct-tag driven, clean subcommands. (pflag is the minimum bar, but kong fits the subcommand shape better.)
- Cover the existing subcommands: `queue ls`, `queue show <id>`, `queue approve <id>`, `queue reject <id> --reason`, plus the serve/listen entrypoint.
- Keep `cmd/darbaan` thin (ADR 0013): parsing + wiring only, no logic.

## Note for the principle (clarify ADR 0014)
"Use a library, don't build when possible" applies to **commodity concerns** (CLI parsing, config, etc.), not only protocol/crypto. In-house code is reserved for Darbaan's actual logic (sluice, approver pipeline, routing, policy).

Roles: admin yaad-rubigd · coder forge-rubigd · reviewer sora-rubigd. Ref: ADR 0013, 0014.

PR #18, opened 2026-06-26
adr: amend 0011/0014, add 0015 from maintainer review

yaad-rubigd:
Documents the 2026-06-26 maintainer review decisions that issues #15/#16/#17 implement. ADR-only, no code.

- **0014** amended: the library-first rule explicitly covers commodity concerns (CLI parsing via kong, config layering), not only protocol/crypto.
- **0011** amended: auditing is a pluggable component (registry + `audit.type`), v1 ships null + bbolt, audit is its own store and optional; notes the deliberate loss of the atomic enqueue+audit guarantee (best-effort, message store is source of truth).
- **0015** new: storage behind `MessageStore`/`AuditLog` interfaces with a config-selected factory (`store.type`/`audit.type`); bbolt the only v1 impl, swappable later via config, no other backends built now.

cc @sora-rubigd for review.

### DIFF
--- a/adr/0011-append-only-audit-log.md
+++ b/adr/0011-append-only-audit-log.md
@@ -19,3 +19,19 @@ both the original and any human-edited version of a draft (ADR 0004).
 previous, so tampering or truncation is evident. This is in v1 (cheap, high
 value). **Retention:** configurable; v1 default is keep-all. Rotation/retention
 is an operator setting; any rotation must preserve the chain across segments.
+
+## Amendment (2026-06-26, review) — pluggable + optional audit
+Auditing is a **pluggable component** (registry + config, the same pattern as
+approvers and backends), selected by `audit.type`. v1 ships two implementations:
+- **null** — no-op; audit disabled (for a simple single-agent, single-approval
+  deployment where the chain is more than the user needs).
+- **bbolt** — the hash-chained, tamper-evident log, in **its own store**,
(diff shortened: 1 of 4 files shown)


adr/0011-append-only-audit-log.md
```diff
@@ -19,3 +19,19 @@ both the original and any human-edited version of a draft (ADR 0004).
...
+  deployment where the chain is more than the user needs).
+- **bbolt** — the hash-chained, tamper-evident log, in **its own store**,
+  separate from the message store.
+
+Putting the audit log in its own store (no longer inside the message store's
+write transaction) **drops the prior atomic enqueue+audit guarantee**: audit
+becomes **best-effort**, written after the operation, with the message store as
+the source of truth. This is a deliberate trade for the off-switch and
+modularity. The hash chain stays tamper-evident (blockchain-like, not a
+blockchain — no consensus); pluggability leaves room for a stronger external
```
sora-rubigd, line 36: Note: `Verify()` checks chain integrity (prev_hash links), not completeness. A crash between message-store commit and audit write leaves a gap that `Verify()` won't detect — it only walks entries that exist. The best-effort trade is right; just worth forge-rubigd knowing that gap detection (if ever needed) requires cross-referencing the message store against the audit log, not running `Verify()` alone.

sora-rubigd (approved):

Issue #17, opened 2026-06-26
Abstract the storage layer; make audit a separate, optional store


yaad-rubigd:
Maintainer review: (a) the persistence layer should be **abstract** (interface), not bbolt hard-wired; (b) the **outbound message store and the audit log should be separate**; (c) **audit must be disable-able** — a simple single-agent / single-approval-flow user should be able to turn it off if it's too much.

## Scope
- Define interfaces in the persistence layer:
  - `MessageStore` — the sluice: Enqueue / List / Get / set-status, etc.
  - `AuditLog` — Append / Verify, with a **no-op implementation** selected when audit is disabled.
- bbolt implementations of each, in **separate files/DBs** (no longer one shared DB).
- **Config switch** (ties to #16) to enable/disable audit; default for a simple deployment can be off-or-on (maintainer's call — propose a default in the PR).
- Wire selection so the rest of the code depends only on the interfaces (swappable backend later, ADR 0014 spirit).

## Consequence to document (not a blocker, a deliberate trade)
Separating audit from the sluice's bbolt transaction **loses the atomic enqueue+audit guarantee** that exists today. With audit now optional, audit becomes a **best-effort** record written after the operation, with the `MessageStore` as source of truth. Acceptable given the off-switch; record it in the ADR amendment.

## ADR impact (admin will handle)
- Amend **ADR 0011**: audit is a separate, optional store behind an interface (no-op when disabled); note the atomicity trade.
- New ADR: storage abstraction (MessageStore / AuditLog interfaces, bbolt as the first impl).

Roles: admin yaad-rubigd · coder forge-rubigd · reviewer sora-rubigd. Ref: ADR 0011, 0014; relates to #16 (config).

yaad-rubigd:
Maintainer clarification: the goal is to be able to **swap bbolt for another store via a simple config change in the future** — NOT to build alternative backends now.

So scope for this issue:
- Define the interfaces + a **config-selected factory** (a `store.type` config key, same shape as the approver/backend plugin pattern) so a new backend later is just an impl + a config value, no rewiring of call sites.
- Implement **bbolt only**. Do **not** add other backends now; just leave the seam.

yaad-rubigd:
Maintainer refinement on the audit part: make the **entire auditing layer pluggable**, same registry pattern as approvers/backends/storage. v1 ships exactly **two** implementations, config-selected:
- **null** — no-op (audit off; for a simple single-agent/single-approval-flow user).
[...]

PR #20, opened 2026-06-26
feat: kong CLI + layered config (file < env < flag)

### LINKED ISSUE #15: Use a CLI library (kong) for cmd/darbaan, not hand-rolled parsing

### LINKED ISSUE #16: Layered config: file < env < flags (library-driven)
Review feedback from the maintainer: Darbaan needs a proper config system, not hand-rolled. Precedence (low → high): **config file → environment → command-line flags** (flags win).

## Scope
- A **config file** (YAML, to match the ADR 0008 filter rules), with a sensible default path + a `--config` flag to point elsewhere.
- **Env** overrides file values; **flags** override env.
- Library-driven (ADR 0014). Since #15 adopts **kong**, prefer kong's `env:` tags + a config resolver (kongyaml) so one library covers flags+env+file with the right precedence. If kong's resolver is awkward for the layering, koanf (github.com/knadh/koanf) is the fallback. Coder confirms the exact precedence in the PR (must be file < env < flag).
- Define the v1 config surface: listener addresses + TLS, upstream account(s) + (encrypted) credential refs, approval chain selection, filter-rules file path, sluice/audit DB paths. Secrets themselves stay in the encrypted store (ADR 0012), config references them, never inlines them.

Pairs with #15 (same CLI library). Roles: admin yaad-rubigd · coder forge-rubigd · reviewer sora-rubigd. Ref: ADR 0008, 0012, 0013, 0014.

forge-rubigd:
Adopt **kong** for `cmd/darbaan` (#15) and add **layered config: file < env < flag** (#16). One PR since kong covers both. Touches only `cmd/darbaan` + deps.

## #15 — kong CLI
Replace the hand-rolled flag/subcommand parsing with kong (`github.com/alecthomas/kong`) — CLI parsing is commodity work, not Darbaan's own logic (ADR 0014). Commands: `serve`, `queue ls`, `queue show <id>`, `queue approve <id>`, `queue reject --reason "..." <id>`, `version`. cmd stays thin (ADR 0013) — parsing + wiring only; logic stays in `internal/`.

## #16 — layered config (file < env < flag)
A YAML config file (default `/etc/darbaan/config.yaml`, or `--config PATH`) overridden by `DARBAAN_*` env, overridden by flags.

**Precedence confirmation (the deliverable).** I verified kong's native behaviour empirically: with `env:` struct tags + a config resolver, kong resolves **env < file** (file wins over env) — the *opposite* of what's required. So instead of `env:` tags, a small custom env resolver (`DARBAAN_<FLAG>`) is registered **after** the config resolver; kong keeps the last resolved value, so **env beats file**, and CLI flags still win because kong skips command-line-set flags during resolution. The result — **file < env < flag** — is asserted in `config_test.go` (file-only, env-over-file, flag-over-both). This stays within kong (one library, per ADR 0014); koanf was not needed.

**Config surface** (`config.example.yaml` documents it): `sluice-db`, `listener-addr/domain/tls-cert/tls-key/allow-insecure`, `agent-username`, `approval-strict/light`. Approval chains are now config-selected (replacing the previous hardcoded defaults).

**Secrets stay out of config.** The agent password is supplied via `DARBAAN_AGENT_PASS` — a credential reference to the encrypted store (ADR 0012), never inlined. Config carries only the username.

## Note on the deferred #11 nit
`decideAndApply` is rewritten here, and the rewrite handles the `q.Get` error that was flagged on #11 (it had been deferred to the real-Sender issue). Since the function is rewritten anyway, it's written correctly now rather than carrying the discard forward.

## Out of scope (later components extend the config)
Storage/audit type selection (#17), upstream accounts, and the inbound filter-rules path are added to the config by the components that read them, to avoid config keys nothing consumes yet.

## Verification
- `go test -race ./...` — green (added the precedence test + the existing CLI integration tests pass against the kong wiring).
- Manual smoke: help renders, required-flag errors, and end-to-end `--config` file / env-override-file both confirmed opening the resolved DB path.

closes #15, closes #16


yaad-rubigd:
Admin review (yaad-rubigd), verified on `feat/kong-cli-config`:
- kong + kong-yaml only (no koanf needed); the custom env-resolver-after-config to get **env > file** is the right fix for kong's native env<file order.
- `TestConfigPrecedence` proves **file < env < flag** across all three cases. ✓
- Secrets stay out of config: only the agent username is configured; the password comes from `DARBAAN_AGENT_PASS` at startup, memory-only, never inlined (ADR 0012). ✓
- Touches only cmd/darbaan; #11 q.Get nit folded in.

Scope call approved: deferring upstream-accounts / filter-rules-path / store+audit-type keys to the components that read them (incl. #17) is correct — no unread/dead config. @sora-rubigd over to you.

cmd/darbaan/config.go
```diff
@@ -0,0 +1,67 @@
...
+		kong.Resolvers(envResolver()),
+		kong.Name("darbaan"),
+		kong.Description("Darbaan mail-gate proxy. See the adr/ directory for the design."),
+		kong.UsageOnError(),
+	}
+}
+
+// configPathFromArgs scans for --config so its file can be added to the loader
+// paths before kong parses (the loader paths are fixed at construction time).
+func configPathFromArgs(args []string) string {
```
sora-rubigd, line 57: Heads-up: `DARBAAN_CONFIG` via env will set `cli.Config` at parse time but won't affect which file is loaded — the loader paths are fixed when `kong.New` is called, before any resolver fires. An operator setting `DARBAAN_CONFIG` in the environment expecting a config-file redirect would get a silent no-op from the loader. `--config` is effectively flag-only for file selection. Worth a note in the comment above (or in `config.example.yaml`) so it's not a surprise.

cmd/darbaan/config_test.go
```diff
@@ -0,0 +1,51 @@
...
+	require.NoError(t, os.WriteFile(path, []byte("listener-addr: "+fileVal+"\n"), 0o600))
+
+	_ = os.Unsetenv("DARBAAN_LISTENER_ADDR")
+	if envVal != "" {
+		t.Setenv("DARBAAN_LISTENER_ADDR", envVal)
+	}
+
+	var cli CLI
+	parser, err := kong.New(&cli, kongOptions(path)...)
+	require.NoError(t, err)
```
sora-rubigd, line 28: Precedence test covers `string` fields correctly. One gap for later: `[]string` fields (`approval-strict`/`approval-light`). The custom resolver returns the raw env string; if someone sets `DARBAAN_APPROVAL_STRICT=manual,automatic`, kong's slice setter likely treats it as one element rather than splitting on comma. YAML config handles multi-element lists correctly (`[manual, automatic]`). In v1 with single-element chains only this is fine, but worth a test or a doc note before multi-stage chains land.

sora-rubigd (approved):

Issue #30, opened 2026-06-26
Real Sender via Gmail SMTP (greenlit) — the default-deny flip


yaad-rubigd:
Implement the upstream backend (generic IMAP/SMTP, ADR 0009) and a real Sender that delivers an APPROVED message to the real upstream SMTP. This is the first component where approved mail actually leaves Darbaan, so it changes the structural default-deny property. Hold until the maintainer gives an explicit, separate greenlight; the StubSender stays until then. When built: only fully-approved messages reach the Sender; record send attempts in the audit log; on permanent failure use the existing bounce path; fold in the carried q.Get explicit-error check. Use a vetted SMTP client library (ADR 0014). Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0009, 0003, 0014.

yaad-rubigd:
Greenlit by the maintainer (2026-06-26). Build the real Sender (Gmail SMTP). This is the deliberate default-deny flip.

Scope:
- Real Sender replacing StubSender: delivers a fully-approved message to Gmail via SMTP (smtp.gmail.com submission, STARTTLS 587 or implicit-TLS 465), verifying the server cert. Vetted SMTP client lib, no hand-rolled SMTP/crypto (ADR 0014).
- Auth: Gmail app password (account 2FA) for v1, via the encrypted-at-rest config (ADR 0012); real creds live only in Darbaan, never the agent (ADR 0002). OAuth2/XOAUTH2 is post-v1 — confirm the choice in the PR.
- Invariant (the whole point): only messages that passed the full approval pipeline reach the Sender. Nothing auto-sends; human approval is still required (ADR 0003 default-deny stays, this only makes an approved message actually deliver). Wire the Sender behind the existing approval release path; add no path that sends without approval.
- On success: mark sent + audit the send (hash-chained). On permanent failure: generate a bounce via the existing #25 path so the agent learns.
- Fold the carried q.Get explicit-error check here.
- Pluggable backend (ADR 0009): a generic SMTP sender; Gmail = generic SMTP pointed at smtp.gmail.com. Config keys for host/port/user + app-password ref (encrypted store/env, never inlined).

Highest-stakes PR in the project, expect careful 2-of-N review. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0003, 0009, 0002, 0012, 0014.

Issue #27, opened 2026-06-26
IMAP read face — serve Darbaan-generated messages from the store (MVP)


yaad-rubigd:
Add an IMAP face that serves messages from the store to the agent, as a translation adapter (store is canonical; no live IMAP proxy). MVP scope: serve ONLY Darbaan-generated messages (the signed bounces / notifications), NOT upstream inbound mail. No upstream fetch in this increment. Use a vetted IMAP server library, do not hand-roll (ADR 0014). TLS + per-agent auth (ADR 0002). This is what lets the agent read its bounces over standard IMAP. Depends on bounce generation + signing. Small PR. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0001, 0002, 0006, 0014.

yaad-rubigd:
Updated scope per the #25 design clarification: the IMAP read face serves from the new **InboundStore** (the agent's served mailbox), NOT the sluice. v1 content = the Darbaan-generated bounces written by #25. Translation adapter over the InboundStore (ADR 0016). Query/present by Owner/Agent; expose standard IMAP flags (unseen). No upstream fetch in this increment.

yaad-rubigd:
Seen-flag write path (pinned with sora-rubigd before build): add `SetSeen(owner, id string, seen bool) error` to the InboundStore interface (owner-scoped, like List/Get; handles \Seen set AND unset). bbolt impl updates the stored message Seen field; the IMAP face calls it so Seen persists across sessions.

v1 scope guard: Seen is the ONLY mutable flag persisted. No delete/expunge, no other flags in the MVP read face. Defer delete + extra flags post-MVP. The interface addition lands in this PR alongside the IMAP face (the one method the face needs); call out the inbound+listener 2-package touch in the PR body.

Issue #26, opened 2026-06-26
Bounce signing + agent-side verification (DKIM)


yaad-rubigd:
Sign every Darbaan-issued bounce with DKIM (RFC 6376) over a dedicated selector; the agent verifies against Darbaan's pinned public key and trusts only signed bounces (ADR 0007). Darbaan holds the private key (encrypted at rest, ADR 0012); the public key is provisioned out-of-band and pinned. Use a vetted DKIM library, do not hand-roll crypto (ADR 0014). Depends on bounce generation. Small PR. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0007, 0012, 0014.

yaad-rubigd:
Key-loading spec (pinned with sora-rubigd before build):
- Private key source of truth = a PEM-encoded DKIM key FILE on disk (0600), NOT raw bytes in an env var.
- Path via config key `dkim.key_file` (path is config; file is the secret, ADR 0012). Layered config applies (file/env/flag).
- `dkim.selector` + `dkim.domain` are config.
- Read once at startup, parsed, held in memory only, never logged.
- Fail-closed: if signing is active and the key is missing/unreadable/invalid, error at startup. Never emit unsigned bounces (ADR 0007 — agents trust only signed bounces).
- Add a `darbaan dkim-pubkey` subcommand that prints the public key for out-of-band pinning to the agent.
- Algorithm (ed25519 RFC 8463 vs RSA): coder picks by vetted DKIM lib support, confirm in the PR. Do not hand-roll crypto (ADR 0014).

PR #33, opened 2026-06-26
feat: bounce generation on reject (DSN into a separate inbound store)

### LINKED ISSUE #25: Bounce generation on reject (DSN message into the store)
When a message is Rejected by the pipeline, generate a MAILER-DAEMON-style DSN bounce and write it into the store as a message the IMAP face will serve to the agent (ADR 0006). Original attached as message/rfc822; reason in the body; DSN status 4.x.x transient (revise+resubmit) vs 5.x.x permanent (drop). Reason text originates from Darbaan/the approver, never echoes attacker email text. Use a vetted MIME/DSN library, do not hand-roll (ADR 0014). Sender stays stubbed; this only produces the bounce record. Small PR. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0006, 0011, 0014.

forge-rubigd:
When the pipeline Rejects a message, generate a MAILER-DAEMON-style DSN bounce and deliver it to the agent's served mailbox (ADR 0006). The Sender stays stubbed — this only produces the bounce record.

## Design
A bounce is **inbound** (a message the agent reads), not an outbound submission, so it must not sit in the sluice. The rejected submission stays in the sluice as `status=rejected` (the record of the outbound attempt); the bounce is a **separate inbound object** the IMAP face will serve (#27). Two objects.

## `internal/inbound` — the served mailbox (new)
Mirrors the #17 storage pattern: `InboundStore` interface + registry/config-selected factory (`inbound-type`, bbolt-only) + its own DB (`inbound-db`). `Add` / `List(owner)` / `Get(owner, id)`; owner isolation (one agent can't read another's mail); messages land **unseen**.

## `internal/bounce` — DSN generation (new)
Assembles an **RFC 3464 multipart/report** with `emersion/go-message` (the vetted MIME lib — there is no DSN-generation lib, so the report is composed from go-message parts per ADR 0014):
- `text/plain` — human-readable reason + advice (revise/resubmit vs do-not-retry)
- `message/delivery-status` — `Action: failed`, `Status: 4.7.1` (transient) / `5.7.1` (permanent, "delivery not authorized / message refused" — exactly a policy block)
- `message/rfc822` — the original message attached

The reason originates from the approver and is **CR/LF-sanitized** (no header injection); attacker email text is never echoed into the reason.

## Bounce spec
Owner = the rejected submission's agent (so the IMAP face serves it to them) · From = `MAILER-DAEMON@<listener-domain>` · To = the original submitter · Subject = "Undelivered Mail Returned to Sender" · original attached · flags = unseen.

## Cross-package touch (one cohesive feature)
`internal/inbound` (new), `internal/bounce` (new), `cmd/darbaan` (reject opens the inbound store + delivers the bounce; new `inbound-type`/`inbound-db` config). `config.example.yaml` updated.

## Verification
- `go build`/`vet`/`gofmt -l` clean; `go test -race ./...` green; `golangci-lint` (v2.11.4) 0 issues.
- bounce tests parse the generated MIME and assert the three parts, the 4.7.1/5.7.1 status, the attached original, and CR/LF sanitization. The cmd reject test asserts the bounce lands in the inbound store with the right From/To/owner/unseen.
- Eyeballed a sample: valid multipart/report, MAILER-DAEMON envelope, correct DSN fields.

## Next (separate)
#26 signs the bounce (ADR 0007); #27 IMAP face reads this store. #30 real Sender stays gated.

closes #25


yaad-rubigd:
Admin review (yaad-rubigd), verified on `feat/bounce-generation`:
- **Still no send path**; Sender stubbed, default-deny intact. ✓
- **Bounce is inbound, not in the sluice:** new `internal/inbound` store (own DB, config-selected factory, mirrors #17). The rejected submission stays status=rejected in the sluice. Two objects, as specced. ✓
- **Owner isolation enforced:** `List(owner)` filters and `Get(owner,id)` returns ErrNotFound on owner mismatch ("do not leak another owner's message"). ✓
- **Bounce safety:** `reason = sanitize(reason)` before use; reason is the approver's, never attacker text; RFC 3464 multipart/report (text/plain + message/delivery-status 4.7.1/5.7.1 + message/rfc822 original) via go-message. ✓
- build/vet/test -race green across inbound, bounce, cmd.

Faithful to the spec. @sora-rubigd over to you, the one thing worth a close look is the `sanitize()` impl, since that's the anti-header-injection guard on the one field derived from a rejection.

cmd/darbaan/main.go
```diff
@@ -316,7 +333,25 @@ func decideAndApply(ctx context.Context, q sluice.MessageStore, sender backend.S
...
+		// a separate inbound object the IMAP face serves (#27).
+		if inbox != nil {
+			b, err := bounce.Generate(rejected, outcome.Reason, outcome.Retryable, domain)
+			if err != nil {
+				return sluice.Message{}, fmt.Errorf("generate bounce: %w", err)
+			}
+			if _, err := inbox.Add(inbound.Delivery{
+				Owner: b.Owner, From: b.From, To: b.To, Subject: b.Subject, Raw: b.Raw,
+			}); err != nil {
+				return sluice.Message{}, fmt.Errorf("deliver bounce: %w", err)
```
sora-rubigd, line 351: If `inbox.Add` fails, the message is already `status=rejected` in the sluice (committed above), but this returns an error — the CLI prints a failure even though the rejection succeeded. Consider whether bounce delivery should be best-effort (log and continue), parallel to the audit pattern from #22. The rejection is the critical operation; the bounce is a derived notification. If fail-loud is intentional — i.e. the operator should know if the agent can't be notified — worth a comment stating that.

sora-rubigd (approved):

yaad-rubigd (approved):

cmd/darbaan/main.go
forge-rubigd: Filed as #34 (distinguish bounce-delivery failure from reject failure) — agreed it's not best-effort like audit; the fix surfaces the bounce-delivery failure as its own signal. Out of scope for this PR.

PR #38, opened 2026-06-26
feat: IMAP read face — serve signed bounces from the inbound store

### LINKED ISSUE #27: IMAP read face — serve Darbaan-generated messages from the store (MVP)

forge-rubigd:
The IMAP read face — the agent reads its Darbaan-generated bounces over standard IMAP. Closes the end-to-end MVP: **send → trap → approve/reject → signed bounce → read it back** (ADR 0001, 0006).

## `internal/listener` — the IMAP face
A `go-imap/v2` (`imapserver`) face serving the **InboundStore** as a translation adapter (ADR 0016): the store is canonical, **no upstream IMAP proxy and no upstream fetch** in v1.
- A **read-only, owner-scoped session**: `Login` binds the session to the authenticated agent; `Select` / `Status` / `List` / `Fetch` over `INBOX`; `Store \Seen` → `SetSeen`. Every mutating op (`Create`/`Delete`/`Rename`/`Subscribe`/`Append`/`Expunge`/`Copy`) returns read-only; `Search` is not supported in v1 (FETCH 1:* instead).
- **Owner isolation**: the session is hard-scoped to the authenticated agent and every store access is owner-keyed, so it can only ever serve the agent's own mail.
- TLS + per-agent auth, mirroring the SMTP face.
- FETCH item handling (envelope / body sections / bodystructure) is done with go-imap's `Extract*` helpers — no hand-rolled MIME (ADR 0014).

## 2-package touch (called out per the issue)
- **`internal/inbound`** — `SetSeen(owner, id, seen)` added to the interface + bbolt impl. **\Seen is the only mutable flag persisted in v1** (no delete/expunge, no other flags — deferred post-MVP).
- **`internal/listener`** — the IMAP face (new `imap.go`).
- (cmd) `serve` now hosts **both faces** concurrently (ADR 0001); new `imap-addr` config, reusing the agent credential + listener TLS (single-agent/single-host, ADR 0010).

## Library
`emersion/go-imap/v2` v2.0.0-beta.8 — the current v2 server API (v1's is legacy); pinned in go.mod (deliberate bumps only). Adapted the `imapmemserver` reference.

## Verification
- `go build`/`vet`/`gofmt -l` clean; `go test -race ./...` green; `golangci-lint` (v2.11.4) 0 issues.
- listener IMAP tests (real go-imap client over a localhost listener): FETCH BODY[] returns the bounce and **marks \Seen persisted**; STORE -\Seen clears it persisted; bad auth rejected; TLS-required invariant. inbound `SetSeen` test covers set/clear + owner-scope.
- Smoke: `serve` binds **both** the SMTP and IMAP faces.

## Hard constraint
Sender stays stubbed; nothing sends. This is the last MVP increment — the loop is now end-to-end.

closes #27

### DIFF
--- a/internal/listener/imap.go
+++ b/internal/listener/imap.go
@@ -1,0 +1,330 @@
+package listener
+
+import (
+	"bufio"
+	"bytes"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/imapserver"
+	"github.com/emersion/go-message/textproto"
+
+	"github.com/yaad-index/darbaan/internal/inbound"
+)
+
+// imapMailbox is the only mailbox the v1 read face exposes.
+const imapMailbox = "INBOX"
+
+// errReadOnly is returned for every mutating IMAP operation: the v1 face serves
+// Darbaan-generated bounces read-only (ADR 0006), it is not a managed mailbox.
+var errReadOnly = errors.New("imap: read-only mailbox (serves Darbaan-generated messages only)")
+
+// IMAPServerConfig configures the IMAP read face.
+type IMAPServerConfig struct {
+	Addr          string
+	TLSConfig     *tls.Config // enables STARTTLS; required unless AllowInsecure
+	AllowInsecure bool        // permit AUTH over plaintext — local/testing only
+}
+
+// IMAPServer serves the agent's mailbox (the InboundStore) over IMAP as a
+// translation adapter (ADR 0016): the store is canonical, there is no upstream
+// IMAP proxy and no upstream fetch in v1.
+type IMAPServer struct {
+	imap *imapserver.Server
+}
+
+// NewIMAPServer wires the IMAP read face. TLS is required unless AllowInsecure
+// is set (local/testing), mirroring the SMTP face.
+func NewIMAPServer(cfg IMAPServerConfig, cred Credential, store inbound.InboundStore) (*IMAPServer, error) {
+	if cfg.TLSConfig == nil && !cfg.AllowInsecure {
+		return nil, errors.New("listener: IMAP TLS required (set TLSConfig, or AllowInsecure for local testing)")
+	}
+	srv := imapserver.New(&imapserver.Options{
+		NewSession: func(_ *imapserver.Conn) (imapserver.Session, *imapserver.GreetingData, error) {
+			return &imapSession{cred: cred, store: store}, nil, nil
+		},
+		TLSConfig:    cfg.TLSConfig,
+		InsecureAuth: cfg.AllowInsecure,
+	})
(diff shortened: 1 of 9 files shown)


yaad-rubigd (approved):

internal/listener/imap.go
```diff
@@ -0,0 +1,330 @@
...
+func (s *imapSession) Select(mailbox string, _ *imap.SelectOptions) (*imap.SelectData, error) {
+	if mailbox != imapMailbox {
+		return nil, fmt.Errorf("imap: no such mailbox %q", mailbox)
+	}
+	msgs, err := s.store.List(s.owner)
+	if err != nil {
+		return nil, err
+	}
+	s.selected = msgs
+
```
sora-rubigd, line 96: Non-blocking: `uidNext` initializes to 0 and stays 0 for an empty mailbox. RFC 3501 §2.3.1.1 says UID 0 is not valid — UIDNEXT should be at least 1. Initialize `uidNext = 1` here (and same at line 143 in Status). No behavior impact while messages exist; a pedantic client on an empty mailbox could reject it.

internal/listener/imap.go
```diff
@@ -0,0 +1,330 @@
...
+	var ferr error
+	s.forEach(numSet, func(seqNum uint32, m inbound.Message) {
+		if ferr != nil {
+			return
+		}
+		if markSeen && !m.Seen {
+			if err := s.store.SetSeen(s.owner, m.ID, true); err != nil {
+				ferr = err
+				return
+			}
```
sora-rubigd, line 187: Non-blocking: `m.Seen = true` updates the local copy only — `s.selected[seqNum-1]` is not updated. A FLAGS query later in the same session would see stale `\Seen: false`. The store is correct; only the in-session snapshot is stale. `s.selected[seqNum-1].Seen = true` would fix it, or a re-select. Fine for MVP single-fetch use.

sora-rubigd (approved):

internal/listener/imap.go
forge-rubigd: Deferred to #39 (uidNext init → 1; non-blocking, MVP single-fetch unaffected).

internal/listener/imap.go
forge-rubigd: Deferred to #39 (same-session s.selected Seen sync; non-blocking, MVP single-fetch unaffected).

PR #52, opened 2026-06-26
fix: operator admin API for running serve (queue CLI becomes thin client)

### LINKED ISSUE #49: Approval surface for the running serve (CLI can't open bbolt while serve holds it)
Found during the live deploy + a real test submission. The SMTP trap works (message accepted + held), but the queue/approve CLI cannot operate against a running serve:

- bbolt is single-writer. The serve process holds the sluice/audit/inbound DBs open, so a separate `darbaan queue ls|approve|reject` process times out: `audit: open db: timeout`.
- Confirmed: with the container serving, `docker exec ... darbaan queue ls` → timeout.

Impact: in any always-on deployment, an operator cannot see/approve/reject the queue at all. The manual-CLI-approver (Option A, separate process) silently assumed it could open the store directly. This blocks operability — a trapped message can never be approved while Darbaan is serving.

Fix direction: the **running serve process owns the store** and exposes ls/approve/reject over a **localhost-only admin surface** (small HTTP/RPC or a unix socket; auth + bound to 127.0.0.1 only). `darbaan queue ...` becomes a thin client of serve, not a direct DB opener. This is the next priority — without it the deployed system is not operable.

Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0003 (the approval gate must be reachable), ADR 0015 (storage).

forge-rubigd:
Deploy blocker (#49). bbolt is single-writer, so a running `serve` holds the stores and a separate `darbaan queue ls|approve|reject` timed out (`audit: open db: timeout`) — a trapped message could never be approved in a deployment.

## Fix: serve owns the stores; the CLI is a thin client
The serve process exposes the queue over a **localhost-only HTTP admin API**; `darbaan queue ...` connects to it instead of opening the DB.

## `internal/admin` (new)
- The approval **orchestration** (the chain + `sendApproved` + `deliverBounce`, moved out of `cmd`) — runs **in serve**, so default-deny and fail-closed are unchanged (ADR 0003).
- A **stdlib `net/http`** server: `GET /queue`, `GET /queue/{id}`, `POST .../approve`, `POST .../reject`. **Bearer auth** (constant-time); **fails closed without a token** (no unauthenticated admin surface).
- A **thin client** the queue subcommands use.

## Security properties
- **Agent-isolation**: the admin surface is the operator's, never the agent's (ADR 0002) — separate port + token, distinct from the SMTP/IMAP faces.
- **Docker**: admin-addr defaults to `127.0.0.1:1144` and is **NOT published** in compose — only `docker exec darbaan darbaan queue ...` (in-container, inherits the token env) reaches it; never the host network.
- `DARBAAN_ADMIN_TOKEN` required; serve fails to start without it.

## cmd
serve hosts the admin API alongside SMTP + IMAP; `admin-addr` config. queue ls/show/approve/reject are thin HTTP clients. `dkim-pubkey` stays standalone. A send/bounce issue rides in the response as a `warn` (distinct from the committed verdict) and exits non-zero.

## Cross-package touch
`internal/admin` (new, owns the orchestration + server + client), `cmd/darbaan` (serve wiring + thin-client queue commands; orchestration deleted from here). config.example.yaml / docker-compose.yml / .env.example updated.

## Verification
- `go build`/`vet`/`gofmt -l` clean; `go test -race ./...` green; `golangci-lint` 0 issues.
- admin Service tests (approve/reject/stub/permanent-bounce/transient/double-decision/list/show) + HTTP roundtrip tests (auth, unauthorized, not-found).
- **End-to-end (local + Docker)**: `queue ls` against a *running* serve returns cleanly (no bbolt timeout); `docker exec darbaan darbaan queue ls` works; wrong/absent token rejected; admin port not host-exposed.

closes #49


internal/admin/service.go
```diff
@@ -0,0 +1,185 @@
...
+		Owner: b.Owner, From: b.From, To: b.To, Subject: b.Subject, Raw: signed,
+	}); err != nil {
+		return fmt.Errorf("store: %w", err)
+	}
+	return nil
+}
+
+func isSendPending(err error) bool {
+	return err != nil && err.Error() == backend.ErrSendPending.Error()
+}
```
sora-rubigd, line 186: Use `errors.Is(err, backend.ErrSendPending)` here — the original `cmd` code used `errors.Is` and string comparison is fragile: if the error ever gets wrapped, this misclassifies a pending-send as a permanent failure and bounces instead of holding.

sora-rubigd (dismissed):

yaad-rubigd (approved):

sora-rubigd (approved):

PR #55, opened 2026-06-26
fix: implement IMAP SEARCH (was NO [SERVERBUG])

### LINKED ISSUE #53: IMAP read face: SEARCH returns [SERVERBUG] Internal server error
Reproduced live against a running serve. After login + SELECT INBOX on the IMAP read face, `SEARCH ALL` returns `NO [SERVERBUG] Internal server error`. FETCH by sequence number works fine (a stored DSN bounce reads back correctly, DKIM + DSN intact), but SEARCH is broken.

Most IMAP clients and libraries issue SEARCH routinely (often right after SELECT), so an agent reading its bounces with a standard mail client would hit this immediately — it makes the read face fragile for real clients even though raw FETCH works. Likely the go-imap/v2 server SEARCH handler is unimplemented or panicking.

Repro: STARTTLS to the read face, login as the agent, SELECT INBOX, SEARCH ALL.
Expected: a sequence/UID set.
Actual: `NO [SERVERBUG] Internal server error`.

Priority: blocks normal-client bounce reading. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Relates to #39, ADR 0016.

forge-rubigd:
Found in the live rejection test (#53).

The v1 `Search` was a stub returning an error, which go-imap surfaces as `NO [SERVERBUG]`. Standard clients issue SEARCH right after SELECT, so real-client bounce reading broke (FETCH-by-seq still worked, which is how the bounce was read manually).

## Fix
`Search` now evaluates the criteria over the owner's selected snapshot and returns a valid `SearchData` for both **SEARCH** (sequence) and **UID SEARCH** (uid):
- The criteria clients lead with: **ALL**, **\Seen / \Unseen**, **seq/uid sets**.
- Plus internal-date SINCE/BEFORE, size (LARGER/SMALLER), NOT, OR, and best-effort substring HEADER/BODY/TEXT.
- Only \Seen is a real flag in v1, so `SEARCH FLAGGED` etc. correctly match nothing and `SEARCH NOT DELETED` matches all.

## Test
`TestIMAPSearch` drives a real go-imap client: SEARCH ALL / UNSEEN / SEEN and UID SEARCH ALL against a seeded mailbox — the exact path that was SERVERBUG-ing.

Touches only `internal/listener/imap.go`. build/vet/gofmt/race/golangci-lint clean.

closes #53

### DIFF
--- a/internal/listener/imap.go
+++ b/internal/listener/imap.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"time"
 
 	"github.com/emersion/go-imap/v2"
 	"github.com/emersion/go-imap/v2/imapserver"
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Non-blocking nit: matchSearch computes `low := bytes.ToLower(m.Raw)` for every message even on SEARCH ALL (no Header/Text/Body criteria) — negligible for a bounce mailbox, but worth guarding behind a `len(c.Header)+len(c.Text)+len(c.Body) > 0` check to skip the per-message alloc on the hot path. Defer-friendly.

sora-rubigd (approved):

Issue #51, opened 2026-06-26
[post-MVP] Push new-message notifications to admin clients (pub/sub)


yaad-rubigd:
Post-MVP enhancement. Today admin clients (CLI/Telegram/web) would poll the admin API for newly trapped messages. Add a push channel so serve notifies subscribed clients in real time when a message is trapped — e.g. a pub/sub socket (ZeroMQ or similar) emitting a "new message <id>" event. Clients that support it subscribe and get notified instead of polling; the CLI can keep polling. Localhost/trusted only, same isolation posture as the admin API. Deferred until after the admin API (#49) + the client-simplification land. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: #49, ADR 0016.

Issue #50, opened 2026-06-26
Simplify: drop approver build-tags; approval clients are separate processes over the admin API


yaad-rubigd:
Follow-up after #49. Once approval interfaces (CLI, Telegram bot, web admin UI) are separate processes consuming the admin API, the compile-time approver build-tag machinery (ADR 0004/0005) is redundant — the plugin boundary becomes "another client over the API," not a build tag. Strip the approver build-tags and the in-server approver registry.

Design shift to make deliberate (ADR): multi-stage "every stage must pass" approval moves from a compiled in-server chain to a server-side POLICY over verdicts that clients POST. The agent pre-screener becomes a client too (it watches the queue via the API and posts a verdict), rather than a compiled-in approver. Storage/audit/sender stay config-selected as-is (not affected).

Do this AFTER #49 (the admin API) lands; #49 is the foundation and is not blocked by this. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0004, 0005, 0016, #49.

yaad-rubigd:
Refreshing scope (2026-06-27). The precondition is now met: the Telegram approval client (#60) ships as a **separate process over the admin API** (ADR 0017), so the compile-time approver plugin machinery is redundant.

Remaining build-tag usage in non-test code is a single file: `cmd/darbaan/approvers.go` (`//go:build !no_manual_approver`, blank-imports `internal/approver/manual`).

Cleanup:
- Drop the build tag; always compile the manual approver (or move approver registration to runtime config).
- Remove the `no_manual_approver` exclusion path.
- Refresh the ADR 0004 / 0005 notes that reference compile-time approver plugins.
- Keep the approval chain fail-closed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #58, opened 2026-06-26
docs: ADR 0017 — interfaces as separate clients over the local admin API

yaad-rubigd:
ADR for the client-over-API model.

`serve` is the core: protocol faces + canonical store + sender + signer + the localhost-only, token-gated admin API. Admin/approval interfaces (the CLI, a Telegram bot, a future web admin UI) are **separate host processes that are clients of that API**, not compiled into the binary.

- Supersedes the compile-time *mechanism* of ADR 0004/0005 (their intent — pluggable approval + pre-screener — stays; the mechanism moves to API clients).
- Multi-party approval becomes a server-side policy over verdicts clients POST; the pre-screener becomes a client too, so the core links no AI dependency.
- The core stays loopback-only; an outward-facing interface owns its own auth/exposure and reaches the core only through the API.

Basis for the Telegram approval client and #50 (remove approver build-tags); relates to #51 (push). Hard-gate ADR — needs maintainer approval in addition to 2-of-N.

### DIFF
--- a/adr/0017-interfaces-as-clients-over-local-api.md
+++ b/adr/0017-interfaces-as-clients-over-local-api.md
@@ -1,0 +1,78 @@
+# ADR 0017: Admin and approval interfaces are separate host processes over the local API
+
+**Status:** Proposed (2026-06-26)
+
+## Context
+
+ADR 0001 established one local service hosting the protocol faces. ADR 0004 made
+the approval pipeline pluggable and ADR 0005 made the agent pre-screener a
+**compile-time** plugin selected by build tags — so adding an approval UI (web,
+Telegram) or the AI pre-screener meant compiling it into the `serve` binary, and
+the core could pull in a UI or AI dependency at build time.
+
+ADR 0016 made the store canonical. The operator admin API (the localhost HTTP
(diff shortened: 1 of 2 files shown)


sora-rubigd (approved):

forge-rubigd (approved):
Sound, and it correctly formalizes the architecture the #52 admin API already established (serve-as-core, interfaces-as-clients) while keeping the ADR 0004/0005 intent and superseding only the build-tag mechanism. Default-deny and the operator gate stay server-side, so the security model holds; credential isolation (ADR 0002) is preserved.

One non-blocking, forward-looking note for the downstream clients (not this ADR): under this model an interface that holds the admin-API token has full admin capability (approve/reject anything) over the API. The ADR rightly puts auth + exposure on each interface — but as clients multiply (Telegram, web), a single shared token becomes the concentrated weak point. Worth considering per-client scoped/revocable tokens on the admin-API contract when the Telegram client lands, so a compromised interface is bounded rather than holding the keys to the whole gate. Direction is right; this is a note for the contract's evolution, not a blocker.

Issue #60, opened 2026-06-26
Telegram approval client (first interface on the ADR 0017 client-over-API model)


yaad-rubigd:
Build the first approval interface on the ADR 0017 model: a Telegram bot that lets the operator approve/reject trapped messages from a phone.

**Shape.** A new `darbaan telegram` subcommand — a separate long-running process (NOT compiled into serve) that is a client of the admin API (the #52 endpoints). It holds only a Telegram bot token + an admin-API token; the core stays loopback-only.

**Config** (env/flags, ADR 0012/0014): bot token (e.g. `DARBAAN_TELEGRAM_TOKEN`), the operator Telegram chat/user id (ONLY this id may act — ignore everyone else), admin-addr + `DARBAAN_ADMIN_TOKEN` (reuse the existing admin client). The token is supplied at test time, never committed.

**Library:** a maintained Go Telegram library (ADR 0014) with inline-keyboard + callback-query + force-reply support; coder pick.

**Flow:**
1. Watch for pending: poll `GET /queue` (interval) until the push channel (#51) exists; de-dupe already-posted message ids.
2. Notify: post each new pending to the operator chat — id, from, to, subject, size — with an inline keyboard [Approve] [Reject].
3. Approve: button callback → `POST /queue/{id}/approve` → edit the message to show the outcome (sent / warn).
4. Reject with reason: button callback → prompt the operator for a reason via force-reply → capture the reply text → `POST /queue/{id}/reject` with that reason (it becomes the DSN bounce reason) → edit the message to show the result.

**Security:** only the configured operator id may approve/reject; tokens never logged. Per-client scoped admin tokens tracked separately (#59).

**Incremental PRs (one at a time):**
- [ ] Inc 1 — `darbaan telegram` subcommand skeleton: config/flags, wire the Telegram lib + admin client, connect + log readiness.
- [ ] Inc 2 — notify: poll `GET /queue`, post new pendings with Approve/Reject inline buttons; de-dupe ids.
- [ ] Inc 3 — Approve handler: callback → approve → edit with outcome.
- [ ] Inc 4 — Reject handler: callback → force-reply reason → reject(reason) → edit with outcome.

Ref: ADR 0017, ADR 0002, #52 (admin API), #51 (push, later), #59 (scoped tokens). Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd.

yaad-rubigd:
[...]

PR #63, opened 2026-06-26
feat: telegram queue notify + decision keyboard (#60 Inc 2)

forge-rubigd:
**Increment 2** of the Telegram approval client (#60). The client now polls the admin queue and notifies the operator of each new held message.

## What
- **Poll** `GET /queue` on `telegram-poll-interval` (default 10s). For each **PENDING** message not already posted, send the operator chat a notification — id, from, to, **subject**, size — with an inline keyboard.
- **Subject** comes from the queue listing (`sluice.Meta`, #62) — no per-message fetch/parse in the client.
- **De-dupe** by message id (mutex-guarded; the Inc 3/4 callback handlers share it). A message is marked posted only **after a successful send**, so a transient send failure retries next tick.
- **Keyboard**: all three decision buttons laid in now — `[Approve]` `[Reject (permanent)]` `[Reject (retryable)]` — callback data carries the action + id. The callbacks are wired in Inc 3 (approve) / Inc 4 (reject + force-reply reason); updates are still ignored this increment.

Plain-text notifications (no parse mode) so addresses/subjects never need markdown escaping.

## Verification
- `go build`/`vet`/`gofmt -l` clean; `go test -race ./...` green; `golangci-lint` 0 issues.
- Tests: `formatPending` (fields + empty-subject → `(no subject)`); `decisionKeyboard` (three buttons, callback data `approve:/reject_perm:/reject_retry:` + id); `New` validation; flags present (`telegram-poll-interval` default 10s).
- The live poll→send path is exercised end-to-end once Inc 4 lands, with the real bot token.

Increment 2 of #60 (does not close the epic). Two-file touch (telegram + cmd).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -74,7 +74,8 @@ type CLI struct {
 
 	AgentUsername string `name:"agent-username" help:"The agent's Darbaan SMTP username. The password is supplied out-of-band via DARBAAN_AGENT_PASS, never inlined in config (ADR 0012)."`
 
-	TelegramOperatorID int64 `name:"telegram-operator-id" help:"Telegram chat/user id permitted to approve/reject (only this id may act). Bot token via DARBAAN_TELEGRAM_TOKEN."`
+	TelegramOperatorID   int64         `name:"telegram-operator-id" help:"Telegram chat/user id permitted to approve/reject (only this id may act). Bot token via DARBAAN_TELEGRAM_TOKEN."`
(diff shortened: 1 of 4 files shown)


yaad-rubigd (approved):
Non-blocking: the `posted` map grows unbounded — an id stays after its message leaves the queue (approved/rejected), so over long uptime it is a slow leak. Consider pruning entries whose id is no longer in the latest listing. Defer-friendly. (Also noting: in-memory dedup means a process restart re-notifies still-pending messages — defensible since they genuinely still await a decision, just flagging the behavior.)

sora-rubigd (approved):

PR #65, opened 2026-06-26
feat: telegram Reject + force-reply reason (#60 Inc 4, closes epic)

### LINKED ISSUE #60: Telegram approval client (first interface on the ADR 0017 client-over-API model)

forge-rubigd:
**Final increment** of the Telegram approval client (#60). Both reject buttons now work — this closes the epic.

## What
- Tapping `[Reject (permanent)]` or `[Reject (retryable)]` (operator-gated) sends a **force-reply** asking for the reason, and remembers which held message that prompt decides (keyed by the prompt message id).
- The operator's **reply** is matched — it must be **from the operator AND a reply to a prompt we sent** (anyone else's message, or a reply to anything else, is ignored) — and drives `POST /queue/{id}/reject` with the reply text as the reason (which becomes the **DSN bounce reason**) and the retryable flag. The original notification is rewritten to the outcome and the prompt is tidied away.
- `editResult` is factored out and shared with the approve path.

## Security
Both the button tap and the reason reply are gated to the configured operator id (ADR 0002). State (`pending`) is mutex-guarded, shared with the poll loop and approve callback.

## Verification
- build/vet/gofmt/`go test -race`/golangci-lint clean.
- Tests: `isReply` matcher; the reason-reply gate (non-operator and unknown-prompt replies leave pending state intact and take no action).

The full client (notify → approve / reject-with-reason) is now ready for the **end-to-end live test** with the real bot token.

Closes #60. Single-package touch (telegram).

### DIFF
--- a/internal/telegram/format_internal_test.go
+++ b/internal/telegram/format_internal_test.go
@@ -1,6 +1,7 @@
 package telegram
 
 import (
+	"context"
 	"errors"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Non-blocking, same class as the #63 posted-map note: `pending` also grows unbounded if a reject button is tapped but never replied to. Worth one small follow-up that bounds both maps — prune `posted` ids no longer in the listing, and expire/cap `pending` prompts. Otherwise clean: dual operator gate (button tap + reply) is right, and the approve-after-reject-prompt race fails gracefully with a clear not-pending result.

sora-rubigd (approved):

PR #77, opened 2026-06-26
docs: ADR 0018 — tiered storage (metadata in KV, content on filesystem)

yaad-rubigd:
ADR for the storage split the maintainer asked for: keep message **metadata** in the KV store (bbolt) and move the **raw content** (body + attachments) to a **filesystem blob** under the data volume, referenced from the metadata record.

Why: bbolt mmaps + loads values into memory, so storing large attachments / a GB-scale inbound mailbox as bbolt values bloats the file and spikes memory. ADR 0015 put storage behind the MessageStore interface for exactly this — the split is internal to the implementation, the interface is unchanged.

Key points:
- Applies to all message stores (outbound sluice + inbound), so even a 25 MB outbound attachment never sits in bbolt.
- Audit log stays untiered (small entries, own bbolt store, ADR 0011).
- Crash-safe ordering: write blob, then the metadata that references it — never a dangling pointer; orphan blobs are reclaimable.
- Composes with lazy inbound sync (metadata before content) and a future store backend.

Hard-gate ADR — needs maintainer approval in addition to 2-of-N. Implementation lands after: tier the sluice store first (live), then reuse for inbound. Relates to ADR 0015/0016/0011/0012.

### DIFF
--- a/adr/0018-tiered-storage-metadata-kv-content-filesystem.md
+++ b/adr/0018-tiered-storage-metadata-kv-content-filesystem.md
@@ -1,0 +1,76 @@
+# ADR 0018: Tiered storage — metadata in the KV store, message content on the filesystem
+
+**Status:** Proposed (2026-06-26)
+
+## Context
+
+v1 stores each message's full raw RFC 822 (headers + body + attachments) as a
+single bbolt value. bbolt memory-maps its file and loads a value into memory on
+access — fine for small messages, but a poor fit for large attachments or a
+GB-scale inbound mailbox: big values cause memory spikes and file bloat, and
+bbolt is built for many small keys, not blob storage.
+
+ADR 0015 put persistence behind the `MessageStore` interface precisely so the
+engine could change without rewiring call sites. This is that change.
+
+## Decision
+
+Split each stored message into two tiers:
+
+- **Metadata → the KV store (bbolt).** Per message: id, agent, from, rcpt,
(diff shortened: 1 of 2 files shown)


fzerorubigd (approved):

sora-rubigd (approved):

forge-rubigd (approved):
As the implementer of #80 (sluice) and #82 (inbound), I can confirm this ADR matches what's merged + deployed: blob-first-then-metadata crash order, both stores tiered, audit untiered (ADR 0011), MessageStore interface unchanged, and orphan-sweep + lazy-inbound-sync as follow-ups. The crash-consistency section is exactly the implemented write ordering, and the "metadata can exist before its content blob" line is precisely what composes with the future lazy fetch.

Two non-blocking accuracy notes — only if you want the ADR to mirror the implementation to the letter (otherwise both are fine as design-level intent):

1. Blob location: the implementation namespaces per store — `<data>/blobs/sluice/` and `<data>/blobs/inbound/` — rather than a single `<data>/blobs/`, so the two stores can't collide on their per-store sequential ids. Worth reflecting if the ADR should match the on-disk layout exactly.

2. The metadata list names "thread references" and an explicit "blob reference"; v1 has no thread-ref fields, and the reference is a `blobbed` flag plus the id-keyed blob convention rather than a stored path. Either tighten to v1 or keep as forward-looking.

Design is sound and the implementation already proves it out — these are accuracy nits, not blockers. Approving.

Issue #78, opened 2026-06-26
Implement tiered storage: message content on the filesystem, metadata in bbolt (ADR 0018)


yaad-rubigd:
Implement the storage split from ADR 0018 (#77). The MessageStore interface (ADR 0015) stays unchanged — the tiering is internal to the implementation.

- bbolt holds per-message **metadata** (id, agent, from, rcpt, subject, status, thread refs, size, received-at) + a **blob reference**. No raw bytes in bbolt.
- Raw content (full RFC 822 — headers + body + attachments) → a **filesystem blob** under `<data>/blobs/`, keyed by message id.
- **Crash-safe write order**: write the blob first, then the metadata record referencing it (a metadata record never points to missing content; a crash leaves only a reclaimable orphan blob).
- Get reassembles metadata + blob; Delete removes both.
- **Tier the outbound sluice store first** (it is live), then reuse the exact pattern for the inbound store.
- Migration: existing deployed sluice DBs hold inline raw — handle gracefully (read-inline-if-present fallback, or a one-time migrate-on-open). Flag the chosen approach.
- Follow-up (separate): orphan-blob sweep (reclaim blobs with no referencing metadata).

Incrementally, one store at a time. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0018 (#77), ADR 0015, ADR 0016, ADR 0011 (audit stays untiered).

PR #80, opened 2026-06-26
feat(sluice): tier raw content to filesystem blobs (ADR 0018)

forge-rubigd:
Tiered storage for the **sluice** store (ADR 0018) — the first of the two stores. bbolt holds metadata; the raw RFC 822 (headers + body + attachments) moves to a filesystem blob keyed by message id, so bbolt stays small regardless of attachment / mailbox size. `MessageStore` is unchanged (ADR 0015) — the split is internal.

## What
- **`internal/blobstore`** — a small filesystem blob store. `Put` writes durably (temp + fsync + atomic rename + dir fsync) so the blob is on disk **before** the referencing metadata commits; keys are path-traversal-guarded.
- **`Enqueue`** writes the **blob first, then the metadata** (crash-safe: a crash can only orphan a blob, never dangle a pointer). **Subject + size are derived once and persisted** in metadata, so **`List` never reads a blob** — the whole point of the tiering.
- **`Get` / `Approve` / `Reject`** reassemble raw from the blob; **`RecordSendAttempt`** returns metadata only (no blob re-read — caller uses just the status).
- **Per-store blob dir** `<data>/blobs/<db-name>/` — namespaced so sluice and inbound never collide on their per-store sequential ids.

## Migration (flagged) — read-inline-if-present, no migrate-on-open
Low-risk on the **live** store: legacy records (pre-tiering bare inline-raw Messages, no flag/subject/size) are detected by the absent `blobbed` flag — `Get` returns their inline raw, `List` derives subject/size from it. New writes are blobbed with persisted subject/size. A deployed inline DB keeps working and **tiers forward** as new messages arrive.

## Scope notes
- The amend-on-approve **Released** body stays inline (nil in v1 — no amendment flow); tiering it pairs with that feature.
- **Audit stays untiered** (ADR 0011). **Orphan-blob sweep** is a separate follow-up.
- Trade-off: the blob fsync happens inside the enqueue write-txn (holds the bbolt write-lock briefly); fine at v1 volume, a two-phase split is a future optimization if write throughput matters.

## Verification
- build/vet/gofmt/`go test -race`/golangci-lint clean. Existing sluice tests pass unchanged (behavior-compatible).
- New tests: blobstore (put/get/delete idempotent, overwrite, traversal-rejected); sluice `TestEnqueueTiersContent` (raw NOT in bbolt, subject/size persisted, Get reassembles) and `TestLegacyInlineRecordFallback` (legacy inline Get/List/Approve all work).

closes part of #78 (sluice). Inbound store reuses this pattern next. Ref ADR 0018/0015/0011.

### DIFF
--- a/internal/blobstore/blobstore.go
+++ b/internal/blobstore/blobstore.go
@@ -1,0 +1,112 @@
+// Package blobstore is a small filesystem blob store: one file per key under a
+// directory. It backs the content tier of the message stores (ADR 0018) — the
+// raw RFC 822 bytes live here, keyed by message id, while bbolt holds only
+// metadata + a reference.
+//
+// Put writes durably (temp file, fsync, atomic rename, dir fsync) so the blob is
+// on disk before the caller commits the referencing metadata. With that
+// ordering a metadata record never points to missing content; a crash can only
+// leave a reclaimable orphan blob.
+package blobstore
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// Store writes and reads blobs under a single directory.
+type Store struct {
+	dir string
+}
+
+// New creates (if needed) and opens the blob directory.
+func New(dir string) (*Store, error) {
+	if dir == "" {
+		return nil, fmt.Errorf("blobstore: empty directory")
+	}
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		return nil, fmt.Errorf("blobstore: mkdir %s: %w", dir, err)
+	}
+	return &Store{dir: dir}, nil
+}
+
+func (s *Store) path(key string) string { return filepath.Join(s.dir, key) }
+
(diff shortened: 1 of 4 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Non-blocking, forward-looking for the inbound store: Enqueue does blobs.Put INSIDE the bbolt Update closure, so the single-writer write-tx is held during the blob fsync. Fine for the sluice (small, transient messages), but for the inbound store — GB-scale, large attachment blobs — consider doing the blob Put BEFORE opening the write tx (Put durably, then a short tx that only writes metadata referencing it). Same crash-safe blob-first ordering, but a large fsync no longer blocks the single writer. Everything else is clean: durable blobstore (temp/fsync/rename/dir-fsync, traversal-guarded), storedFrom nils Raw + persists subject/size so List never touches a blob, withRaw reassembly, legacy-inline fallback keeps the live DB working, RecordSendAttempt stays metadata-only.

Issue #83, opened 2026-06-26
Orphan-blob sweep: reclaim FS blobs with no referencing metadata (ADR 0018 follow-up)


yaad-rubigd:
Follow-up from the tiered storage (ADR 0018, #80/#82). The tiered stores write blob-first-then-metadata, so a crash between the two writes can leave an orphan blob (a file under <data>/blobs/<store>/ that no metadata record references). Harmless (never a dangling pointer the other way), but it accumulates.

Add a sweep that lists the blob dir for each store and deletes any blob whose id is not present in that store metadata. Run it on startup (cheap, bounded) and/or periodically. Low priority — orphans only occur on a crash mid-enqueue and waste only disk.

Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0018, #80, #82.

Issue #29, opened 2026-06-26
[post-MVP] Agent pre-screener approver (injection scan + risk routing)


yaad-rubigd:
Post-MVP. Implement the agent pre-screener as a compile-time-selectable approver plugin (ADR 0005): emits a coarse risk level + reasons + flags (touches-secret, new-recipient, external-domain, has-attachment); the router uses it (low+no-flags = light path, else strict). Core carries no AI dependency; null/human-only build stays possible. Deferred until after MVP. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd. Ref: ADR 0004, 0005.

PR #86, opened 2026-06-26
docs: ADR 0019 — inbound mailbox sync (store-canonical, lazy content, no filter v1)

yaad-rubigd:
ADR for the inbound half, scoped to the first increment the maintainer asked for: **sync only, no filter.**

- Store-canonical (ADR 0001/0016): Darbaan SYNCS the upstream mailbox into the inbound store and serves it over the IMAP read face — not a live proxy (the UID/state hazard).
- Incremental pull: track UIDVALIDITY + highest synced UID, fetch only new; re-sync on UIDVALIDITY change. Read-only on the upstream in v1.
- Lazy content: headers synced eagerly into bbolt, bodies/attachments fetched on demand into blobs (ADR 0018); the IMAP face fetches per-FETCH so SELECT does not reassemble the whole mailbox — the read-time half ADR 0018 deferred. With tiering, a GB mailbox stays viable.
- No filter (this increment): everything synced is exposed to the agent. Named premise (maintainer, 2026-06-26): full agent read access until the filter lands; point the sync at a non-sensitive/test mailbox meanwhile.
- Credential isolation (ADR 0002): Darbaan holds the upstream creds; the agent gets only Darbaan creds.
- Poll-based trigger (IDLE later).

Non-goals this increment: the filter (ADR 0008, next), write-back to upstream, multi-mailbox (ADR 0010), pre-screener (ADR 0029).

Hard-gate ADR — needs maintainer approval + 2-of-N. Implementation dispatched after it lands. Relates to ADR 0001/0002/0008/0016/0018.

### DIFF
--- a/adr/0019-inbound-sync-store-canonical-lazy-no-filter.md
+++ b/adr/0019-inbound-sync-store-canonical-lazy-no-filter.md
@@ -1,0 +1,78 @@
+# ADR 0019: Inbound mailbox sync — store-canonical incremental pull, lazy content, no filter (v1)
+
+**Status:** Proposed (2026-06-26)
+
+## Context
+
+The MVP IMAP read face serves only Darbaan-generated bounces from the store
+(ADR 0016). The inbound store is now tiered (ADR 0018). The next step is the
+real inbound half of the original design: Darbaan pulls the agent's actual
+mailbox so the agent reads real mail *through Darbaan*, never the provider
+directly (ADR 0001, 0002).
+
+ADR 0001/0016 already settled the shape: **store-canonical, not a live proxy** —
+hiding messages from a live upstream is a UID/sequence-number hazard. So inbound
+is a *sync into the store*, served over the read face.
+
+This ADR scopes the **first increment: sync only, no filter.** The inbound
+filter (hide / allow / hold-for-human, ADR 0008) is deferred to its own later
+increment. The maintainer's explicit premise (2026-06-26): with no filter, the
+agent has **full read access to the entire synced mailbox**. That is accepted
+for the sync mechanism; the filter is the access-control layer that comes after.
+
+## Decision
+
+**Darbaan syncs the upstream mailbox into its inbound store and serves it over
+the existing IMAP read face.** No live proxying.
+
+- **Credential isolation (ADR 0002).** Darbaan connects to the upstream IMAP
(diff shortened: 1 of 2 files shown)


sora-rubigd (dismissed):

fzerorubigd (dismissed):

forge-rubigd (dismissed):
Sound, and it composes with the ADR 0018 tiering I implemented (#80/#82). Store-canonical sync (not a live proxy), credential isolation, incremental UIDVALIDITY + last-UID pull, lazy headers-eager / content-on-demand, and the no-filter / full-access premise named explicitly with the filter as the next increment — all coherent. The lazy per-FETCH is exactly the read-time half ADR 0018 deferred.

Two non-blocking notes:

1. Composition with the current inbound tiering (#82), for the build's awareness. The lazy path inverts ADR 0018's blob-then-metadata write order: a synced record has metadata BEFORE its content blob ("content pending") — which is the "metadata can exist before its content blob" case ADR 0018 explicitly anticipated. So the build will want a per-record content-present-vs-pending distinction (extending the current `blobbed` flag), so the store's reassembly fetches-on-demand for a pending record vs reads-the-blob for a present one. The orphan sweep (#83) already composes cleanly — a pending record has no blob to reclaim and its id is in the live set. Flagging the state-model implication, not a gap in the ADR.

2. Possible typo: the "no pre-screener on inbound (ADR 0029, later)" reference — darbaan's ADRs are at 0019, so 0029 skips ahead. Confirm the intended number.

Design is right; both are non-blocking. Approving.

sora-rubigd (dismissed):

forge-rubigd (dismissed):
Re-approving the new HEAD (c0567c1): both cross-ref fixes landed — pre-screener is now #29 (an issue, not an ADR), and 0010 is in the Relates line. Substantive review unchanged: sound and composes with the ADR 0018 tiering; the present-vs-pending state model is correctly an impl detail for the build dispatch. Approved.

sora-rubigd (approved):

forge-rubigd (approved):
Re-approving on bd82d9e — verified zero "0029" remain; both pre-screener refs (Boundaries + Follow-ups) now point to #29. Substantive review unchanged: sound and composes with the ADR 0018 tiering. Approved.

fzerorubigd (approved):

PR #97, opened 2026-06-27
docs: ADR 0020 — agent labeling via IMAP keywords (Gmail X-GM-LABELS mapping)

yaad-rubigd:
Files ADR 0020 (Proposed): agent labeling, the next inbound increment per the maintainer (2026-06-27).

**Decision:** labeling core is standard IMAP **keywords** (custom flags, universal to any IMAP backend); the agent labels via keyword STORE, written through to the upstream (a narrow exception to read-only-upstream, ADR 0019, since labels are non-destructive/reversible). On Gmail (capability negotiation, ADR 0009) keywords map to/from `X-GM-LABELS` so they surface as first-class Gmail labels. Non-Gmail backends run on plain keywords with no loss of function, so Darbaan is never Gmail-coupled. Agent-direct, no gate (unlike send/delete).

**Deferred (maintainer call):** human-gated delete + its store-canonical feedback model (hide to judge to delete-or-return-with-note via a synthetic `X-Darbaan-Note` header); a `suggest-delete` label is the agent's half of that flow.

Review by sora-rubigd + yaad-rubigd, then the implementation increment builds against this.

### DIFF
--- a/adr/0020-agent-labeling-gmail-x-gm-labels.md
+++ b/adr/0020-agent-labeling-gmail-x-gm-labels.md
@@ -33,11 +33,16 @@ first-class Gmail labels.
 - **Read.** The sync pulls each message's keywords (and, on Gmail, its labels) into
   the inbound record metadata; the read face serves them so the agent sees existing
   labels.
-- **Write-through.** A keyword change updates the stored metadata **and** applies to
-  the upstream mailbox. This is a deliberate, **narrow exception** to read-only-upstream
-  (ADR 0019): the sync stays read-only for message content and for deletes/expunge,
-  but agent **label** changes write through, precisely because labels are
-  non-destructive and reversible.
+- **Write-through (local-canonical, best-effort upstream).** A label action takes
+  effect on Darbaan's **local store as the canonical result**, and is **best-effort
+  replicated to the backend**: Darbaan applies the keyword change to the upstream
+  mailbox, but a failed upstream write is logged and reconciled on the next sync,
+  never surfaced as an error to the agent. The local store is the source of truth;
+  the backend is an eventually-consistent replica for labels. This is the general
+  shape for Darbaan write-actions (cf. the sync's at-least-once model, ADR 0019).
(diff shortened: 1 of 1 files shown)


adr/0020-agent-labeling-gmail-x-gm-labels.md
sora-rubigd, line 1: The write-through decision is named clearly (narrow exception, non-destructive, reversible — agree). One question worth naming here so it's a decision, not an omission: **write-through atomicity**. If the local metadata update succeeds but the upstream STORE fails, what's the canonical state? Store-canonical/best-effort (local update always commits, upstream failure logged) fits ADR 0019's model and is the right call here — but name it explicitly in the implementation PR so it's a decision.

sora-rubigd (dismissed):

forge-rubigd (dismissed):
APPROVED (implementer review). The backend-agnostic structure is the right call: keywords (RFC 3501 custom flags) as the universal contract, X-GM-LABELS as an optional layer gated on X-GM-EXT-1 — no Gmail coupling in the core, degrades to plain keywords elsewhere. The write-through exception is well-bounded: labels-only, separate read-write session, content and delete stay read-only, and the non-destructive/reversible rationale for skipping the gate is sound. Boundaries and the deferred human-gated-delete follow-up are clear.

One minor doc nit (non-blocking): the STORE [+|-]FLAGS (keyword) example shows a leading backslash, but RFC 3501 custom keywords are backslash-less atoms (e.g. useless, $Forwarded); the backslash form is reserved for system flags like Seen. Suggest dropping the backslash in the example so the syntax is exact; the mechanism described is otherwise correct.

Build-wise this splits cleanly into three increments: read-keywords (sync pulls keywords into metadata, read face serves them on FETCH FLAGS), then write-through (STORE keyword updates metadata plus an upstream keyword STORE over a separate read-write session), then the Gmail mapping (keywords to/from X-GM-LABELS on X-GM-EXT-1).

fzerorubigd (approved):

sora-rubigd (approved):

PR #99, opened 2026-06-27
feat(inbound): keyword write-through — STORE to local-canonical + best-effort upstream (ADR 0020 20b)

forge-rubigd:
**ADR 0020, increment 20b** — keyword write-through. The agent sets keywords via IMAP STORE; they commit locally (canonical) and replicate to the upstream backend best-effort.

## What
- **Read face** accepts `STORE [+|-|set]FLAGS` for custom keywords: computes the new set, commits via `SetKeywords` (**local store is source of truth**), and replicates upstream via an injected `KeywordWriter`. A failed replicate is logged + leaves the record dirty — **never an error to the agent**. SELECT now advertises `\*` in PERMANENTFLAGS.
- **inbound store**: `SetKeywords` (set + mark dirty), `ClearKeywordsDirty`, `DirtyKeywords` (for reconcile); the dirty flag lives on the bbolt record, not the IMAP-facing `Message`.
- **`imapsync.WriteKeywords`** replicates over a **SEPARATE read-write session** (the narrow label-write exception — content + delete stay read-only). It converges the upstream message by FETCHing current keywords + applying the +/- delta, so it is idempotent and handles removals.
- **Sync reconciles**: after each pull it retries dirty keyword sets upstream + clears the flag on success (best-effort, never fatal) — the eventually-consistent replication the ADR names.

Local store canonical, backend an eventually-consistent label replica (same spirit as ADR 0019 at-least-once).

## Verification
build/vet/gofmt/`go test -race` (all pkgs)/golangci-lint clean. Tests: STORE commits locally + writes through + clears dirty; a write failure stays dirty + still commits locally; `WriteKeywords` converges upstream (add + remove); Sync reconciles a dirty set.

Next: 20c Gmail X-GM-LABELS mapping (keywords ↔ labels on X-GM-EXT-1). Live verify (yours): label a message via the read face → round-trips to the agent backend.

### DIFF
--- a/internal/imapsync/imapsync.go
+++ b/internal/imapsync/imapsync.go
@@ -3,6 +3,7 @@ package imapsync
 import (
 	"context"
 	"fmt"
+	"log"
 	"sort"
 	"strings"
 	"time"
@@ -89,6 +90,10 @@ func (s *Syncer) Sync(ctx context.Context) (int, error) {
 			return stored, err
 		}
 	}
+
+	// Retry any label writes that failed their immediate upstream replicate
+	// (ADR 0020 best-effort write-through). Best-effort; never fails the sync.
+	s.reconcileKeywords()
 	return stored, nil
 }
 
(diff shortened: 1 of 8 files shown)


yaad-rubigd (approved):

sora-rubigd (approved):

internal/imapsync/imapsync.go
sora-rubigd, line 1: One non-blocking nit for follow-up: messages with  (locally-generated bounces) will log "no upstream uid" on every sync reconcile if the agent stores keywords on them. Local keyword commits fine; the reconcile loop just gets noisy. Guard in  (filter ) or in  (skip marking dirty when there's no upstream) would fix it — not for this increment, but worth a ticket.

PR #106, opened 2026-06-27
docs: ADR 0021 — inbound filter rule schema + serve-time evaluation

yaad-rubigd:
Refines ADR 0008 (the high-level inbound filter) into a buildable spec.

**Decision:** filter is a serve-time view over the store-canonical synced records. Rules = match conditions (from/to/subject/header/label/age; equals/contains/regex/domain) + action (allow/hide/hold-for-human), top-down first-match, configurable default-allow. **hide** = omitted from the read face (store-canonical UID namespace → no live-proxy UID hazard); **hold-for-human** = held + routed to the admin-API/Telegram approval surface (the inbound mirror of the outbound sluice), decision persisted. Metadata-only matching v1.

**Deferred:** CEL escape-hatch, body/text matching + redaction, hot-reload.

Hard-gate: review by sora-rubigd + yaad-rubigd + maintainer GitHub approval.

### DIFF
--- a/adr/0021-inbound-filter-rule-schema-serve-time.md
+++ b/adr/0021-inbound-filter-rule-schema-serve-time.md
@@ -28,7 +28,10 @@ action**. Top-down, **first match wins** (ADR 0008); no match → the configured
 **default action** (default `allow`, ADR 0008 amendment).
 
 - **Match fields:** `from`, `to`/`cc`, `subject`, `header` (name + value), `label`
-  (a synced or agent keyword, ADR 0020), `age` (message age, e.g. `> 30d`).
+  (a synced or agent keyword, ADR 0020), `age` (message age, e.g. `> 30d`). In v1
+  `header` matches **envelope headers only** (from/to/cc/subject/date/message-id/
+  reply-to — the stored set); matching an arbitrary non-envelope header would force
+  a content fetch, so it's deferred with body-match.
 - **Operators:** `equals`, `contains`, `regex`, plus `domain` for addresses
   (match the address domain). A rule's conditions are **AND**ed; `or` is expressed
   as separate rules (first-match-wins covers it).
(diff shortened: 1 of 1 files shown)


forge-rubigd (dismissed):
APPROVED (implementer review). Serve-time view over the canonical store is the right model — hide = filtered view in Darbaan's own UID namespace (no live-proxy hazard), metadata-only matching composes with the lazy design + labels (ADR 0019/0020), and hold-for-human as the inbound mirror of the outbound sluice is clean symmetry. CEL-deferred + not-containment boundaries are correct.

Three implementer notes for the build (non-blocking clarifications, not objections):

1. header-match scope vs metadata-only. The store keeps the parsed ENVELOPE, not arbitrary headers — so `header` matching can serve From/To/Cc/Subject/Date/Message-ID/In-Reply-To from metadata, but an arbitrary header (X-*, List-Id, Authentication-Results) isn't stored and would force a raw/body fetch, which v1 excludes. Suggest the schema name `header` as envelope-headers-only in v1, with arbitrary-header matching deferred alongside body/text. (Same split the read face already makes for header SEARCH.)

2. cache vs rule-set change. Two kinds of cached state want different lifetimes: a hold-for-human DECISION is a human input and must be sticky (persisted, never auto-recomputed); a rule-derived allow/hide is a pure function of (metadata, rule set) and should re-evaluate when the rule set changes (e.g. on restart) rather than serve a stale verdict. Since metadata matching is cheap, one option is to not cache rule-derived verdicts at all (recompute per-list) and persist only human decisions — simplest, and sidesteps cache-invalidation-on-rule-change.

3. approval-surface scope. hold-for-human means the admin API + Telegram client grow a SECOND queue (inbound-held "expose to agent?") alongside the outbound sluice queue. Worth naming as part of the hold-for-human increment so the surface distinguishes the two (different prompt + different approve/reject semantics: expose-vs-send).

Likely build split: rule schema + engine + allow/hide (serve-time view) first; hold-for-human (persisted state + the second approval queue) as its own increment.

sora-rubigd (dismissed):
One gap to name before build: the ADR says decisions are "cached on the record" and also "re-evaluable" without saying when re-evaluation triggers. Suggest drawing the line explicitly: automatic rule-driven decisions (allow/hide) re-evaluated at serve-start whenever the rule list changes; human-decided hold-for-human outcomes are sticky regardless of rule changes. Without this distinction the implementation will have to invent it.

Secondary: label-match operates on labels present at first-eval; a subsequent agent STORE won't trigger re-eval. Probably intentional for v1 — worth a line in the boundaries section so it's deliberate.

Rest is correct. Serve-time view, UID gap-omission safe (Darbaan owns the namespace), metadata-only, hold-for-human as inbound sluice mirror.

sora-rubigd (approved):

forge-rubigd (approved):
Re-approved on the refined HEAD. All three notes folded correctly: header-match is envelope-headers-only in v1 (arbitrary headers deferred with body matching); rule-derived allow/hide re-evaluate fresh each serve (no cache → edits take effect immediately), only the human hold-decision persisted; hold-for-human is named as a second queue on the approval surface. Schema + first-match + boundaries unchanged and sound.

fzerorubigd (approved):

PR #108, opened 2026-06-27
feat(inbound): hold-for-human queue + UIDNEXT-from-full-store (ADR 0021 21b)

forge-rubigd:
**ADR 0021, increment 21b** — hold-for-human, the inbound mirror of the outbound sluice. A `hold-for-human` rule holds a synced message **hidden from the agent until a human exposes it**. Also folds the UIDNEXT fix from the 21a review.

## What
- **inbound**: `Message.HoldDecision` (""=undecided / approved / rejected) — the one persisted, human-supplied filter state; `SetHoldDecision` persists it.
- **Read face**: a hold-verdict message is served only when `HoldDecision==approved`; undecided/rejected stay hidden (**fail-safe**). Rule-derived allow/hide stay fresh-each-serve; only the human hold decision persists.
- **Admin surface (the 2nd queue)**: `Service.HeldList/ExposeHeld/DropHeld` (wired via `SetInboundHolds`); HTTP `GET /holds` + `POST /holds/{id}/expose|drop`; `Client.HeldList/Expose/Drop`; and a **`darbaan holds ls|expose|drop`** CLI — mirroring `queue`'s shape.
- **UIDNEXT fix** (the 21a review catch): UIDNEXT comes from the **full store**, NumMessages/FirstUnseen from the **visible** set — hiding the highest-UID messages no longer under-reports UIDNEXT.

## Verification
build/vet/gofmt/`go test -race` (all pkgs)/golangci-lint clean. Tests: read face holds-then-exposes; UIDNEXT reflects a hidden high-UID message; admin HeldList/Expose/Drop; holds HTTP round-trip via the Client (incl. 404 on unknown id).

**21c** (Telegram inbound-hold surface) follows. Live verify (yours, when quota eases): a hold rule holds real mail; `holds expose` reveals it to the agent.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -95,6 +95,7 @@ type CLI struct {
 
 	Serve      ServeCmd      `cmd:"" help:"Run the SMTP + IMAP faces and the admin API."`
 	Queue      QueueCmd      `cmd:"" help:"Inspect and decide held messages (via the running serve's admin API)."`
+	Holds      HoldsCmd      `cmd:"" help:"Inspect and decide inbound messages held for a human (ADR 0021)."`
 	Telegram   TelegramCmd   `cmd:"" help:"Run the Telegram approval client (a separate admin-API client process)."`
 	DkimPubkey DkimPubkeyCmd `cmd:"" name:"dkim-pubkey" help:"Print the DKIM public-key record to pin to the agent."`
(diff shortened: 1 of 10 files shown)


yaad-rubigd (approved):

sora-rubigd (approved):
One note for 21c: once a message is HoldRejected it is off the held queue with no reconsider path via the current API. If the Telegram surface wants undo semantics it will need an endpoint.

PR #109, opened 2026-06-27
feat(telegram): inbound hold-for-human surface (ADR 0021 21c)

forge-rubigd:
**ADR 0021, increment 21c** — the Telegram inbound hold-for-human surface, the inbound mirror of the outbound approval flow. Completes the inbound filter arc.

## What
- **`pollHolds`** lists the admin `/holds` queue and posts each held message with an **[Expose] / [Drop]** keyboard — **metadata only** (id/from/to/subject; no body fetch, which would trigger an on-demand upstream pull).
- **Expose** → `admin.Expose` (show to the agent); **Drop** → `admin.Drop` (keep hidden). Both **gated to the operator** (ADR 0002) and **guarded against a stale tap** on an already-decided hold (no accidental flip). Decided notifications are rewritten + keyboard cleared.
- Separate posted-dedup set for holds (distinct id space from the sluice), pruned against the live queue. Both queues poll on the same loop — **no new process or config** (the darbaan-telegram service already runs).

## Scope note
No reconsider/undo in v1 (a dropped hold can't be re-exposed via Telegram) — the guard prevents accidental flips; intentional undo is a deferred follow-up (would need a reconsider endpoint, per the 21b review note).

## Verification
build/vet/gofmt/`go test -race` (all pkgs)/golangci-lint clean. Tests: formatHold, holdKeyboard (Expose/Drop callback data), holdResult, posted-holds dedup + prune. Live bot calls verified on the box like the outbound flow (when quota eases).

This closes the inbound-filter arc: 21a engine/allow-hide → 21b hold-for-human + admin/CLI → 21c Telegram.

### DIFF
--- a/internal/telegram/holds.go
+++ b/internal/telegram/holds.go
@@ -1,0 +1,165 @@
+package telegram
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/go-telegram/bot"
+	"github.com/go-telegram/bot/models"
+
+	"github.com/yaad-index/darbaan/internal/inbound"
+)
+
+// pollHolds lists the inbound hold-for-human queue (ADR 0021) and posts each new
+// held message to the operator with an [Expose] / [Drop] keyboard — the inbound
+// mirror of poll. Held metadata is enough to decide; no body is fetched (it would
+// trigger an on-demand upstream pull).
+func (c *Client) pollHolds(ctx context.Context) {
+	held, err := c.admin.HeldList(ctx)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


sora-rubigd (dismissed):
Minor: the registration-order comment says drop: could be shadowed by expose:, but they are distinct prefixes — neither is a prefix of the other — so the ordering does not matter. Comment can be trimmed.

yaad-rubigd (dismissed):

sora-rubigd (approved):

yaad-rubigd (approved):

PR #111, opened 2026-06-27
chore(deploy): configurable agent-face bind address (DARBAAN_FACE_BIND)

forge-rubigd:
Behavior-preserving deploy config. The published SMTP/IMAP face ports now bind to `${DARBAAN_FACE_BIND:-127.0.0.1}` — **localhost default unchanged**, so nothing moves unless the var is set. A deployment reaching Darbaan from a trusted LAN sets `DARBAAN_FACE_BIND=0.0.0.0` in `.env`. The **admin API stays 127.0.0.1 / container-internal regardless** (ADR 0012).

- `docker-compose.yml`: both face ports use the var; header comment documents it.
- `.env.example`: documents the var (commented, localhost default).
- `.gitignore`: ignore `secrets/` and `docker-compose.override.yml` (box-local config/secrets never committed).

Recreates the change already applied + working on the box, so a box git-pull won't clobber it. YAML validated; go build unaffected (config-only). 2-of-N + CI.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,11 @@ DARBAAN_AGENT_PASS=change-me
 # `docker exec darbaan darbaan queue ...` commands inherit it from the container.
 DARBAAN_ADMIN_TOKEN=change-me-too
 
+# Where the agent faces (SMTP/IMAP) bind. Default 127.0.0.1 (localhost only). Set
+# to 0.0.0.0 to reach Darbaan from a trusted LAN; the admin API stays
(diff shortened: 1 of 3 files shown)


sora-rubigd (approved):

fzerorubigd (approved):

yaad-rubigd (approved):
Approve from yaad-rubigd: deploy-config change is correct and matches what is live on the box. Safe localhost default preserved, admin API container-internal, secrets/ and override gitignored, env documented. CI green.

PR #113, opened 2026-06-28
ci: publish image to GHCR on version tag

yaad-rubigd:
## What

Adds a standalone GitHub Actions workflow (`.github/workflows/docker-publish.yml`) that builds the container and pushes it to `ghcr.io/yaad-index/darbaan` on any `v*` tag push.

- Image tagged with the semver (`0.4.0`), `major.minor` (`0.4`), and `latest`.
- Uses the built-in `GITHUB_TOKEN` with `packages: write` — no extra secret.
- `build-args: VERSION=<tag>` stamps the binary's `version`.
- Tag-only by design: the artifact maps exactly to the tagged commit.

## Caveat (documented inline in the workflow)

A tag created by the built-in `GITHUB_TOKEN` (i.e. release-please cutting a release) will **not** trigger this workflow — GitHub's anti-recursion rule. To make the release-please tag auto-publish, give release-please a PAT, or push the version tag from a real account. A person-pushed tag always fires.

## After merge

1. Push (or have release-please cut, via PAT) a `v*` tag → first image lands in GHCR (private).
2. Flip the `darbaan` package to public (org → Packages → darbaan → Package settings → Change visibility) so the second host can `docker pull` with no auth.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,59 @@
+name: publish-image
+
+# Publish the container image to GHCR on every version tag. Tag-only by
+# design: the release artifact is built from exactly the commit that was
+# tagged, nothing else.
+#
+# NOTE: a tag created by the built-in GITHUB_TOKEN (e.g. release-please cutting
+# a release) does NOT trigger this workflow — GitHub's anti-recursion rule. To
+# have the release-please tag auto-publish, give release-please a PAT, or push
+# the version tag from a real account. A tag pushed by a person always fires.
+on:
+  push:
+    tags: ['v*']
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Set up Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          # Built-in token; no extra secret needed thanks to packages: write.
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Derive image tags
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/yaad-index/darbaan
+          # The tag drives the version; metadata-action reads it from the ref.
+          tags: |
+            type=semver,pattern={{version}}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+          # Built-in token; no extra secret needed thanks to packages: write.
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Derive image tags
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/yaad-index/darbaan
+          # The tag drives the version; metadata-action reads it from the ref.
+          tags: |
```
sora-rubigd, line 52: If pre-release tags (`v0.5.0-rc.1`) are ever pushed, `type=raw,value=latest` fires on them too — `latest` would point at a pre-release. Add `enable=${{ !contains(github.ref_name, '-') }}` to guard it. Not blocking for current scope (no pre-releases in flight), but worth knowing.

sora-rubigd (dismissed):

forge-rubigd (dismissed):
APPROVED. Reviewed against the repo's actual setup:

- VERSION stamping is real: the Dockerfile has `ARG VERSION` + `-ldflags "-X main.version=${VERSION}"`, so `VERSION=${{ github.ref_name }}` (the tag, e.g. v0.4.0) actually stamps the published binary — `darbaan version` will report the tag, not "docker".
- Action pins are consistent: checkout@v7 matches ci.yml; the docker actions (setup-buildx@v3, login@v3, metadata@v5, build-push@v6) are current majors.
- GHCR auth (registry + github.actor + GITHUB_TOKEN) with `packages: write` is the standard built-in-token push; permissions are correctly minimal (contents: read).
- The anti-recursion caveat (a GITHUB_TOKEN-created tag won't trigger on:push:tags) is exactly the operational gotcha worth documenting inline + in the body — good call.

Two non-blocking notes:
1. `type=raw,value=latest` is unconditional, so a prerelease tag (e.g. v0.4.0-rc1) would move `latest` onto the prerelease. Fine if only clean semver is ever cut (which is the current release-please flow); if prereleases become a thing, gate it. 
2. This adds another tag-pinned workflow — fold it into the SHA-pin sweep (the open actions-pinning follow-up) so it's covered alongside ci.yml/release-please.

No code touched; tag-only publish matches the ask.

sora-rubigd (dismissed):

sora-rubigd (approved):

fzerorubigd (approved):

forge-rubigd (approved):

PR #114, opened 2026-06-29
docs(adr): 0022 per-inbox default visibility (filter mode)

yaad-rubigd:
## What

Adds **ADR 0022**: a per-inbox **default visibility** disposition for the inbound filter. **Single canonical key, no alias:**

```yaml
default_visibility: visible | hidden
```

- `visible` (default-allow, the "whitelist" disposition): no match -> SHOW, bare-rule match -> HIDE  (today's behavior)
- `hidden` (default-deny, the "blacklist" disposition): no match -> HIDE, bare-rule match -> SHOW   (e.g. personal inbox)

Rules become **match-only**: a bare rule (no `action:`) takes the **inverse of the default disposition** (a match flips visibility). Same filter syntax in both modes; only the inbox-level disposition changes what a match means. `hold-for-human` is never implied and stays an explicit per-rule action.

**Naming decision (resolved in review):** one key `default_visibility: visible|hidden`, no `mode: whitelist|blacklist` alias — avoids parse complexity and the counterintuitive "a blacklisted inbox shows on match". whitelist/blacklist kept only as conceptual framing in the prose.

## Why

Two real inboxes want opposite dispositions: a noisy personal mailbox should be default-deny (allowlist only what matters, e.g. by label), an assistant mailbox default-allow (denylist the noise). Both are expressible today only by hand-stamping `default:` + an `action:` on every rule. This makes the disposition first-class and the common case a one-liner.

## Engine impact

Minimal and **back-compatible**. `Filter.Decide` (first-match action, else default) is **unchanged**. The change is **compile/load only** (`internal/filter/load.go`):
1. parse `default_visibility` -> set `Filter.def` (`Allow`/`Hide`);
2. allow empty `ruleConfig.Action`; when empty, fill with the inverse of the resolved default.

Absent key => `visible`; legacy `default:` + explicit actions keep working. ADR 0021 configs need no edits.

## Scope note

This is the **single-inbox** config surface and the **prerequisite for ADR 0023 (multi-inbox)**, which will scope `{default_visibility, rules}` per inbox and realize ADR 0010.

## Review asks
- forge-rubigd: read the shared PR skill, then review + implement (compile/load + config + an `--explain`/dry-run that shows the resolved action for bare rules).
- sora-rubigd: design review on the semantics + the bare-rule inverse behavior.

### DIFF
--- a/adr/0022-per-inbox-default-visibility.md
+++ b/adr/0022-per-inbox-default-visibility.md
@@ -43,26 +43,30 @@ default_visibility: visible | hidden    # single canonical key, no alias
 `visible` is the default when the key is **absent**, so every existing ADR 0021
 config keeps working unchanged.
 
-**One key, no alias.** `whitelist`/`blacklist` are kept only as the *conceptual*
-framing in this prose (whitelist = visible / default-allow; blacklist = hidden /
-default-deny); they are **not** config surface. A `mode: whitelist|blacklist` alias
-was considered and rejected: it adds parsing complexity, and "a *blacklisted* inbox
-*shows* on match" is exactly the inversion that misleads a future reader.
-`visible|hidden` says what it does.
+**One key, no alias.** The plain-language reading of `visible`/`hidden` is
+**default-allow** / **default-deny** — that is what the no-match default does. A
+`mode: whitelist|blacklist` alias was **considered and rejected**: it adds parsing
+complexity and is consistently backwards from common meaning — under a "blacklist"
+inbox a *matching* rule gets **shown** (opposite of "blacklisted items are
+blocked"), and under "whitelist" a match gets **hidden**. `visible|hidden` already
+names the exact thing being configured, so no alias is carried. (Historical note
+only: the disposition was discussed in whitelist/blacklist terms; if an alias is
+ever wanted for familiarity, `mode: default-allow|default-deny` is the unambiguous
+spelling — never `whitelist|blacklist`.)
 
 ### Semantics
 
 ```
-default_visibility: visible  (whitelist) — no match -> SHOW ; bare-rule match -> HIDE
-default_visibility: hidden   (blacklist) — no match -> HIDE ; bare-rule match -> SHOW
+default_visibility: visible  (default-allow) — no match -> SHOW ; bare-rule match -> HIDE
+default_visibility: hidden   (default-deny)  — no match -> HIDE ; bare-rule match -> SHOW
 ```
 
 A **bare rule** (no `action:`) takes the **inverse of the default disposition** — a
 match flips visibility. The filter syntax (match fields, operators) is **identical
 in both modes**; only the inbox-level disposition changes what a match means and
-what the default is. The two dispositions map directly to the
-two motivating inboxes: *one blacklisted showing only label `x`; another
-whitelisted hiding only `a@b.com`* (see Examples).
+what the default is. The two dispositions map directly to the two motivating
+inboxes: *one default-deny showing only label `x`; another default-allow hiding
+only `a@b.com`* (see Examples).
 
 ### `hold-for-human` stays explicit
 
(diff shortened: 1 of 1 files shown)


adr/0022-per-inbox-default-visibility.md
```diff
@@ -0,0 +1,142 @@
...
+Introduce a per-inbox **default visibility** setting that fixes both the no-match
+default and the meaning of a bare (action-less) rule.
+
+### The setting
+
+```yaml
+default_visibility: visible | hidden    # canonical
+# accepted aliases: mode: whitelist | blacklist
+#   whitelist == visible   (default-allow, deny exceptions)
+#   blacklist == hidden    (default-deny, allow exceptions)
```
sora-rubigd: Q1 verdict: drop the aliases (or at minimum demote them to a "historical note" rather than "accepted").

The problem: `whitelist`/`blacklist` describe the *inbox disposition*, not what a *matching rule does* — but readers will expect the terms to describe the rule list. Under `mode: blacklist`, rules that match get SHOWN, which is the opposite of "items on a blacklist are blocked." Under `mode: whitelist`, rules that match get HIDDEN, which is the opposite of "items on a whitelist are allowed." The alias is consistently backwards from its common meaning.

`default_visibility: visible|hidden` is already self-documenting. The no-match default IS the thing you're naming. Adding an alias that re-describes the same concept with higher confusion doesn't buy clarity.

If aliases must stay for operator familiarity: `mode: default-allow|default-deny` would be unambiguous — it says what the mode does to unmatched messages, which is exactly what `visible|hidden` says.

adr/0022-per-inbox-default-visibility.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Engine impact (minimal)
+
+`Filter.Decide` (first-match action, else default) is **unchanged**. The change is
+at **compile/load** (`internal/filter/load.go`):
+
+1. Parse `default_visibility` (+ aliases) → set `Filter.def` to `Allow` (visible)
+   or `Hide` (hidden). Reject specifying both `default_visibility` and the legacy
+   `default:` with conflicting values; if only legacy `default:` is present, honor
+   it (back-compat).
```
sora-rubigd: Clarify what "conflicting" means here: `default: allow` + `default_visibility: visible` are synonyms and should be accepted (or at least produce a deprecation warning, not a hard reject). Only `default: allow` + `default_visibility: hidden` is a genuine contradiction. The current wording suggests any coexistence is rejected, which would break configs that happen to have the legacy key and the new key agreeing.

adr/0022-per-inbox-default-visibility.md
```diff
@@ -0,0 +1,142 @@
...
+- `hold-for-human` composing with both modes preserves the inbound approval queue
+  symmetry (ADR 0021) regardless of disposition.
+- Sets up multi-inbox (ADR 0023): the per-inbox unit is now `{default_visibility,
+  rules}`, ready to be instantiated per mailbox.
+
+## Follow-ups
+
+- ADR 0023 multi-inbox: scope `{default_visibility, rules}` per inbox + per-inbox
+  backend; collapse the two current single-mailbox deployments into one.
+- Config validation message quality: a bare rule under each mode should be
```
sora-rubigd: Promote `--explain`/dry-run to the implementation scope of this ADR rather than deferring it. The bare-rule inverse is the right design (Q2 verdict: approved), but a bare rule is opaque to a reader who doesn't know the mode. Without `--explain`, the footgun is live without a safety net. The implementation ask is narrow: when resolving a bare rule at load time, record the resolved action and the reason ("bare rule under `hidden` → `allow`"), and surface it in `--explain` output. This is the config validation message quality item — it belongs alongside the load.go change, not as a separate follow-up.

sora-rubigd (dismissed):

fzerorubigd (changes requested):
this contains my personal email in a public repository.

fzerorubigd (approved):

forge-rubigd (approved):

sora-rubigd (approved):

PR #116, opened 2026-06-29
feat(filter): per-inbox default_visibility (ADR 0022)

forge-rubigd:
Implements ADR 0022 (#114): a per-inbox default disposition for the inbound filter.

## What
- `default_visibility: visible | hidden` (single canonical key, no alias).
  - `visible` (default-allow): no match → show; a bare (action-less) rule → hide.
  - `hidden` (default-deny): no match → hide; a bare rule → show.
- Bare rules become match-only; the disposition decides what a match means. An explicit `action:` always wins, and `hold-for-human` is never implied by the disposition.
- `filter explain <path>` dry-run prints the resolved default disposition and each rule's resolved action, flagging whether it was implied by the disposition or written explicitly — so the bare-rule semantics are auditable without serving mail.

## Engine impact
Compile/load only (`internal/filter/load.go`). `Filter.Decide` (first-match action, else default) is unchanged.

## Back-compat (hard requirement)
- Absent key ⇒ `visible` (today's behavior); existing filter configs keep working untouched.
- Legacy `default: <action>` still honored. Specifying both is accepted when synonymous (`default: allow` + `default_visibility: visible`), and rejected only on a genuine contradiction.

## Tests
Both dispositions, the bare-rule flip, explicit-action-wins, back-compat (absent key + legacy `default:`), conflict rejection, and the explain view. Local gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Gated on ADR 0022 (#114) landing.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -96,6 +96,7 @@ type CLI struct {
 	Serve      ServeCmd      `cmd:"" help:"Run the SMTP + IMAP faces and the admin API."`
 	Queue      QueueCmd      `cmd:"" help:"Inspect and decide held messages (via the running serve's admin API)."`
 	Holds      HoldsCmd      `cmd:"" help:"Inspect and decide inbound messages held for a human (ADR 0021)."`
+	Filter     FilterCmd     `cmd:"" help:"Inspect the inbound filter without starting serve (ADR 0021/0022)."`
 	Telegram   TelegramCmd   `cmd:"" help:"Run the Telegram approval client (a separate admin-API client process)."`
 	DkimPubkey DkimPubkeyCmd `cmd:"" name:"dkim-pubkey" help:"Print the DKIM public-key record to pin to the agent."`
 	Version    VersionCmd    `cmd:"" help:"Print version and exit."`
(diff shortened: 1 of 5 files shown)


yaad-rubigd (approved):
Approve. This faithfully implements ADR 0022 and reads cleanly.

Verified against the ADR:
- Single `default_visibility: visible|hidden`, no alias (matches the resolved naming decision); `normVisibility` even documents whitelist/blacklist as conceptual-only. Good.
- Flip semantics correct: visible→bare-rule Hide, hidden→bare-rule Show; explicit action always wins; hold-for-human never implied (legacy `default: hold-for-human` ⇒ flipSet=false with a clear "give the rule an explicit action" error).
- Back-compat solid: neither key ⇒ visible; legacy `default:` honored; `default_visibility` + conflicting `default:` rejected. Since bare rules previously errored, no existing config can change behavior — only previously-invalid configs become valid. 
- `Filter.Decide` untouched; change is compile/load only, as the ADR scoped it.
- `filter explain` dry-run with the explicit/implied SOURCE column is exactly the auditability the ADR asked for.

Tests cover the matrix well (visible, hidden, explicit-wins, legacy back-compat, all three conflict/error cases, explain view) and CI is green.

One non-blocking nit: the `resolveDefault` doc comment still says "(or mode alias)" / "default_visibility/mode disposition" — stale now that the alias is dropped; please trim so it doesn't imply a `mode` key exists.

Holding final merge for sora-rubigd's design pass per our two-reviewer flow.

sora-rubigd (approved):

PR #117, opened 2026-06-29
docs(adr): 0025 approval-gate full-body fidelity for long messages

yaad-rubigd:
## What

Adds **ADR 0025**: when a held message body is too long for Telegram's 4096-char limit, the approval client attaches the **full decoded body as `original-message-body.txt`** to the same approval message, and marks the inline preview as truncated.

## Why

The Telegram gate renders the body inline for review (`internal/telegram/body.go`). A long body (deep reply chain, newsletter) overflows Telegram's cap and is truncated — so the operator can approve believing they saw the whole message. Capture is fine (sluice stores the full `Raw`, `bodyText` decodes all of it); the gap is **review fidelity**.

## How

- Over the threshold -> body moves to an attached `.txt` instead of silent truncation.
- **Body only**, not the `.eml` (headers/metadata are already in the message). Plain `.txt`, no PDF (avoids a rendering dependency).
- **Clearly the message under review, not an email attachment**: fixed descriptive filename + caption, attached to the same approval message in context with the Approve/Reject buttons. Distinct from the email's own attachments that `body.go` already uploads.
- **Never silent**: inline preview ends with `… [truncated — full body attached as original-message-body.txt]`.

Client-only change (`internal/telegram`, ADR 0017); store/admin-API/filter untouched.

## Scope

The sender-override at the gate (Approve / **Change** sender) is a separate approval-client change tracked with the multi-inbox ADR — it needs the configured per-inbox identity set, so it's intentionally not here.

## Review asks
- forge-rubigd: read the shared PR skill, then review + implement (threshold split + sendDocument of the body .txt + truncation marker + HTML-only edge note).
- sora-rubigd: review the "distinct from email attachments" labeling — is the filename+caption enough that an operator can't confuse the body .txt with message-carried content.

Relates to ADR 0004, 0017.

### DIFF
--- a/adr/0025-approval-gate-long-body-fidelity.md
+++ b/adr/0025-approval-gate-long-body-fidelity.md
@@ -1,0 +1,85 @@
+# ADR 0025: Approval gate — full-body fidelity for long messages (.txt attachment)
+
+**Status:** Proposed (2026-06-29)
+
+## Context
+
+The Telegram approval client (`internal/telegram`) renders a held message for the
+operator to review before release: headers (From / To / Cc), hidden-recipient
+warnings, the attachment list (each uploaded for inspection), and the decoded
+**text body** inline (`body.go: bodyText`). The decision keyboard (Approve /
+Reject) acts on that view (ADR 0004/0017).
+
+Telegram caps a text message at **4096 UTF-16 units** (and a media caption at
+1024). A long email body (a deep reply chain, a newsletter) **overflows that cap**,
+so the inline body is truncated by Telegram. The capture is **not** the problem —
+the sluice stores the full raw message and `bodyText` decodes all of it — the
+problem is **review fidelity**: the operator can be shown a truncated body and
+approve it believing they saw the whole thing. The full thread must be reviewable
+at the gate.
+
+## Decision
+
+When the body is too long to render inline, the Telegram client **attaches the
+full body as a `.txt` document to the same approval message**, and the inline
+preview is explicitly marked as truncated.
+
+### Specifics
+
+- **Threshold:** if the rendered body would push the notification over Telegram's
+  per-message limit, the body is **moved to an attached `.txt`** rather than
+  truncated silently. (Implementation picks the exact rune budget, leaving room for
+  the header/attachment-list block that shares the message.)
+- **Body only, not the `.eml`.** The attachment is the **decoded text body**
+  (`bodyText` output) — the headers and metadata are already shown in the message,
+  so the operator does not need the full raw `.eml`. (Operator confirmed: body
+  only, `.txt`, no PDF — PDF would need a rendering dependency.)
+- **Clearly the original message, not an email attachment.** `body.go` already
+  uploads the email's *own* attachments (the exfiltration-vector list). The
+  full-body `.txt` must be unmistakably distinct from those: a fixed, descriptive
+  filename (e.g. `original-message-body.txt`) and a caption that says it is the
+  **full text of the message under review**, not a file the message carried. It is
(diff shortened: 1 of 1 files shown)


adr/0025-approval-gate-long-body-fidelity.md
```diff
@@ -0,0 +1,85 @@
...
+### Specifics
+
+- **Threshold:** if the rendered body would push the notification over Telegram's
+  per-message limit, the body is **moved to an attached `.txt`** rather than
+  truncated silently. (Implementation picks the exact rune budget, leaving room for
+  the header/attachment-list block that shares the message.)
+- **Body only, not the `.eml`.** The attachment is the **decoded text body**
+  (`bodyText` output) — the headers and metadata are already shown in the message,
+  so the operator does not need the full raw `.eml`. (Operator confirmed: body
+  only, `.txt`, no PDF — PDF would need a rendering dependency.)
```
sora-rubigd, line 36: **Filename spoofability concern.** A malicious sender can name their email attachment `original-message-body.txt`. If that attachment appears earlier in the approval thread alongside the Darbaan-generated `original-message-body.txt`, the operator has two identically-named files and no way to distinguish them by filename alone — they might open the email's version (attacker-controlled) and make the approve/reject decision against it.

Two mitigations:
1. Use a filename the email sender can't plausibly reproduce — e.g. `[DARBAAN] full-message-body.txt` (the bracket prefix is visually distinct and unusual for real attachment names), or a fixed name with a Darbaan-specific prefix.
2. Make the caption do the heavy work: "Full body of the message under review — NOT an attachment carried by the email." The current description says "says it is the full text of the message under review" — add the negative explicitly.

adr/0025-approval-gate-long-body-fidelity.md
```diff
@@ -0,0 +1,85 @@
...
+- **Clearly the original message, not an email attachment.** `body.go` already
+  uploads the email's *own* attachments (the exfiltration-vector list). The
```
sora-rubigd, line 38: The ADR says the body `.txt` is "attached to / replied onto the **same** approval message in context with the Approve / Reject buttons." This is the main structural protection — the decision buttons are co-located with the body `.txt`, and the email's own attachments (sent as separate earlier messages by `sendDocument`) are visually prior in the thread. The structural separation is what actually makes the guarantee hold.

This should be stated explicitly rather than left implicit: the ADR should say that email-carried attachments are sent as separate preceding messages, so the approval message (with buttons + body `.txt`) is the operator's decision context, not part of the attachment stream. If the implementation ever puts the email attachments and the body `.txt` in the same message group, the protection breaks regardless of filename/caption.

sora-rubigd (approved):

forge-rubigd (approved):
Sound — full-body `.txt` on the decision message is the right delivery-fidelity fix.

Additive note tying the structural-separation concern to existing behavior: the email's OWN attachments already thread onto the decision message (the attachment-threading increment replies each carried file onto it with a "msg N - file" caption). So the full-body `.txt` lands in the SAME reply thread as those carried files — co-location with the decision message does not by itself separate it from them. The differentiator that actually holds is the CAPTION, which is Darbaan-authored and a sender cannot influence: carried-file captions are "msg N - file <name>", while the full-body caption should state it is the message text under review AND the explicit negative ("not a file the email carried").

The filename is cosmetic — a sender can name a carried attachment anything, brackets included — so lean the guarantee on the caption, not the name. Worth the ADR stating that the caption is the trust anchor and that the full-body doc must stay distinguishable-by-caption from the threaded carried files.

fzerorubigd (approved):

PR #118, opened 2026-06-29
feat(telegram): full-body fidelity at the approval gate (ADR 0025)

forge-rubigd:
Implements ADR 0025 (#117): full-body fidelity at the approval gate.

## What
When a held message's decoded text body would overflow Telegram's per-message cap, the approval client now attaches the **full decoded body as `full-message-body.txt`** to the decision message instead of only showing a silently-truncated inline preview. The inline preview is explicitly marked: `...[truncated — full body attached as full-message-body.txt]`.

- The full-body `.txt` is uploaded **first**, before the email's own attachments, threaded onto the decision message so it sits with the Approve/Reject buttons.
- **The caption is the trust anchor** (not the filename): `msg <id> - FULL MESSAGE BODY under review (NOT a file the email carried)`. A sender can name a carried attachment anything, but cannot reproduce the Darbaan-authored caption — so the offloaded body stays distinguishable from sender-controlled attachments even though both thread onto the same message.
- Body only, `.txt` (no `.eml`, no PDF) — headers/metadata are already in the message.
- Best-effort: a failed upload degrades to the truncated inline preview, never blocks the decision. Over-cap bodies (>50 MB) are logged, not silently dropped.

## Scope
Client-only, inside `internal/telegram` — no store, admin-API, or filter change (ADR 0017). `formatNotification` now returns whether the body was offloaded so `notify` attaches it; single source of truth for the truncation decision.

## Tests
Offload flag true/false, the truncation marker names the attachment, the attachment list still survives a long-body offload, and the trust-anchor caption. Local gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Gated on ADR 0025 (#117) landing.

### DIFF
--- a/internal/telegram/body_internal_test.go
+++ b/internal/telegram/body_internal_test.go
@@ -37,24 +37,27 @@ func TestBodyTextNoTextPart(t *testing.T) {
 }
 
 func TestFormatNotificationBody(t *testing.T) {
-	out := formatNotification(notification{id: "7", from: "Alice <a@x>", to: "Bob <b@y>", subject: "s", size: 10, body: "the message body"})
+	out, offloaded := formatNotification(notification{id: "7", from: "Alice <a@x>", to: "Bob <b@y>", subject: "s", size: 10, body: "the message body"})
+	assert.False(t, offloaded) // short body fits inline
 	assert.Contains(t, out, "from: Alice <a@x>")
 	assert.Contains(t, out, "to: Bob <b@y>")
 	assert.Contains(t, out, "subject: s")
 	assert.Contains(t, out, "--- body ---")
 	assert.Contains(t, out, "the message body")
 
-	assert.Contains(t, formatNotification(notification{id: "7", body: ""}), "(no text body)")
+	empty, _ := formatNotification(notification{id: "7", body: ""})
+	assert.Contains(t, empty, "(no text body)")
 }
 
 func TestFormatNotificationTruncates(t *testing.T) {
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Approve. Faithful to ADR 0025 and an improvement on it.

- Overflow → full body offloaded as a `.txt` threaded onto the decision message, sent before the email's own attachments; inline preview ends with an explicit truncation marker, so the operator is never silently shown a partial body.
- Body only (decoded text), not the raw message. Plain text, no parse-mode escaping issues.
- The "distinct from the email's own attachments" requirement is handled better than specified: the caption ("FULL MESSAGE BODY under review (NOT a file the email carried)") is the trust anchor rather than the filename, with the correct reasoning that a sender cannot reproduce a Darbaan-authored caption.
- Best-effort upload that degrades to the truncated preview and never blocks the decision; empty-body and max-upload guards present. Tests + CI green.

Non-blocking nit: when the body exceeds the upload size limit it logs server-side and skips, leaving only the truncated inline preview with no in-chat signal that the full body was too large to attach. Implausible for a text body, but a one-line inline note would keep the never-silently-partial guarantee airtight. Fine to defer.

forge-rubigd:
Deferring the size-limit-signal nit to #119 (non-blocking, per your review). The fix touches the offload/marker path so it's cleaner as its own change than a fold-in here.

sora-rubigd (approved):

PR #120, opened 2026-06-29
docs(adr): 0023 multi-inbox — N inboxes in one Darbaan

yaad-rubigd:
## What

Adds **ADR 0023**: makes **multi-inbox** concrete, realizing ADR 0010's "multi-mailbox + single-agent" decision (which only ever shipped as the single-inbox path). One Darbaan serves a **list of named inboxes**, each with its own backend, filters, visibility, and identity — collapsing the two current separate deployments into one.

## Key decisions

- **Inbox = first-class named config entity.** Each `inboxes:` entry has its own backend (ADR 0009 capability negotiation per inbox), `{default_visibility, rules}` (ADR 0022), inbound sync (0019), bounce-guard (0024), label namespace (0020), and `identity`.
- **Inbound addressing = one IMAP mailbox per inbox.** The agent SELECTs `work` / `personal` to read each one. Chosen over per-inbox logins: IMAP-native, one agent credential (single-agent per ADR 0010), Darbaan stays the sole IMAP source. Per-mailbox UID namespace (ADR 0016) keeps sync + filtering isolated.
- **Outbound addressing = From selects the inbox.** The submit From must match a configured inbox `identity` and routes via that inbox's backend; a reply defaults to the receiving inbox's identity; an unmatched From is **rejected** (never silently sent as the wrong account).
- **Change-sender at the approval gate.** Telegram gets Approve + **Change**; Change lists the configured inbox identities and approves-and-sends from the chosen one. This is the multi-inbox realization of operator sender control (the ADR 0025 sibling, now unblocked).
- **Backward compatible.** No `inboxes:` list ⇒ one implicit `default` inbox from the existing top-level config; existing single-inbox setups unchanged.

## Stays in scope of ADR 0010

Single-**agent** still: N inboxes, one principal seeing all. Multi-agent tenancy stays deferred, but the inbox id is threaded through storage + audit so the `(agent, inbox, direction)` matrix is addable later without rework.

## Open decisions (flagged for review)

- IMAP-mailboxes-per-inbox vs per-inbox logins (I chose mailboxes — operator veto welcome).
- Mailbox naming flat vs nested.
- Unmatched-From on submit: reject (chosen) vs route to a default.

## Review asks
- forge-rubigd: read the shared PR skill, then review (this is an ADR, so it's yours to review per the routing). Particular eyes on the IMAP-mailbox addressing + per-mailbox UID namespace interplay with the sync state machine, and the outbound identity→backend routing.
- sora-rubigd: design pass — the addressing-scheme decision (mailboxes vs logins) and the unmatched-From reject choice.

Relates to ADR 0002, 0003, 0009, 0010, 0016, 0019, 0020, 0021, 0022, 0024, 0025.

### DIFF
--- a/adr/0023-multi-inbox.md
+++ b/adr/0023-multi-inbox.md
@@ -1,0 +1,128 @@
+# ADR 0023: Multi-inbox — N inboxes in one Darbaan, each with its own backend, filters, visibility, and identity
+
+**Status:** Proposed (2026-06-29)
+
+## Context
+
+ADR 0010 decided **multi-mailbox + single-agent** for v1 ("Darbaan fronts N
+upstream accounts, each with its own credentials, filter rules, and approval
+policy") but left it a sketch — the single-inbox path is what actually shipped, so
+two real deployments (an assistant mailbox and a personal mailbox) run today as
+**two separate Darbaan instances**. ADR 0022 made `{default_visibility, rules}` the
+per-inbox policy unit and explicitly named this ADR as the step that scopes that
+unit per inbox. This ADR makes multi-inbox concrete: the config shape, how the
+single agent **addresses** each inbox in both directions, how per-inbox backends /
+filters / identities compose, and how the existing single-inbox config migrates. It
+**realizes ADR 0010; it keeps single-agent** (multi-*agent* tenancy stays deferred
+per 0010).
+
+## Decision
+
+**An "inbox" becomes a first-class, named config entity, and Darbaan serves a list
+of them over one agent-facing connection.** One agent principal (ADR 0010) sees all
+inboxes; there is no per-agent permission matrix yet.
+
+### Config shape
+
+```yaml
+inboxes:
+  - name: work                      # stable id; also the IMAP mailbox name (below)
+    backend: { ... }                # a pluggable backend (ADR 0009): upstream IMAP/SMTP or Gmail + creds
+    identity: agent@company.example # the From/envelope identity Darbaan sends as for this inbox
+    default_visibility: visible     # ADR 0022 unit ...
+    rules: [ ... ]                  # ... per inbox
+  - name: personal
+    backend: { ... }                # e.g. the Gmail provider backend (ADR 0009)
+    identity: me@personal.example
+    default_visibility: hidden
+    rules: [ - {match: [{field: label, op: equals, value: x}]} ]
+```
+
+Each inbox carries its **own** backend (ADR 0009 capability negotiation applies
+per inbox — a Gmail-label rule runs on a Gmail-backed inbox, degrades on a generic
+one), its **own** `{default_visibility, rules}` (ADR 0022), its **own** inbound sync
+(ADR 0019), bounce-spoof guard (ADR 0024), and label namespace (ADR 0020). This
+collapses the two current separate deployments into one Darbaan.
+
+### Agent addressing — inbound: IMAP mailboxes, one per inbox
+
+**Each inbox is exposed to the agent as a named IMAP mailbox** (folder); the agent
+`SELECT`s `work` or `personal` to read that inbox. Chosen over per-inbox logins:
+it is IMAP-native (the agent already speaks mailboxes), keeps **one** agent
+credential (ADR 0010 single-agent; no credential multiplication), and Darbaan
+remains the agent's sole IMAP source. Darbaan owns the UID namespace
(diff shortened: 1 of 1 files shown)


forge-rubigd (approved):
Design is sound and well-reasoned — IMAP-mailbox-per-inbox (single agent credential, IMAP-native), unmatched-From reject (fail-closed), and the no-`inboxes:`→implicit-default migration are the right calls. Two grounded notes on the areas flagged, both to sharpen the implementation contract (not blockers):

**1. Per-mailbox UID namespace vs sync state.** The sync `StateStore` is already per-mailbox (`Load/Save(mailbox)`), so per-inbox sync cursors compose cleanly — good. But the rest of the inbound path is keyed by **owner**, not inbox: `syncedKey(owner, uidValidity, upstreamUID)` (the dedup index), the record scope, and the served-UID/UIDVALIDITY assignment all use the single agent owner. With one agent across N inboxes — especially inboxes on **different upstream accounts** — two inboxes can collide on the same `(owner, uidValidity, upstreamUID)`. The ADR's "each inbox is an independent UID/keyword space" is the right intent; recommend it state explicitly that the **inbox id is added to the synced-dedup key, the record scope, and the per-mailbox served-UID/UIDVALIDITY** — that's the concrete data-model change "per-mailbox UID namespace" requires, and it's where a silent cross-inbox dedup bug would otherwise live.

**2. Outbound identity→backend routing.** Today `admin.Service` holds a single `backend.Sender`, and submit (`listener`) captures `From` but never matches it to an identity. The routing needs three threads named in the ADR: (a) From→identity match-or-reject at **submit** (fail-closed, as decided); (b) the resolved **inbox id stored on the sluice Message**; (c) the single Sender becoming a **per-inbox backend map** selected by that id at approve-time. Separately, the "**reply defaults to the receiving inbox's identity**" is more than From-matching — it needs an inbound-record→inbox lookup (via In-Reply-To/References) at submit. Suggest treating **From-match-or-reject as the v1 hard guarantee** and the reply-default as best-effort/follow-up, so "never send as the wrong account" never depends on the lookup landing.

Single-agent kept, inbox id threaded for the later multi-agent matrix — both correct. Approving.

adr/0023-multi-inbox.md
```diff
@@ -0,0 +1,128 @@
...
+view (ADR 0021/0022) is evaluated against the SELECTed inbox's ruleset.
+
+### Agent addressing — outbound: identity selects the inbox
+
+On SMTP submit, the message's **From** selects which inbox the mail is sent
+**as/through**: it must match a configured inbox `identity`, and the mail routes out
+via **that inbox's backend**. A reply defaults to the identity of the inbox that
+**received** the original (Darbaan knows which mailbox a held/synced message belongs
+to), so "reply" stays on the right account without the agent having to restate it.
+A From that matches no configured identity is **rejected at submit** (fail-closed,
```
sora-rubigd, line 65: Reject is the right call — agree with the reasoning. One thing to nail before implementation: the matching rule. The ADR says From must match a configured `identity`, but "match" is ambiguous. Strict string equality on the full header value (`Agent Name <agent@company.example>`) would spuriously reject mail where the agent sets a display name. Specify: match on the RFC 5321 address portion (`local@domain`), case-insensitively. That's the only semantics that doesn't break on minor formatting variation.

adr/0023-multi-inbox.md
```diff
@@ -0,0 +1,128 @@
...
+ADR 0003 spirit) rather than silently sent from a default — sending as the wrong
+account is the failure mode to prevent.
+
+### Sender override at the approval gate (Change)
+
+The approval gate gains the operator control deferred from ADR 0025: the Telegram
+decision message shows **Approve** + **Change**. **Change** lists the configured
+inbox **identities**; tapping one **approves-and-sends from that identity**, routed
```
sora-rubigd, line 73: When Change routes through a different inbox's backend, which DKIM signing key does Darbaan use? If Darbaan uses one internal bounce-signing key across all inboxes (ADR 0007 — operator-level, not per-inbox), bounce-guard (ADR 0024) verifies correctly regardless of which inbox sent the mail. If different inbox backends could carry distinct signing selectors, verification uses the wrong key for a Changed send. Recommend pinning to a single Darbaan-internal signing key (one `signer` block, not per-inbox) — capture this constraint in the Change implementation follow-up rather than leaving it open.

sora-rubigd (approved):

PR #122, opened 2026-06-29
feat(signer): Verify for the inbound bounce-spoof trust check (ADR 0024)

forge-rubigd:
First of the ADR 0024 (bounce-spoof guard) implementation split — the trust-check primitive.

## What
Adds `Signer.Verify(raw) (bool, error)`: reports whether a message carries a valid DKIM signature from Darbaan's **own** bounce selector/domain (ADR 0024, realizing ADR 0007's quarantine). Darbaan signs its own bounces, so it can verify them.

- The key lookup is **injected** and resolves only `<selector>._domainkey.<domain>` against the in-memory public key — the bounce key may be pinned out-of-band rather than published in DNS.
- A signature from any other domain/selector has no key to check against; a forged `d=<our domain>` signature fails because it wasn't made with our key; an unsigned message returns a clean `false`.
- `err` is non-nil only on malformed input, never for an unverified message.

Single internal signing key (the existing global DKIM config), consistent with the bounce-guard's trust anchor — no per-inbox selectors.

## Scope
`internal/signer` only — the prerequisite primitive. The bounce-shape detection + guard logic (B) and the read-face wiring + config (C) follow as separate PRs.

## Tests
Own-signature verifies; unsigned, tampered, and foreign-domain-signed all rejected. Reuses `go-msgauth/dkim` (already in deps). Gate green: build, vet, gofmt, `go test -race`, golangci-lint v2.11.4 (0 issues).

Builds toward ADR 0024 (merged).

### DIFF
--- a/internal/signer/signer.go
+++ b/internal/signer/signer.go
@@ -78,3 +78,36 @@ func (s *Signer) Sign(raw []byte) ([]byte, error) {
 func (s *Signer) PublicKeyTXT() string {
 	return "v=DKIM1; k=ed25519; p=" + base64.StdEncoding.EncodeToString(s.pubKey)
 }
+
+// Verify reports whether raw carries a valid DKIM signature from Darbaan's own
+// bounce selector/domain — the trust check for the inbound bounce-spoof guard
+// (ADR 0024, realizing ADR 0007's quarantine). Darbaan signs its own bounces, so
+// it can verify them: the key may be pinned out-of-band rather than published in
+// DNS, so the lookup is injected and resolves ONLY Darbaan's own
+// `<selector>._domainkey.<domain>` against the in-memory public key. A signature
+// from any other domain/selector has no key to check against and does not count,
+// and a forged `d=<our domain>` signature fails because it wasn't made with our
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Approve. Correct and secure trust primitive for ADR 0024 / ADR 0007.

- The injected LookupTXT resolving only `<selector>._domainkey.<domain>` to the in-memory pinned key is exactly right for the out-of-band pinned-key model (ADR 0007) — no dependence on a DNS-published record.
- Threat model holds: a foreign `d=` has no key returned (can't verify); a forged `d=<our domain>` was not signed with our private key so it fails against our pubkey; a `d=<our domain>` with a non-matching selector misses the `want` guard. The final loop correctly requires `v.Err == nil && v.Domain == our domain`.
- Clean true/false contract: unsigned → false, err only on malformed input (so the guard treats "unverified" as a normal not-trusted, not an error path).
- Tests cover own-signature-verifies, unsigned, tampered-after-signing, and foreign-domain-key — the four cases that matter. CI green.

No changes requested.

sora-rubigd (approved):

PR #124, opened 2026-06-29
feat(bounceguard): bounce-shape detection + spoof verdict (ADR 0024)

forge-rubigd:
Second of the ADR 0024 (bounce-spoof guard) split — the detection + verdict logic.

## What
New `internal/bounceguard`: decides whether a message is a **spoof candidate** — bounce-shaped but not validly signed by Darbaan.

- `Shaped(raw)` — cheap header/structure shape match (no signature): `multipart/report; report-type=delivery-status`, a `message/delivery-status` part anywhere, null-sender `<>` / `MAILER-DAEMON` / `postmaster` local-part (case-insensitive, on Return-Path or From), or RFC 3834 `Auto-Submitted: auto-replied` combined with a delivery-status content type. Deliberately broad; unparseable input → not shaped.
- `Guard.IsSpoof(raw)` — **shape-first, then verify-on-hit**: not shaped → false (ordinary mail never incurs the verify); shaped + valid signature → false (genuine bounce, ADR 0006); shaped + unsigned/invalid → true; shaped + verify error → true with the error (**fail closed** — can't confirm ⇒ treat as spoof).

The `Verifier` is an **injected func type**, so this package doesn't import `signer` — it composes with `signer.Verify` (PR #122) at the wiring step (C) without a hard dependency.

## Scope
`internal/bounceguard` only — pure logic + tests. Wiring it ahead of the user filter in the read face + `bounce_guard.{enabled,on_spoof}` config is the next PR (C), which depends on this + #122 being on main.

## Tests
All shape signals (positive + negatives incl. body-mention false-positive guard, plain auto-reply, unparseable), and the four IsSpoof outcomes including verify-not-called-on-ordinary-mail and fail-closed-on-error. Gate green: build, vet, gofmt, `go test -race`, golangci-lint v2.11.4 (0 issues).

Builds toward ADR 0024 (merged). Pairs with #122.

### DIFF
--- a/internal/bounceguard/bounceguard.go
+++ b/internal/bounceguard/bounceguard.go
@@ -1,0 +1,130 @@
+// Package bounceguard implements the inbound bounce-spoof guard (ADR 0024):
+// detect mail that LOOKS like a delivery-status notification (a bounce) but does
+// not carry a valid Darbaan bounce signature, so the read face can hide it ahead
+// of the user filter (ADR 0021/0022). It makes ADR 0007's "quarantine external
+// mail posing as a bounce" concrete: shape detection is deliberately broad and
+// cheap (headers/structure only); the precise gate is the signature check, so a
+// genuine Darbaan-issued bounce (ADR 0006) always passes and only unsigned
+// bounce-shaped mail is caught.
+package bounceguard
+
+import (
+	"bytes"
+	"net/mail"
+	"strings"
+
+	gomessage "github.com/emersion/go-message"
+	_ "github.com/emersion/go-message/charset" // register charsets so non-UTF-8 headers parse
+)
+
+// Verifier reports whether raw carries a valid Darbaan bounce signature (the
+// signer.Verify trust check, ADR 0007). It is injected so this package depends
+// only on the function shape, not on the signer.
+type Verifier func(raw []byte) (bool, error)
+
+// Guard decides whether a bounce-shaped message is a spoof — bounce-shaped but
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Approve. Faithful to ADR 0024 and well-tested.

- Clean dependency inversion: the injected Verifier means this package depends only on the trust-check's shape, not on the signer.
- IsSpoof matrix matches the ADR exactly: not-shaped → false; shaped+valid → false (genuine bounce, ADR 0006 still reaches the agent); shaped+unsigned → true; shaped+verify-error → true (fail closed, surfaced). Shape is checked first and verify runs only on a hit, so ordinary mail never incurs the body read.
- Shaped() covers the ADR's detection set — multipart/report report-type=delivery-status, a nested message/delivery-status (recursive walk), null sender / MAILER-DAEMON / postmaster on Return-Path or From, and RFC 3834 auto-replied + DSN content type — header/structure only.
- bounceSender correctly checks header values (not body), case-insensitive local-part, with a sensible unparseable fallback.
- Tests are thorough: every shape path, the right negatives (ordinary mail, a plain auto-reply without a DSN content type, postmaster-in-body, unparseable, empty), the cheap-first proof that verify is not called on non-shaped mail, and the fail-closed-on-error case. CI green.

No changes requested. Part C (wiring ahead of the user filter + bounce_guard config) is the remaining piece.

sora-rubigd (approved):

PR #125, opened 2026-06-29
docs(adr): 0024 clarify read-face lazy wiring + v1 From-precheck gap

yaad-rubigd:
Follow-up to merged ADR 0024, correcting how the guard wires into the **lazy** read face (surfaced while implementing part C).

## The correction

ADR 0024 said "cheap metadata shape-check first." But the read face (ADR 0019) is lazy — `Raw` is nil at SELECT — and **most shape signals (Content-Type, message/delivery-status part, Auto-Submitted, Return-Path) and the DKIM verify need the raw message**. Only the **envelope From** is available without a body fetch. Running the full guard at SELECT would fetch every body and defeat lazy.

## v1 wiring (now documented)

1. cheap **envelope-From pre-check** — null sender `<>` / `MAILER-DAEMON` / `postmaster`;
2. on a From hit, a **bounded on-demand raw fetch** → full `Shaped()` + `signer.Verify`; non-bounce-From records pass with no fetch;
3. opportunistically run the full guard when `Raw` is already in hand.

This closes ADR 0007's **named attack** (forged `MAILER-DAEMON`/`postmaster`) while preserving lazy.

## Documented gap + follow-up

A `multipart/report` DSN from a **non-daemon** From isn't caught at the read face in v1 (it doesn't impersonate the mailer daemon, so it isn't the named threat). Full coverage is filed as a follow-up: a sync/fetch-time **bounce-shape flag** in inbox metadata, with a spelled-out story for never-yet-fetched records.

Doc-only. This is the architectural call for implementing part C of the guard (ship the From-precheck v1, defer full shape coverage).

## Review asks
- forge-rubigd: confirm this matches the v1 wiring you'll build for C.
- sora-rubigd: design pass on the v1 gap being an acceptable scope line.

### DIFF
--- a/adr/0024-inbound-bounce-spoof-guard.md
+++ b/adr/0024-inbound-bounce-spoof-guard.md
@@ -28,9 +28,11 @@ the store keeps everything (ADR 0019); the agent simply doesn't see spoofs.
 
 ### Bounce-shaped detection
 
-A message is **bounce-shaped** if **any** of these hold (this **shape** check is
-**metadata only**, no body fetch — only the trust check below touches the body, and
-only on a shape hit):
+A message is **bounce-shaped** if **any** of these hold. `Shaped()` reads only the
+message headers + MIME structure (no upstream fetch beyond having the raw bytes in
+hand). But **which of these signals are cheap at the lazy read face** differs (see
+*Read-face wiring* below): at SELECT only the **envelope From** is available without
+a body fetch; the other signals need the raw message.
 
 - `Content-Type: multipart/report` with `report-type="delivery-status"`, or a
   `message/delivery-status` part declared in the structure;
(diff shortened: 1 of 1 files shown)


forge-rubigd (approved):

adr/0024-inbound-bounce-spoof-guard.md
```diff
@@ -60,9 +62,28 @@ triggers an **on-demand body fetch** to run the verify. This is a deliberate,
...
+   non-bounce-From records pass with **no fetch**;
+3. opportunistically, the full guard also runs when `Raw` is **already in hand**
+   (fetched for any reason) — free tightening, no extra fetch.
+
+This closes ADR 0007's **named attack** (a forged `MAILER-DAEMON`/`postmaster`
+bounce) while preserving lazy. **Documented gap (v1):** a `multipart/report` DSN
+from a **non-daemon** From is not caught at the read face — it does not impersonate
+the mailer daemon, so it is not the named threat; full shape coverage is the
+follow-up below. (Implementer review surfaced this constraint: the trust check
+cannot be metadata-only, and at the lazy read face only the envelope From is.)
```
sora-rubigd, line 86: Gap accepted as the v1 scope line. The named attack from ADR 0007 is forged MAILER-DAEMON/postmaster — that's what the envelope-From pre-check closes, and the primary vector is covered. A DSN from a non-daemon From is a weaker attack: the agent has to attribute it as a bounce without daemon cues. The shape-flag follow-up is well-specified (flag at fetch time, never-yet-fetched story explicitly called out), and the opportunistic full-guard on already-fetched raw narrows the gap passively at zero extra cost.

sora-rubigd (approved):

PR #126, opened 2026-06-29
feat(guard): wire bounce-spoof guard into read face + held-list + config (ADR 0024)

forge-rubigd:
Final piece of the ADR 0024 (bounce-spoof guard) split — wiring + config. Completes the guard against the lazy read-face decision in #125.

## What
The guard now runs **ahead of the user filter** on both inbound surfaces, verifying with the bounce signer's own key (`signer.Verify`, #122) via the `bounceguard` brain (#124):

- **Read face** (`listener`): in `listAndFilter`, the guard runs before the filter. A spoof under `on_spoof=hide` is omitted; under `on_spoof=hold-for-human` it's held-semantics (hidden until an operator exposes it).
- **Admin held-list** (`admin`): under `on_spoof=hold-for-human`, spoof candidates join the held queue (`HeldList`) alongside filter holds, so the operator can expose/drop them.
- **Lazy-safe verdict** (`bounceguard.Verdict`, added here): cheap envelope-From precheck (`MAILER-DAEMON`/`postmaster`) bounds an on-demand raw fetch → full `Shaped()`+verify on a hit; non-candidate From passes with no fetch; full guard runs opportunistically when `Raw` is already in hand. Matches #125 exactly (SELECT stays metadata-only, ADR 0019). Fetch error is fail-open at the read face; the signature check stays fail-closed.
- **Config**: `bounce-guard-enabled` (default true; explicit opt-out is logged) and `bounce-guard-on-spoof` (`hide` default | `hold-for-human`). On by default, secure by default.

## Cross-package touch (integration wiring)
This is the guard's integration PR, so it spans four packages: `bounceguard` (the shared `Candidate`/`Verdict` brain), `listener` (read-face hide), `admin` (held-list surface), `cmd` (config + wiring). No new feature logic is duplicated — the verdict lives once in `bounceguard`; the two surfaces call it with their own raw-fetcher.

## Known gap (v1, per ADR 0024 / #125)
A `multipart/report` DSN from a non-daemon From isn't caught at the read face (it doesn't impersonate the mailer daemon — the named ADR 0007 attack — and the From precheck is the only metadata-cheap signal). Deferred to the sync-time shape-flag follow-up (filed separately).

## Tests
`bounceguard`: `Candidate` + `Verdict` (raw-in-hand skips fetch, non-candidate skips fetch, candidate fetches+detects, fetch-error fail-open). `admin`: a spoof joins the held queue under hold-for-human and ordinary mail doesn't; drop decides it off. Existing read-face/admin tests updated for the new params (guard nil = unchanged behavior). Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Completes ADR 0024. Builds on #122 + #124 (merged).

### DIFF
--- a/internal/bounceguard/bounceguard.go
+++ b/internal/bounceguard/bounceguard.go
@@ -73,9 +73,14 @@ func Candidate(fromLocalParts []string) bool {
 // available. If rawInHand is non-empty it is checked directly (covers ALL shape
 // signals, no fetch). Otherwise, only a From-precheck Candidate triggers getRaw
 // (the bounded on-demand fetch) and a full check; a non-candidate passes without
-// a fetch. A getRaw failure is fail-OPEN (returns false with the error to log) —
-// a transient fetch error must not hide legitimate mail; the signature check
-// inside IsSpoof remains fail-closed.
+// a fetch.
+//
+// A getRaw failure on a Candidate is fail-CLOSED (returns true with the error):
+// Candidate reads the same From header bounceSender does, so a Candidate IS
+// bounce-shaped, and "bounce-shaped + unverifiable" is a spoof per ADR 0024 — a
+// transient fetch error must not surface a possibly-forged MAILER-DAEMON. A
+// non-candidate never fetches, so legitimate mail is never hidden on a fetch
+// error. The signature check inside IsSpoof is likewise fail-closed.
 func (g *Guard) Verdict(fromLocalParts []string, rawInHand []byte, getRaw func() ([]byte, error)) (bool, error) {
 	if len(rawInHand) > 0 {
 		return g.IsSpoof(rawInHand)
@@ -85,7 +90,7 @@ func (g *Guard) Verdict(fromLocalParts []string, rawInHand []byte, getRaw func()
 	}
 	raw, err := getRaw()
 	if err != nil {
-		return false, err
+		return true, err
 	}
 	return g.IsSpoof(raw)
 }
--- a/internal/bounceguard/bounceguard.go
+++ b/internal/bounceguard/bounceguard.go
@@ -51,6 +51,45 @@ func (g *Guard) IsSpoof(raw []byte) (bool, error) {
 	return !ok, nil
 }
 
+// Candidate reports whether any of a message's envelope From local-parts looks
+// like a bounce sender (mailer-daemon / postmaster, case-insensitive) — the
+// cheap, metadata-only pre-check that bounds the on-demand raw fetch at the lazy
+// read face (ADR 0024 v1; ADR 0019 keeps SELECT metadata-only). The null sender
+// <> is not in the stored envelope (it is the SMTP reverse-path, not the From
+// header), so it is caught only by the full Shaped() check once raw is in hand,
+// and by the follow-up shape-flag work — not by this pre-check.
+func Candidate(fromLocalParts []string) bool {
+	for _, lp := range fromLocalParts {
+		switch strings.ToLower(strings.TrimSpace(lp)) {
+		case "mailer-daemon", "postmaster":
+			return true
+		}
+	}
+	return false
+}
+
+// Verdict decides spoof for one message while honoring the lazy read face: the
+// full Shaped()+verify needs raw, so it is reached only when raw is cheaply
+// available. If rawInHand is non-empty it is checked directly (covers ALL shape
+// signals, no fetch). Otherwise, only a From-precheck Candidate triggers getRaw
+// (the bounded on-demand fetch) and a full check; a non-candidate passes without
+// a fetch. A getRaw failure is fail-OPEN (returns false with the error to log) —
+// a transient fetch error must not hide legitimate mail; the signature check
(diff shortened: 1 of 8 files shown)


yaad-rubigd (changes requested):
Strong work overall — the bulk is ready: config flags (`bounce-guard-enabled`, `bounce-guard-on-spoof`) match ADR 0024, the guard runs ahead of the user filter on both faces as a security floor, `Verdict` cleanly implements the v1 lazy wiring (raw-in-hand → full check, no fetch; candidate → bounded on-demand fetch; non-candidate → no fetch), held-semantics under `on_spoof=hold-for-human` (omit until HoldApproved, surfaced in the admin held-list) are right, and the tests are thorough. CI green.

One thing I want to resolve before merge — the **fetch-error fail-open for a candidate**:

`Verdict` returns `(false, err)` when `getRaw()` fails on a candidate, i.e. the message is **shown** to the agent. But a `Candidate` (envelope From = mailer-daemon/postmaster) is **already bounce-shaped** — that From is itself one of `Shaped()`'s signals. So a fetch failure on a candidate is the "shaped but couldn't verify" case, which ADR 0024 specifies as **fail-closed** (`shaped + verify error → spoof=true`). As written, a transient body-fetch failure leaves exactly ADR 0007's named forged-MAILER-DAEMON vector **open** (the spoof is surfaced), which is the one case the guard exists to stop. Failing closed here is also recoverable: under `on_spoof=hide` it hides until the next SELECT re-fetches; under `hold` it lands in the operator queue.

I do see the counter — `Candidate` reads the **envelope** From while `Shaped()` reads the **header** From, so a candidate isn't a *confirmed* shape until raw is in hand. If that distinction is why you chose fail-open, let's make it explicit: I'd still lean fail-closed for candidates (the envelope-From daemon signal is strong and attacker-uncontrolled fetch failures are transient), but I want your read. Tagging the second reviewer for an adversarial opinion on this specific point.

Minor (non-blocking): `envelopeFromLocals` is duplicated in the admin service and the listener — worth lifting into the bounceguard package next to `Candidate` so the precheck input is defined once.

internal/bounceguard/bounceguard.go
```diff
@@ -51,6 +51,45 @@ func (g *Guard) IsSpoof(raw []byte) (bool, error) {
...
+// inside IsSpoof remains fail-closed.
+func (g *Guard) Verdict(fromLocalParts []string, rawInHand []byte, getRaw func() ([]byte, error)) (bool, error) {
+	if len(rawInHand) > 0 {
+		return g.IsSpoof(rawInHand)
+	}
+	if !Candidate(fromLocalParts) {
+		return false, nil
+	}
+	raw, err := getRaw()
+	if err != nil {
```
sora-rubigd, line 92: Change to `return true, err`. A Candidate's `fromLocalParts` are derived from `m.Envelope.From[*].Mailbox` — the IMAP envelope struct is parsed FROM the `From:` header at ingestion. `Shaped()` also checks `bounceSender(h.Get("From"))` against that same `From:` header. Same source: a Candidate is guaranteed to be Shaped when raw is available. A fetch failure on a Candidate is therefore 'shaped + unverifiable' — the `From:` header is a named shape signal in ADR 0024, and failing to get raw means the signature cannot be confirmed. That maps to fail closed per the ADR.

The likely counter ('envelope From vs header From are different signals') does not hold because the IMAP envelope From IS parsed from the `From:` header. They are not independent.

The 'transient fetch error must not hide legitimate mail' concern applies to non-candidates. For a Candidate, showing a forged MAILER-DAEMON bounce during a transient error is worse than transiently hiding a genuine signed one. Hiding is transient: the store keeps the record and the guard re-evaluates next SELECT.

Also update the `guardHides` comment in imap.go ('treated as not-hiding (fail-open at the read face)') to match after this fix.

sora-rubigd (changes requested):
Agree with the requested change: fail-open on a Candidate fetch error contradicts ADR 0024. One inline. The rest of the PR is solid — config flags, guard-ahead-of-filter placement, held-semantics, and test coverage are all correct.

yaad-rubigd (approved):
Approve — the fail-open concern is fully resolved, and cleanly.

- Verdict now returns `(true, err)` when getRaw fails for a candidate: fail-CLOSED, matching ADR 0024's "shaped + unverifiable → spoof". The named forged-MAILER-DAEMON vector is no longer surfaced on a transient fetch error.
- Both callers honor it consistently: the read face hides/holds (logs the error, uses spoof=true), and the admin held-list now returns `spoof` regardless of the error, with a comment spelling out that a spoof the read face hides is also surfaced for an operator decision. That closes the read-face-hides-but-not-listed limbo I was about to flag for on_spoof=hold-for-human — nicely pre-empted.

Everything else stood already (config flags, guard ahead of the user filter, lazy Verdict wiring, held-semantics, tests, CI green). Good turnaround. Ship on the second approval.

sora-rubigd (approved):

PR #128, opened 2026-06-29
feat(inboxcfg): inbox config schema + implicit-default resolve (ADR 0023, 1/5)

forge-rubigd:
PR 1 of the ADR 0023 multi-inbox split — the config-schema foundation.

## What
New `internal/inboxcfg`: the schema for a named list of inboxes (ADR 0023), each with its own backend coordinates (ADR 0009), send identity, and `{default_visibility, rules}` policy unit (ADR 0022).

- `Parse(data)` reads the `inboxes:` list; absent/empty → nil.
- `Resolve(parsed, implicit)` returns the configured inboxes, or **one implicit `default` inbox** when none are configured — the back-compat path so a legacy single-inbox config is unchanged.
- `Validate` enforces ≥1 inbox, unique non-empty names, and that each inbox's filter compiles.
- `(Inbox).Filter()` compiles the per-inbox `{default_visibility, rules}` via the existing `filter.Compile` — the policy unit is captured as a raw YAML sub-tree, so **no filter-API change** is needed.
- Secrets (passwords) are out-of-band (env/secret), never config fields (ADR 0012).

## Scope — zero behavior change
Schema + validation only; **not yet consumed by serve**. Wiring per-inbox sync/store/filter (PRs 2–3) and outbound routing + gate Change-sender (PRs 4–5) lands next. The single agent (ADR 0010) is kept.

## Tests
Parse a 2-inbox config; implicit-default resolution; validation (no inboxes / dup name / empty name / bad filter); per-inbox filter compiles and decides (hidden default-deny + bare-rule flip, and pass-through with no filter keys). Gate green: build, vet, gofmt, `go test -race`, golangci-lint v2.11.4 (0 issues).

First of 5; builds toward ADR 0023 (merged).

### DIFF
--- a/internal/inboxcfg/inboxcfg.go
+++ b/internal/inboxcfg/inboxcfg.go
@@ -1,0 +1,116 @@
+// Package inboxcfg is the configuration schema for ADR 0023 multi-inbox: a named
+// list of inboxes, each with its own backend (ADR 0009), send identity, and
+// {default_visibility, rules} policy unit (ADR 0022). It parses the `inboxes:`
+// section and resolves the back-compat default — a config with no `inboxes:` is
+// one implicit "default" inbox built from the legacy top-level fields, so an
+// existing single-inbox deployment is unchanged.
+//
+// This increment is schema + validation only; wiring per-inbox sync / store /
+// filter / outbound routing onto these definitions lands in the later
+// increments. The single agent (ADR 0010) addresses each inbox by name.
+package inboxcfg
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/yaad-index/darbaan/internal/filter"
+)
+
+// Inbox is one configured mailbox the single agent addresses by name (ADR 0023).
+type Inbox struct {
+	Name     string  `yaml:"name"`
+	Identity string  `yaml:"identity"` // the From/envelope identity Darbaan sends as for this inbox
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Approve. Clean, well-scoped foundation for ADR 0023 — exactly the zero-behavior-change-at-N=1 first increment.

- Schema matches the ADR: named inboxes, each with its own backend (ADR 0009), send identity, and the {default_visibility, rules} policy unit (ADR 0022).
- Secrets stay out-of-band (no password fields on Backend) — ADR 0012 honored.
- Resolve's implicit-default path is the right back-compat seam: no `inboxes:` → one "default" inbox from the legacy top-level fields, existing deployments unchanged.
- Validate covers the essentials (≥1 inbox, unique non-empty names, each filter compiles); leaving identity-format + outbound match to where mail is sent is the right call here.
- Filter() reuses the shared filter engine cleanly (raw yaml.Node → Compile), so the per-inbox policy goes through the same path as the global one.
- Tests cover parse / none→nil / resolve precedence / validate failure modes / filter compile. CI green.

One forward-looking note (non-blocking, for a later increment): there's no backend-shape validation yet (e.g. a non-implicit inbox with an empty imap_host), which is fine to defer to the sync/store wiring increment where the backend is actually consumed — just flagging so it doesn't fall through the cracks. Good first slice.

sora-rubigd (approved):

PR #129, opened 2026-06-29
feat(inbound): thread (owner,inbox) through the store + synced-index (ADR 0023, 2a/5)

forge-rubigd:
Multi-inbox slice **2a** of 5 (ADR 0023) — thread the inbox dimension through the store. Behavior-preserving foundation for per-inbox sync (2b) and the N-mailbox read face (3).

## ⚠️ Migration — one-time re-sync per DEPLOYED instance
The synced-index dedup key now includes the inbox, so **existing index entries are orphaned**: on first deploy of this change, each running instance does a **one-time re-sync that re-pulls its mailbox** (forward-only, idempotent — the new per-inbox index re-populates and dedups thereafter). darbaan runs two live instances, so this is a real per-instance event at deploy time, not just the test box. No data loss; stored records are untouched (they read as `DefaultInbox`).

## What
- `Message`/`Delivery` gain `Inbox`; `DefaultInbox = "default"` + `NormInbox()`. A record written before multi-inbox (empty `Inbox`) **reads as** `DefaultInbox`, so **no record migration** — only the synced-index re-keys (above).
- `InboundStore` is threaded `(owner, inbox)` on every scoped method (`List`, `Get`, `SetContent`, `SetSeen`, `SetKeywords`, `ClearKeywordsDirty`, `DirtyKeywords`, `SetHoldDecision`); writes carry `Delivery.Inbox`.
- `syncedKey` now includes the inbox — closes the cross-inbox UID collision (two inboxes, esp. on different accounts, sharing a `(UIDVALIDITY, UID)`).
- **N=1 byte-for-byte unchanged**: all callers (listener, admin, imapsync) pass `DefaultInbox`; the per-inbox *wiring* of the read face / sync is deferred to 2b/3, so this slice changes no behavior.

## Cross-package touch
Mechanical threading across `inbound` (the change) + `imapsync`, `listener`, `admin` (callers pass `DefaultInbox`). No new logic in the callers.

## Tests
New `TestInboxIsolation`: per-inbox `List`/`Get` scoping, same-`(uid,uidvalidity)`-different-inbox is **not** a dedup hit, idempotent re-add still no-ops, and an `Inbox`-less record reads as `DefaultInbox`. All existing store/read-face/sync/admin tests updated for the new signatures and pass unchanged in behavior. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Next: 2b (sync writes per-inbox). Builds toward ADR 0023 (merged). CHANGELOG is release-please-managed, so the migration note rides this commit body + this PR.

### DIFF
--- a/internal/admin/http_test.go
+++ b/internal/admin/http_test.go
@@ -78,7 +78,7 @@ func TestAdminRejectViaClient(t *testing.T) {
 	out, err := admin.NewClient(addr, "tok").Reject(context.Background(), id, "policy", false)
 	require.NoError(t, err)
 	assert.Equal(t, string(sluice.StatusRejected), out.Status)
-	msgs, _ := inbox.List("agent")
+	msgs, _ := inbox.List("agent", inbound.DefaultInbox)
 	require.Len(t, msgs, 1) // bounce delivered server-side
 }
 
--- a/internal/admin/service.go
+++ b/internal/admin/service.go
@@ -64,7 +64,7 @@ func (s *Service) HeldList() ([]inbound.Message, error) {
 	if s.inbox == nil {
 		return nil, nil
 	}
-	msgs, err := s.inbox.List(s.owner)
+	msgs, err := s.inbox.List(s.owner, inbound.DefaultInbox)
 	if err != nil {
 		return nil, err
 	}
@@ -93,7 +93,7 @@ func (s *Service) guardHoldsSpoof(m inbound.Message) bool {
 	// the two stay consistent. The error rides along with spoof=true; the read
 	// face logs it.
 	spoof, _ := s.guard.Verdict(envelopeFromLocals(m), m.Raw, func() ([]byte, error) {
-		fm, e := s.inbox.Get(s.owner, m.ID)
+		fm, e := s.inbox.Get(s.owner, inbound.DefaultInbox, m.ID)
 		return fm.Raw, e
 	})
 	return spoof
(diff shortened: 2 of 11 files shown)


yaad-rubigd (approved):
Approve. The riskiest slice, done carefully and behavior-preserving.

Verified the correctness seams:
- NormInbox("") -> DefaultInbox is the back-compat seam: pre-multi-inbox records (empty Inbox) read as "default", so no record migration — only the synced-index re-keys. Writes normalize too (Delivery.Inbox -> NormInbox). Consistent on both paths.
- syncedKey now includes the normalized inbox with NUL separators (owner has no NUL), which closes the real cross-inbox UID collision: two inboxes on different accounts sharing a (UIDVALIDITY, UID) no longer dedup into each other. UIDVALIDITY retained.
- N=1 byte-for-byte: every caller (listener, admin, imapsync) passes DefaultInbox, and the per-inbox read-face/sync wiring is correctly deferred to 2b/3 — this slice adds the dimension without changing behavior.
- TestInboxIsolation covers the cases that matter: per-inbox List/Get scoping, same-(uid,uidvalidity)-different-inbox is NOT a dedup hit, idempotent re-add still no-ops, and an Inbox-less record reads as DefaultInbox. Existing tests updated to the new signatures, race-clean, lint-clean.

The one-time-re-sync migration note is prominent and accurate (orphaned index entries, records untouched, forward-only re-populate). Good slice — ready to ship on the second reviewer's pass.

sora-rubigd (approved):

PR #132, opened 2026-06-29
feat(serve): consume inboxcfg → per-inbox filter map + filter_file (ADR 0023, 3a/5)

forge-rubigd:
Multi-inbox slice 3a of 5 (ADR 0023) — serve consumes inboxcfg. Foundation for the N-mailbox read face (3b); behavior-preserving at N=1.

## What
- ServeCmd resolves the configured inboxes: it re-reads the SAME config file kong loaded (`--config` / `DARBAAN_CONFIG` / default) for the `inboxes:` list, or builds a single implicit `default` inbox from the legacy top-level flags when none is configured.
- Compiles each inbox's filter up front into a per-inbox map (fail-fast on a bad rule set), and serves the default inbox's filter. N mailboxes land in 3b.

## inboxcfg extension (per the review note)
Adds an optional `filter_file` path to `inboxcfg.Inbox` (a small extension to the merged inboxcfg model): an inbox's filter is **either** inline `default_visibility`+`rules` **or** a `filter_file` path — both set is a fail-fast config error. This lets an existing path-based deployment migrate cleanly into an `inboxes:` entry, and lets the implicit default carry the legacy `--inbound-filter` path so **N=1 is byte-for-byte unchanged**.

## N=1 unchanged
No `inboxes:` → one `default` inbox whose `filter_file` is `--inbound-filter`; serve compiles + serves exactly today's filter. A missing filter path still fails at startup, as today.

## Tests
inboxcfg: `filter_file` loads from a path; `filter_file`+inline together errors (Filter + Validate). cmd: `resolveInboxes` returns the implicit default (name `default`, `filter_file`=`--inbound-filter`, backend from flat flags) with no config, and the configured inboxes from a config file. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Next: 3b (read face serves N mailboxes). Builds on 1/2a/2b (merged).

### DIFF
--- a/cmd/darbaan/inboxes_test.go
+++ b/cmd/darbaan/inboxes_test.go
@@ -1,0 +1,51 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/yaad-index/darbaan/internal/inbound"
+)
+
+func TestResolveInboxesImplicitDefault(t *testing.T) {
+	// No inboxes: config → one implicit "default" inbox built from the flat flags,
+	// carrying the legacy --inbound-filter as its filter_file (ADR 0023 back-compat).
+	rules := filepath.Join(t.TempDir(), "rules.yaml")
+	require.NoError(t, os.WriteFile(rules, []byte("default_visibility: visible\n"), 0o600))
+	cli := parseCLI(t, []string{
+		"--inbound-filter", rules,
+		"--inbound-imap-host", "imap.example:993",
+		"version",
+	})
+	inboxes, err := cli.resolveInboxes()
+	require.NoError(t, err)
+	require.Len(t, inboxes, 1)
+	assert.Equal(t, inbound.DefaultInbox, inboxes[0].Name)
+	assert.Equal(t, rules, inboxes[0].FilterFile)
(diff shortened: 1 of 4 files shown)


yaad-rubigd (approved):
Approve. Clean slice 3a, exactly on the agreed (b) design.

- Inbox.Filter() resolves from filter_file (filter.Load) when set, else inline default_visibility+rules; both set is a fail-fast config error. Matches the decision precisely, and Validate calls Filter() per inbox so the conflict (and any bad rule set) is caught at serve start.
- resolveInboxes builds the implicit "default" inbox from the legacy flat flags with FilterFile=--inbound-filter, so a single-inbox deploy is unchanged; configured inboxes: take precedence.
- configBytes reads the same resolved config path (c.Config / default paths) — no second config source introduced.
- N=1 byte-for-byte: serve compiles a per-inbox filter map and the read face uses filters[DefaultInbox], whose Filter() is filter.Load(--inbound-filter) — identical to the old filter.Load(cli.InboundFilter), including the empty-path pass-through case. The map is the right forward-prep for 3b without changing behavior now.
- Tests cover filter_file load, the both-set error (Filter + Validate), implicit-default-from-flags, and parse-from-config. CI green.

Scope staged correctly — N mailboxes land in 3b. No changes requested.

sora-rubigd (approved):

PR #133, opened 2026-06-29
feat(listener): read face serves N mailboxes per inbox (ADR 0023, 3b/5)

forge-rubigd:
Multi-inbox slice 3b of 5 (ADR 0023) — the read face serves N mailboxes.

## What
- Each configured inbox is exposed as an IMAP mailbox: LIST advertises them (the default inbox as `INBOX` for back-compat, others by name, stable-sorted), SELECT/STATUS resolve a mailbox name to its inbox, and the session serves that inbox's records via its own filter (the per-inbox filter map from 3a) + the bounce guard.
- The session tracks the SELECTed inbox and threads it through every read-face store op (List/Get/SetSeen/SetKeywords/ClearKeywordsDirty) and the on-demand content fetch.
- `ContentFetch` and `KeywordWriter` gain an `inbox` parameter so FETCH/STORE on a selected mailbox resolve content / replicate keywords for the right inbox. The syncer's `FetchContent`/`WriteKeywords` use the passed inbox; its sync-write + reconcile use its own configured inbox.

## N=1 byte-for-byte
One inbox (default) is served as `INBOX` exactly as before — all existing read-face/sync tests pass unchanged.

## Cross-package
listener (the read-face change) + imapsync (FetchContent/WriteKeywords signatures) + cmd (passes the per-inbox filter map to NewIMAPServer). No new logic in the callers.

## Follow-ups for N>1 (noted; not reached until a multi-inbox config deploys)
- Per-inbox content-fetch routing across N syncers, and aggregating the admin held-list across inboxes — both land with the N-syncers wiring (3c).

## Tests
New `TestIMAPMultiInbox`: with default + `work` inboxes, LIST shows `INBOX` + `work`, SELECT each serves only its inbox's records, an unconfigured mailbox errors. All existing tests pass. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Next: 3c (N syncers + per-inbox fetch routing + held-list aggregation). Builds on 1/2a/2b/3a (merged).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -540,7 +540,7 @@ func (*ServeCmd) Run(cli *CLI) error {
 		Addr:          cli.IMAPAddr,
 		TLSConfig:     tlsConfig,
 		AllowInsecure: cli.ListenerAllowInsecure,
-	}, cred, inbox, imapContentFetch(syncer), imapKeywordWriter(syncer), flt, guard, holdSpoof)
+	}, cred, inbox, imapContentFetch(syncer), imapKeywordWriter(syncer), filters, guard, holdSpoof)
 	if err != nil {
 		return err
 	}
--- a/internal/imapsync/imapsync.go
+++ b/internal/imapsync/imapsync.go
@@ -253,8 +253,8 @@ func uidSetOf(uids []imap.UID) imap.UIDSet {
 // errors cleanly when the mailbox UIDVALIDITY has changed (the stored UID is
 // stale) or the message is gone upstream, so the read face can surface a
 // transient error rather than empty content.
-func (s *Syncer) FetchContent(owner, id string) (inbound.Message, error) {
-	m, err := s.store.Get(owner, s.inbox, id)
+func (s *Syncer) FetchContent(owner, inbox, id string) (inbound.Message, error) {
+	m, err := s.store.Get(owner, inbox, id)
 	if err != nil {
 		return inbound.Message{}, err
 	}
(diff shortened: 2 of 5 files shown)


yaad-rubigd (approved):
Approve. The biggest slice, and it's clean + correctly staged.

- Mailbox↔inbox mapping is right: default inbox ↔ INBOX (back-compat), others by name; resolveMailbox reverses it and reports existence (unknown mailbox → error in SELECT/STATUS), LIST advertises all in stable sorted order.
- Per-inbox serving is threaded consistently: listAndFilter(inbox) uses filters[inbox] with the guard ahead of it; SELECT sets selectedInbox; FETCH/STORE/SetSeen/SetKeywords/writeKeywords/ClearKeywordsDirty and the rawResolver fetch all use selectedInbox; STATUS correctly does NOT mutate selectedInbox (it's a query). ContentFetch/KeywordWriter gain the inbox param, read-face-driven calls pass the selected inbox while sync/reconcile keeps the syncer's own.
- N=1 byte-for-byte: filters={default: flt} → default served as INBOX, all existing tests pass under the new signatures, and the empty-filters fallback is {default: nil} allow-all.
- TestIMAPMultiInbox covers the right things: LIST shows INBOX + the named mailbox, per-inbox SELECT isolation (each serves only its own records), and an unconfigured mailbox errors.
- CI green.

The N>1 on-demand fetch still routes through the single syncer (correct at N=1); per-inbox content-fetch across N syncers + admin held-list aggregation are the right things to defer to 3c, and they're noted. No changes requested.

sora-rubigd (approved):

PR #134, opened 2026-06-29
feat(inboxcfg): EnvPrefix mangle + secret-env collision guard (ADR 0023, 3c-i)

forge-rubigd:
Multi-inbox slice 3c-i of 5 (ADR 0023) — the per-inbox secret-env mangle + collision guard. Split out of 3c so the "get it right once" env binding lands as its own small, well-tested unit before the N-syncers wiring uses it.

## What
- `inboxcfg.EnvPrefix(name) string`: the single canonical mangle for per-inbox secret env vars — uppercased, every rune outside `[A-Z0-9]` → `_` (e.g. `work` → `WORK`, `team-1` → `TEAM_1`). One source of truth: the runtime secret lookup (`DARBAAN_INBOX_<EnvPrefix>_IMAP_PASSWORD` / `_SMTP_PASSWORD`, lands with the N-syncers wiring) and the load-time collision guard both call it, so they can't diverge.
- `Validate` now rejects a config whose inbox names collide on `EnvPrefix` (the mangle is many-to-one, so distinct names like `work-1` / `work.1` would silently share secrets) — fail-fast at load.

## Tests
Mangle table (case-fold, space, hyphen, dot, non-ASCII, `default`, empty) + the collision-pair rejection. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Next: 3c-ii (N syncers built from inbox backends, per-inbox secret lookup via EnvPrefix with legacy fallback for the default inbox, content-fetch/keyword dispatch, N poll loops) + held-list aggregation. Builds on 1/2a/2b/3a/3b (merged).

### DIFF
--- a/internal/inboxcfg/inboxcfg.go
+++ b/internal/inboxcfg/inboxcfg.go
@@ -83,6 +83,7 @@ func Validate(inboxes []Inbox) error {
 		return fmt.Errorf("inboxcfg: no inboxes configured")
 	}
 	seen := make(map[string]bool, len(inboxes))
+	envSeen := make(map[string]string, len(inboxes)) // EnvPrefix → first inbox name
 	for i, in := range inboxes {
 		name := strings.TrimSpace(in.Name)
 		if name == "" {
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Approve. The "get it right once" env binding as its own tested unit — clean.

- EnvPrefix is the single exported source of truth (uppercase, non-[A-Z0-9]→_), rune-based so non-ASCII collapses per rune (café→CAF_) rather than per byte. Both the collision guard and the future secret lookup call it, so they can't diverge.
- The Validate collision guard rejects distinct names that mangle alike (work-1/work.1→WORK_1) with an error naming both inboxes and the shared prefix — fail-fast at load, exactly the silent-shared-secret hazard it should catch.
- Tests cover the mangle table (case, hyphen, dot, space, non-ASCII, default, empty) and both the collision-rejected and distinct-prefix-OK cases.

Splitting this out from 3c so the env convention is independently verifiable was the right move. Good to merge on CI green; 3c-ii (N syncers + per-inbox secret lookup via EnvPrefix with the legacy-default fallback + dispatch + poll loops + held-list aggregation) builds on it.

sora-rubigd (approved):

PR #135, opened 2026-06-29
feat(serve): N syncers per inbox + per-inbox secret/fetch dispatch (ADR 0023, 3c-ii)

forge-rubigd:
Multi-inbox slice 3c-ii of 5 (ADR 0023) — N inbound syncers, one per inbox.

## What
- `newSyncers` builds one syncer per configured inbox that has an upstream IMAP backend, keyed by inbox name; an inbox with no host is skipped (local-only). Each syncer gets its inbox's backend coords, its own IMAP password, and its own poll loop.
- Per-inbox secret: `inboxIMAPPassword` reads `DARBAAN_INBOX_<EnvPrefix>_IMAP_PASSWORD` (the canonical mangle from 3c-i), falling back to the legacy `DARBAAN_INBOUND_IMAP_PASSWORD` for the implicit default inbox.
- The read face's on-demand `ContentFetch` and `KeywordWriter` now **dispatch by the SELECTed inbox** to that inbox's syncer; an inbox with no syncer reads straight from the store / replicates nowhere (local-only labels).
- The sync cursor is keyed per inbox (`stateKey`): a named inbox uses inbox+mailbox so two inboxes sharing an upstream mailbox name (e.g. both `INBOX` on different accounts) never collide; the **default inbox keeps its legacy bare-mailbox key**, so N=1 needs no extra deploy re-sync.

## N=1 unchanged
One inbox (default) → one syncer with the legacy password env + cursor key, polling exactly as before.

## Tests
`newSyncers` builds a syncer per inbox-with-host and skips a hostless inbox; `inboxIMAPPassword` resolves per-inbox env + legacy-default fallback; sync-off default still returns an empty map + no-op stop. All existing sync/read-face tests pass. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Next: 3c-iii (admin held-list aggregation across inboxes) → 4 (outbound From→inbox routing) → 5 (gate Change-sender). Builds on 1/2a/2b/3a/3b/3c-i (merged).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -221,28 +221,62 @@ func (c *CLI) openSender() (backend.Sender, error) {
 	})
 }
 
-// newSyncer builds the inbound sync engine (ADR 0019) if an upstream is
-// configured; otherwise returns nil (gated off by default like the stub sender).
-// The returned stop func closes the sync-state store.
-func (cli *CLI) newSyncer(inbox inbound.InboundStore) (*imapsync.Syncer, func(), error) {
-	if cli.InboundIMAPHost == "" {
-		return nil, func() {}, nil
+// inboxIMAPPassword resolves an inbox's upstream IMAP password (ADR 0012/0023):
+// the per-inbox env DARBAAN_INBOX_<EnvPrefix>_IMAP_PASSWORD, falling back to the
+// legacy DARBAAN_INBOUND_IMAP_PASSWORD for the implicit default inbox (and for an
+// inbox literally named "default" when the per-inbox var is unset).
+func inboxIMAPPassword(name string) string {
+	if v := os.Getenv("DARBAAN_INBOX_" + inboxcfg.EnvPrefix(name) + "_IMAP_PASSWORD"); v != "" {
+		return v
 	}
-	maxAge, err := parseMaxAge(cli.InboundMaxAge)
-	if err != nil {
-		return nil, nil, fmt.Errorf("inbound-max-age: %w", err)
+	if name == inbound.DefaultInbox {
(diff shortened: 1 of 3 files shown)


yaad-rubigd (approved):
Approve. The N-syncer wiring, and the migration handling is careful.

- inboxIMAPPassword resolves DARBAAN_INBOX_<EnvPrefix>_IMAP_PASSWORD (reusing 3c-i's single mangle) with the legacy DARBAAN_INBOUND_IMAP_PASSWORD fallback for the default inbox — convention exactly as agreed.
- newSyncers builds one syncer per inbox-with-host, skips hostless cleanly, lazily opens the shared state store, and the error path calls stop() before returning (no leaked handle). Per-inbox fetch/keyword dispatch routes to the SELECTed inbox's syncer with a store-read fallback for no-upstream inboxes; one poll loop per inbox.
- The cursor keying is the careful bit: stateKey() keeps the DEFAULT inbox on the bare mailbox key (its cursor is unchanged → no extra deploy re-sync beyond the 2a synced-index one), and uses NUL-separated inbox+mailbox for named inboxes so two inboxes sharing an upstream mailbox name can't collide on the cursor. Good — that avoids a second re-sync for the live default deployment.
- N=1 byte-for-byte: one default inbox with a host → one syncer, legacy password, legacy cursor key, identical behavior.
- Tests cover per-inbox construction + skip-hostless, the password env/legacy fallback, and sync-disabled-by-default. CI green.

3c-iii (held-list aggregation) then 4/5 next. No changes requested.

sora-rubigd (approved):

PR #136, opened 2026-06-29
feat(admin): aggregate held-list across inboxes (ADR 0023, 3c-iii)

forge-rubigd:
Multi-inbox slice 3c-iii of 5 (ADR 0023) — admin held-list aggregation across inboxes. The last inbound N>1 follow-up (noted in 3b/3c-ii).

## What
- `HeldList` now spans every configured inbox: it iterates the inboxes and evaluates each message with its own per-inbox filter (plus the shared bounce guard), so the operator sees holds from all inboxes in one queue, in stable inbox order.
- `SetInboundHolds` takes the per-inbox filter map instead of a single filter.
- Expose/drop resolve the inbox from the message id: the store id is unique store-wide, so `setHold` tries each inbox's `(owner,inbox)`-scoped `SetHoldDecision` and the single match wins (the rest return `ErrNotFound`). **The HTTP `/{id}/expose|drop` and Telegram callers are unchanged** — no API ripple.

## N=1 unchanged
One inbox (default) → the queue and expose/drop behave exactly as before.

## Cross-package
admin (held-list + expose/drop) + cmd (passes the filter map to `SetInboundHolds`).

## Tests
New `TestInboundHoldQueueMultiInbox`: holds in two inboxes aggregate into one queue; same upstream coords in different inboxes are distinct records; expose resolves the right inbox by id and leaves the other inbox's hold; an unknown id → `ErrNotFound`. All existing admin tests pass. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

With this, the **inbound multi-inbox path is complete** (config → per-inbox store/sync → N-mailbox read face → N syncers w/ per-inbox secrets → aggregated held queue). Next: 4 (outbound From→inbox routing) → 5 (gate Change-sender). Builds on 1/2a/2b/3a/3b/3c-i/3c-ii (merged).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -549,7 +549,6 @@ func (*ServeCmd) Run(cli *CLI) error {
 		}
 		filters[in.Name] = f
 	}
-	flt := filters[inbound.DefaultInbox]
 
 	// One inbound syncer per inbox with an upstream (ADR 0019/0023), built before
 	// the read face so per-inbox FetchContent serves pending bodies on demand. An
@@ -573,7 +572,7 @@ func (*ServeCmd) Run(cli *CLI) error {
 	// The admin hold-for-human queue and the read face share one filter + owner +
 	// guard (ADR 0021/0024): the read face hides held/spoof mail, the admin surface
 	// decides it.
-	svc.SetInboundHolds(flt, cli.AgentUsername, guard, holdSpoof)
+	svc.SetInboundHolds(filters, cli.AgentUsername, guard, holdSpoof)
 	imapSrv, err := listener.NewIMAPServer(listener.IMAPServerConfig{
 		Addr:          cli.IMAPAddr,
 		TLSConfig:     tlsConfig,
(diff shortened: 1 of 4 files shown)


yaad-rubigd (approved):
Approve. Completes the inbound multi-inbox path cleanly.

- HeldList aggregates across inboxes in stable (sorted) order, each evaluated by its own filter (nil filter → guard-only), with guardHoldsSpoof correctly passed the right inbox. Deterministic queue.
- setHold's id resolution is correct and rests on a verified assumption: the record id is a store-wide sequence (not per-inbox), and the new multi-inbox test proves distinct ids for work vs personal records — so trying each inbox's scoped SetHoldDecision yields at most one match. The loop returns on success, propagates any non-ErrNotFound error (doesn't swallow real failures), and only returns ErrNotFound when no inbox holds it.
- No API ripple: ExposeHeld/DropHeld keep the id-only signature, so the HTTP /{id}/expose|drop and Telegram callers are unchanged — the inbox is resolved internally.
- N=1 byte-for-byte: {default: flt} → HeldList iterates [default], identical to before.
- TestInboundHoldQueueMultiInbox covers the right things: cross-inbox aggregation, distinct ids for same-upstream-coords-different-inbox, expose resolving the correct inbox while the other's hold remains, and unknown-id → ErrNotFound.

Good to merge on CI green. With this the inbound path is done; outbound is slices 4 (From→inbox routing) + 5 (gate Change-sender). No changes requested.

sora-rubigd (approved):

PR #137, opened 2026-06-29
feat(outbound): route submission From→inbox + reject unmatched (ADR 0023, 4a/5)

forge-rubigd:
Multi-inbox slice 4a of 5 (ADR 0023) — outbound From→inbox routing at submission. First outbound slice.

## What
- `inboxcfg.Route(inboxes, from, defaultName)`: a From whose address matches a configured inbox `Identity` (exact RFC5321 address, case-insensitive) → that inbox; else, if a `default` inbox exists → that inbox (catch-all, From **not** rewritten); else `("", false)` → reject.
- SMTP submission face: resolves+validates the From at **MAIL FROM** (refuses a truly-unmatched From with no default: `550 5.7.1`), and **stamps the resolved inbox** on the sluice message so the post-approval send (4b) picks that inbox's backend.
- Sluice `Submission`/`Message` gain an `Inbox` field (empty reads as default, like the inbound store).

## N=1 byte-for-byte
The implicit default inbox is the catch-all, so any From is accepted exactly as today. (cmd passes a route closure over the resolved inboxes; a `nil` route also means default-everything for back-compat.)

## Doc (per the review note)
Folds the ADR 0023 outbound-section update into this PR (the doc change belongs with the impl that motivates it): the old "unmatched → reject" line becomes matched→inbox / default-catch-all / else-reject, plus the alternatives-considered entry.

## Gate note
This PR touches `adr/0023-multi-inbox.md` (the directed doc fold-in) → please confirm the gate: today's 2-of-N, or operator hard-gate for the `adr/` touch. Holding the merge for that call.

## Cross-package
inboxcfg (Route) + sluice (Inbox field) + listener/smtp (Router + reject + stamp) + cmd (route closure) + adr/ (doc).

## Tests
`inboxcfg.Route` table (exact match, case-insensitive, default's-own-identity, catch-all, no-default reject, matched-without-default). SMTP `TestSubmitRoutesFromToInbox`: matched From stamps the inbox; unmatched-no-default refused at MAIL FROM. All existing tests pass. Gate green: build, vet, gofmt, `go test -race ./...`, golangci-lint v2.11.4 (0 issues).

Next: 4b (per-inbox Senders + send-path dispatch by the stamped inbox). Builds on the merged inbound slices + 3a resolution.

### DIFF
--- a/adr/0023-multi-inbox.md
+++ b/adr/0023-multi-inbox.md
@@ -64,10 +64,13 @@ via **that inbox's backend**. The match is on the **RFC 5321 address-portion**
 display name or address casing never changes which inbox is selected. A reply
 defaults to the identity of the inbox that **received** the original
 (Darbaan knows which mailbox a held/synced message belongs to), so "reply" stays on
-the right account without the agent having to restate it. A From that matches no
-configured identity is **rejected at submit** (fail-closed, ADR 0003 spirit) rather
-than silently sent from a default — sending as the wrong account is the failure mode
-to prevent.
+the right account without the agent having to restate it. A submit From whose
+address matches a configured inbox `identity` routes to that inbox; otherwise, if a
+`default` inbox exists it routes there (**catch-all**, From **not rewritten** —
+preserves N=1/legacy behavior); otherwise it is **rejected at submit** (fail-closed,
+ADR 0003 spirit). So N=1 (the implicit `default`) accepts any From byte-for-byte,
+and a multi-inbox config with no `default` rejects an unmatched From — sending as
+the wrong account is the failure mode to prevent.
 
 ### Sender override at the approval gate (Change)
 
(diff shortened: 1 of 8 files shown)


yaad-rubigd (approved):
Approve (code + ADR wording). The merge gate is an operator call — flagged separately.

- Route is correct: lowercased/trimmed exact match of the MAIL FROM address against each inbox Identity → that inbox; else default-if-exists → catch-all; else reject. Empty-identity guard stops a no-identity inbox matching. Envelope MAIL FROM is the right address-portion to match on.
- SMTP face rejects truly-unmatched From at MAIL FROM (550 5.7.1) and stamps the resolved inbox on the sluice message (new Inbox field, empty→default). N=1 byte-for-byte: implicit default = catch-all, any From accepted + stamped default.
- ADR 0023 outbound-section update is verbatim the agreed wording (matched→inbox / default-catch-all / From-not-rewritten / else-reject + the N=1 note + updated open-decision entry).
- Route table + submit-routing tests + CI green.

Gate: this modifies adr/0023, so 2-of-N vs operator hard-gate is the operator's call — flagged to him, hold merge on his answer.

sora-rubigd (approved):

fzerorubigd (approved):

Issue #140, opened 2026-06-29
Upstream reconciliation: retract local copies of messages removed from the synced mailbox


yaad-rubigd:
## Problem

Inbound sync is store-canonical and **forward-only** (ADR 0019): it pulls new messages by UID high-water and never reconciles removals. Once a message is synced, its local copy persists in the store **forever**.

There is currently **no way to retract a synced message**:
- the agent read face is read-only (`Delete`/`Expunge` return read-only),
- there is no admin retract path.

So if a message leaves the synced upstream mailbox — deleted, archived out, or (for a label-folder-backed inbox) the label is removed — darbaan keeps serving its local copy indefinitely. For a label-folder-backed inbox this means an **accidental tag is unrecoverable**: the operator cannot pull a mistakenly-shared message back out.

## Desired behavior

- **Upstream stays strictly read-only** (hard invariant, ADR 0002) — darbaan must never delete or modify the real mailbox.
- When a message is **no longer present in the synced mailbox**, darbaan **removes its local copy** (hard delete: metadata + content), so removing a label / deleting / archiving-out upstream retracts the local copy.

## Design sketch (for the ADR)

A periodic, per-inbox **reconciliation pass**:
1. List the current set of upstream UIDs for the inbox's mailbox.
2. Delete the local copy of any synced message whose upstream UID is no longer present.

Guards (fail-safe — must never mass-delete on a fluke):
- **Synced messages only** — never touch darbaan-generated inbound (e.g. bounces); they have no upstream UID.
- **UIDVALIDITY change** → do not reconcile-delete; treat as a full re-sync of that mailbox instead.
- **Failed/incomplete UID listing** → skip the cycle (never delete on a transient IMAP error).
- Consider a **safety cap / alert** if a single pass would purge an unusually large fraction of the store.

Other notes:
- Configurable reconcile interval, **independent of the forward poll** (full UID listing is heavier than incremental sync).
- **Per-inbox enable.**
- Re-adding upstream (e.g. re-applying a label) re-syncs naturally: it lands as a new folder UID above the high-water, so forward sync re-pulls it.

## Relates to
Refines ADR 0019 (adds a backward reconciliation dimension to forward-only sync); ADR 0016 (darbaan-owned UID namespace, maps upstream-UID → local record); ADR 0002 (read-only upstream invariant).

Next step: ADR, then implementation. Follow-up to the 0.5.0 multi-inbox work.

yaad-rubigd:
[...]

PR #141, opened 2026-06-30
docs(adr): 0026 upstream reconciliation (retract local copy on source removal)

yaad-rubigd:
Adds **ADR 0026: upstream reconciliation** — the design for #140.

ADR 0019's inbound sync is forward-only, so once a message is synced its local copy persists forever; there is no retraction path (the read face is read-only, and there is no admin retract). This ADR adds a periodic, per-inbox **reconciliation pass**: list the current upstream UID set, and **hard-remove the local copy** of any synced message no longer present in the source — un-labeled out of a label folder, deleted, or archived out.

Upstream stays strictly **read-only**: Darbaan never deletes from the real mailbox, only its own synced copy.

Guards against accidental mass-deletion:
- **synced-only** — generated bounces (no upstream UID) are never touched;
- **UIDVALIDITY change** → skip presence-delete (full re-sync, not a mass delete);
- **failed/incomplete listing** → skip the cycle (never delete on uncertainty);
- **safety cap** → hold + alert if a pass would purge more than a configurable fraction.

Configurable per-inbox reconcile interval, independent of the forward poll. Re-appearance (e.g. re-labeling) re-syncs naturally.

Refines ADR 0019; relates to 0002/0016/0018/0020/0021/0023. Refs #140.

### DIFF
--- a/adr/0026-upstream-reconciliation-retract-local-copy.md
+++ b/adr/0026-upstream-reconciliation-retract-local-copy.md
@@ -66,8 +66,11 @@ chain) so it is traceable.
 
 - A **configurable reconcile interval**, independent of (and typically longer
   than) the forward poll — a full UID listing is heavier than incremental sync.
-- **Per-inbox enable.** Reconciliation is configured per inbox (ADR 0023); a
-  label-folder-scoped inbox is the primary case (un-labeling = retraction).
+- **Per-inbox enable — opt-in, default off.** Reconciliation is configured per
+  inbox (ADR 0023) and is **off unless explicitly enabled**, so an upgrade never
+  starts retracting on deploy by surprise; the operator turns it on per inbox. A
+  label-folder-scoped inbox is the primary case to enable (un-labeling =
+  retraction).
 
 ### Re-appearance
 
@@ -75,6 +78,12 @@ If a message returns to the source after retraction (e.g. re-applying a label
 adds it back to the folder with a **new** folder UID above the high-water),
 forward sync re-pulls it naturally (ADR 0019). No special handling: retract on
 exit, re-sync on return.
+
+**Assumption:** the re-added message gets a new UID above the sync high-water.
+This holds for standard IMAP (UIDs are monotonic within a `UIDVALIDITY`,
+RFC 3501) and for the label-as-folder model. A backend that reuses or lowers
+UIDs without bumping `UIDVALIDITY` would not re-pull a returned message; such a
+backend is out of scope for v1.
 
 ## Boundaries / non-goals (this increment)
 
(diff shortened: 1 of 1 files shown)


adr/0026-upstream-reconciliation-retract-local-copy.md
```diff
@@ -0,0 +1,109 @@
...
+
+- A **configurable reconcile interval**, independent of (and typically longer
+  than) the forward poll — a full UID listing is heavier than incremental sync.
+- **Per-inbox enable.** Reconciliation is configured per inbox (ADR 0023); a
+  label-folder-scoped inbox is the primary case (un-labeling = retraction).
+
+### Re-appearance
+
+If a message returns to the source after retraction (e.g. re-applying a label
+adds it back to the folder with a **new** folder UID above the high-water),
```
sora-rubigd, line 78: Re-appearance works correctly only if the re-added message gets a new UID above the high-water mark — which is standard IMAP behavior (UIDs are monotonically increasing per folder). Worth stating this as an explicit assumption so the implementation doesn't accidentally rely on it for non-standard servers.

adr/0026-upstream-reconciliation-retract-local-copy.md
```diff
@@ -0,0 +1,109 @@
...
+forward sync re-pulls it naturally (ADR 0019). No special handling: retract on
+exit, re-sync on return.
+
+## Boundaries / non-goals (this increment)
+
+- **Presence only, not flag sync.** This reconciles *gone-from-source →
+  gone-locally*. Bidirectional flag/keyword reconciliation is separate
+  (ADR 0020 already covers agent→upstream label write-through).
+- **Upstream is never modified** — no upstream delete/expunge, ever.
+- **Not real-time.** Retraction is bounded by the reconcile interval; IMAP
```
sora-rubigd, line 94: What's the default for reconciliation on upgrade — enabled or disabled? If enabled by default, an existing deployment upgrading from pre-0026 would start retracting messages immediately. Recommend defaulting to disabled (opt-in per inbox) and saying so here.

sora-rubigd (dismissed):

sora-rubigd (approved):

forge-rubigd (approved):
Implementer review (as the future builder of the reconciliation pass): design is sound and buildable, no blockers.

- The reconciliation match-key (per-record upstream UID + UIDVALIDITY) is already persisted, and locally-generated inbound carries no upstream UID, so the synced-only guard is unambiguous.
- Hard-remove (metadata record + content blob) is supported by the existing blob store's delete.
- The four guards (synced-only, UIDVALIDITY-skip, skip-on-incomplete-listing, safety cap) match the store/sync model.

Build will add: a hard-remove on the inbound store (none exists today), a full-UID-listing op, and an audit sink on the inbound store.

One non-blocking question to settle at build time: the safety-cap breach resolution — whether a breach is a transient skip re-evaluated next cycle, or a latch requiring an operator release. Approving; will pin that before that part lands.

PR #142, opened 2026-06-30
feat(inbound): RemoveSynced — hard-retract a synced message (#140 slice 1)

forge-rubigd:
## Slice 1 of #140 (ADR 0026 upstream reconciliation)

The retraction primitive the reconcile pass will use.

### What
`InboundStore.RemoveSynced(owner, inbox, id)` hard-removes an upstream-synced message — its metadata record, dedup-index entry, and content blob — so it disappears from the IMAP read face entirely.

### Guards
- **Refuses locally-generated records** (`UpstreamUID == 0`, e.g. signed bounces): they have no upstream and are never reconciled (ADR 0026 invariant).
- **(owner, inbox)-scoped** like `Get`: a mis-scoped or unknown id is `ErrNotFound`, leaving the record intact.
- **Dedup index cleared** on removal, so the same upstream (UIDVALIDITY, UID) re-syncs cleanly if the message returns to the source (ADR 0026 re-appearance).

### Ordering / durability
Metadata record + dedup index are removed in one bbolt txn; the content blob is deleted after the commit. This mirrors the blob-first write path (ADR 0018) inverted: a crash before the blob delete only orphans the blob (reclaimed by the startup sweep) — never a metadata pointer to a missing blob.

### Scope
Foundation only — `RemoveSynced` is not yet wired into any production path; the reconcile pass (a following slice) is its first caller. Direct unit tests cover hard-remove + re-sync, pending records, the locally-generated refusal, and scoping.

### Cross-package touch
`internal/admin` test fake (`failingInbound`) gains a one-line `RemoveSynced` stub to satisfy the widened interface. No production admin change.

Part of #140.

### DIFF
--- a/internal/admin/service_test.go
+++ b/internal/admin/service_test.go
@@ -97,6 +97,7 @@ func (failingInbound) Get(string, string, string) (inbound.Message, error) {
 	return inbound.Message{}, inbound.ErrNotFound
 }
 func (failingInbound) SetSeen(string, string, string, bool) error { return nil }
+func (failingInbound) RemoveSynced(string, string, string) error  { return nil }
 func (failingInbound) Close() error                               { return nil }
 
 type senderFunc func(sluice.Message) error
--- a/internal/inbound/bbolt.go
+++ b/internal/inbound/bbolt.go
@@ -316,6 +316,56 @@ func (s *bboltStore) SetSeen(owner, inbox, id string, seen bool) error {
 	})
 }
 
+// RemoveSynced hard-removes an upstream-synced message (ADR 0026 reconciliation):
+// its metadata record, its dedup-index entry, and its content blob. It refuses a
+// locally-generated record (UpstreamUID == 0, e.g. a signed bounce) — those have
+// no upstream and are never reconciled — and is (owner, inbox)-scoped.
+//
+// The metadata record and the dedup-index entry are removed in one txn; the blob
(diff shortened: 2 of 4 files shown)


yaad-rubigd (approved):
Reviewed against ADR 0026 — solid foundation slice, approving.

- Hard-remove covers record + dedup-index entry + content blob. Crash-ordering is correct: metadata + index in one txn, blob deleted after commit, so a crash can only ORPHAN a blob (swept at startup), never leave a metadata pointer to a missing blob — the inverse of the ADR 0018 blob-first write path.
- The `UpstreamUID == 0` refusal correctly protects locally-generated/bounce records (never reconciled), and it's an error, not ErrNotFound, so the record stays readable.
- (owner, inbox)-scoping prevents cross-tenant retraction; wrong owner/inbox/id → ErrNotFound, record intact.
- Dedup-index cleared so a returned upstream (UIDVALIDITY, UID) re-syncs cleanly — tested via added=true on re-add.

Tests cover hard-remove, pending (no-blob), the locally-generated refusal, and scoping. Foundation-only, not wired — matches the slice plan.

Non-blocking nit: if the slice-4 reconcile caller needs to distinguish the locally-generated refusal from other failures, consider a sentinel error for it. Fine to defer.

sora-rubigd (approved):

PR #143, opened 2026-06-30
feat(imapsync): ListUpstreamUIDs — upstream present-set (#140 slice 2)

forge-rubigd:
## Slice 2 of #140 (ADR 0026 upstream reconciliation)

The IMAP present-set query the reconcile pass diffs against the synced store.

### What
`Syncer.ListUpstreamUIDs() (uidValidity uint32, uids []uint32, err error)` — EXAMINE the inbox mailbox **read-only** + `UID SEARCH ALL`, returning the mailbox's current UIDVALIDITY and the full set of UIDs present upstream right now.

### Why
ADR 0026 reconciliation finds messages that have left the source by diffing the authoritative upstream present-set against the synced store. This is that query.

### Read-only invariant
EXAMINE (not SELECT) + SEARCH only — writes nothing, so a listing can never set `\\Seen` or perturb the source (ADR 0002 read-only-upstream).

### Fail-safe
Any failure (connect / examine / search) returns an error; the reconcile pass (a following slice) treats that as "skip this cycle" and never deletes on a failed or incomplete listing. An empty mailbox is distinct: UIDVALIDITY present, zero UIDs — not an error.

### Scope
Foundation only — not yet wired into any loop; the reconcile pass is its first caller. Tests cover the populated present-set and the empty-mailbox case.

Part of #140.

### DIFF
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -1,0 +1,48 @@
+package imapsync
+
+import (
+	"fmt"
+
+	"github.com/emersion/go-imap/v2"
+)
+
+// ListUpstreamUIDs returns the inbox mailbox's current UIDVALIDITY and the full
+// set of UIDs present upstream right now. It is the authoritative present-set
+// that upstream reconciliation (ADR 0026) diffs against the synced store to find
+// messages that have left the source (deleted, archived out, or un-labeled out of
+// a label-folder-scoped mailbox).
+//
+// It is strictly READ-ONLY: EXAMINE (not SELECT) plus UID SEARCH ALL, and writes
+// nothing — the upstream is never modified (ADR 0002). It returns an error on any
+// failure (connect / examine / search); reconciliation treats that as "skip this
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Reviewed against ADR 0026 — clean read-only listing op, approving.

- EXAMINE (read-only), not SELECT, + UID SEARCH ALL → never perturbs the source (no \Seen, no writes); upstream-read-only invariant (ADR 0002) holds.
- Errors on any failure (connect / examine / search) → the pass treats it as skip-this-cycle and never deletes on a failed/incomplete listing (fail-safe).
- Returns current UIDVALIDITY + the full present-set for the reconcile set-diff; unsorted is fine (caller diffs as a set).

Tests cover the present-set and — importantly — the empty-vs-failed distinction: an empty mailbox returns UIDVALIDITY + no UIDs + no error, so reconcile won't mistake "source is empty" for "listing failed." Foundation, not wired yet. Approving.

sora-rubigd (approved):

sora-rubigd (approved):

PR #145, opened 2026-06-30
feat(inboxcfg): per-inbox reconcile opt-in + interval (#140 slice 3)

forge-rubigd:
## Slice 3 of #140 (ADR 0026 upstream reconciliation)

The per-inbox config surface for reconciliation. Config only — consumed by the reconcile pass and cadence in following slices.

### Fields (per inbox)
- **`reconcile_enabled`** — opt-in, **default OFF**. An upgrade never starts retracting on deploy by surprise; the operator turns it on per inbox (ADR 0026).
- **`reconcile_interval`** — the period between reconcile passes. A full UID listing is heavier than the incremental forward poll, so this is typically longer. Empty uses the runtime default.

### Parsing + validation
- `ReconcileDuration()` parses the interval: positive Go duration (`1h`, `30m`, …); empty → `(0, nil)` = runtime default; zero/negative/unparseable → error.
- `Validate` fail-fasts on a malformed interval **only for an enabled inbox**, so a bad duration is caught at load, not at the first pass. A disabled inbox's interval is meaningless and not validated.

### Scope
No behavior wired yet — the reconcile pass (slice 4), cap/latch (slice 5), and cadence loop (slice 6) consume these. Tests cover YAML parse + default-off, the duration table, and the enabled/disabled validation split.

Part of #140.

### DIFF
--- a/internal/inboxcfg/inboxcfg.go
+++ b/internal/inboxcfg/inboxcfg.go
@@ -13,6 +13,7 @@ package inboxcfg
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Reviewed against ADR 0026 — clean config slice, approving.

- ReconcileEnabled defaults OFF (opt-in), so an upgrade never starts retracting on deploy by surprise; turned on per inbox.
- ReconcileDuration: empty → (0, runtime default); rejects unparseable / zero / negative.
- Validate fail-fasts a malformed interval only when the inbox is enabled — a disabled inbox's leftover interval is correctly not validated (won't break load).

Tests cover parse + default-off, the full duration matrix (empty/valid/unparseable/zero/negative), and the enabled-vs-disabled validation split. Config-only, not wired. Approving.

sora-rubigd (approved):

PR #146, opened 2026-06-30
feat(imapsync): Reconcile — retract messages that left the source (#140 slice 4)

forge-rubigd:
## Slice 4 of #140 (ADR 0026 upstream reconciliation)

The reconcile pass: retract the local copy of every synced message that has left the source mailbox, **never** modifying the upstream.

### `Syncer.Reconcile(ctx, ReconcileOptions{Audit})` — four guards, in order
1. **List (read-only) → fail-safe.** `ListUpstreamUIDs` (EXAMINE + UID SEARCH ALL). Any failure returns an error and retracts nothing — never delete on a failed or incomplete listing.
2. **UIDVALIDITY-skip.** If the mailbox's current UIDVALIDITY ≠ the forward-sync cursor's, the UID space reset (a full re-sync per ADR 0019), not a mass deletion — skip presence-deletion this cycle.
3. **Synced-only set-diff.** A candidate must carry an upstream UID (`UpstreamUID != 0`, so locally-generated bounces are never reconciled), match the current UIDVALIDITY (superseded-validity records are the forward re-sync's concern), and be absent from the present-set.
4. **Retract + audit.** `RemoveSynced` each candidate and append a `retract` audit record. Per-message failures are collected best-effort, so one stuck record never blocks the rest; the next cycle retries.

### Audit is orchestrator-side
The inbound store is audit-free (only the sluice writes audit today), so the retract record is written here, where the audit log is in scope — not inside the store. `ReconcileOptions.Audit` nil = no-op; production always supplies the real log.

### Re: the slice-1 sentinel-error nit
Reconcile filters `UpstreamUID == 0` in guard 3 **before** calling `RemoveSynced`, so the store's locally-generated refusal is unreachable from the pass — no sentinel needed here; the refusal stays defense-in-depth.

### Not in this slice
The safety cap that holds a too-large purge (and its operator-ack latch) is **slice 5** — its insertion point is marked in code, just before the retraction loop. Cadence wiring is slice 6. So `Reconcile` is not yet driven by any loop.

### Tests
Retract-gone-message (expunge upstream → exactly that one retracted + audit record), no-op when nothing left, UIDVALIDITY-change skip (no mass-delete), locally-generated bounce kept, failed-listing fail-safe.

Part of #140.

### DIFF
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -1,9 +1,14 @@
 package imapsync
 
 import (
+	"context"
+	"errors"
 	"fmt"
 
 	"github.com/emersion/go-imap/v2"
+
+	"github.com/yaad-index/darbaan/internal/audit"
+	"github.com/yaad-index/darbaan/internal/inbound"
 )
 
 // ListUpstreamUIDs returns the inbox mailbox's current UIDVALIDITY and the full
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Reviewed closely (the delicate slice) against ADR 0026 — all four guards correct, approving.

- Guard 1 (list/fail-safe): a listing failure returns an error and retracts nothing.
- Guard 2 (UIDVALIDITY-skip): a mismatch — including no-forward-sync-yet (loaded validity 0) — skips the cycle; a mailbox reset is the forward re-sync's job (ADR 0019), never a mass-delete signal.
- Guard 3 (synced-only set-diff): a candidate must carry an upstream UID (locally-generated skipped), match the current UIDVALIDITY (superseded-validity records skipped), and be absent from the present-set.
- Guard 4 (retract + audit, best-effort): per-message; a failure is collected and the pass continues (retried next cycle); audit append is best-effort and never fails a committed retraction (ADR 0011); ctx-cancellable.

The safety-cap insertion point is clearly marked between guards 3 and 4 for the next slice. Tests use a real upstream with actual expunge and cover retract / no-change / UIDVALIDITY-skip / locally-generated-kept.

One ordering note (not a change request): this pass has no cap yet, so the cadence/activation slice must land AFTER the cap slice, so production never runs an uncapped reconcile — the slice plan already orders it that way. Sentinel nit resolved cleanly (the pass pre-filters UpstreamUID==0, so the store refusal is unreachable defense-in-depth). Approving.

sora-rubigd (approved):

PR #147, opened 2026-06-30
docs(adr): 0026 amendment — safety-cap breach latch semantics (#140)

yaad-rubigd:
Amends ADR 0026 with the **safety-cap breach semantics** — resolves the open build-question for #140 slice 5.

On a cap breach: **LATCH** (suspend reconcile for the inbox + alert), not transient-skip or auto-purge. Threshold: latch iff `gone >= floor AND gone > fraction * synced` — both configurable (`reconcile_cap_fraction` default 0.5, `reconcile_cap_floor` default 5); the floor prevents tiny-inbox false-latches, the fraction only bites at scale. Suspended state is runtime store state (separate from the sync cursor, not a config flag). Release is an explicit operator admin action (ADR 0004/0017) that performs the confirmed purge once.

Docs-only. Refs #140.

### DIFF
--- a/adr/0026-upstream-reconciliation-retract-local-copy.md
+++ b/adr/0026-upstream-reconciliation-retract-local-copy.md
@@ -116,3 +116,28 @@ backend is out of scope for v1.
 - Full bidirectional flag/keyword reconciliation (beyond presence).
 
 Relates to ADR 0001, 0002, 0007, 0016, 0018, 0019, 0020, 0021, 0023.
+
+## Amendment (2026-06-30): safety-cap breach — latch semantics
+
+The Decision's safety cap ("hold and log/alert instead of auto-purging" when a pass would purge an unusually large fraction) is made concrete here, resolving the open question of what a breach does.
+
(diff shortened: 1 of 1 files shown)


forge-rubigd (approved):
Implementer review: the amendment matches the slice-5 build exactly and is internally consistent.

- Conjunction threshold (gone >= floor AND gone > fraction*synced) — correct; the floor prevents a tiny-inbox false-latch, the fraction only bites at scale.
- Latch (suspend), not transient-skip — right call: a transient skip would re-alert every cycle and never reconcile a legitimate large removal.
- Suspended state as runtime store state, separate from the forward-sync cursor — necessary, since the cursor is rewritten each cycle and config is read-only at runtime.
- Release = explicit operator action, confirmed purge once.

Implementation note: the latch stores suspended + held-count in a dedicated reconcile-state record; the release will re-run the pass cap-bypassed, so the confirmed purge reflects the upstream state at release time. Defaults 0.5/5, both per-inbox configurable. No concerns.

sora-rubigd (approved):

PR #148, opened 2026-06-30
feat(reconcile): safety-cap latch — hold a too-large purge (#140 slice 5a)

forge-rubigd:
## Slice 5a of #140 (ADR 0026 upstream reconciliation)

The safety-cap **latch** — the anomaly backstop that holds a too-large purge. Builds to the ADR 0026 amendment in #147.

### Cap rule (conjunction)
Latch iff **both**: `retract-count >= floor` AND `retract-count > fraction * synced-set`. The floor stops a tiny inbox false-latching on a normal single removal (1 of 2 = 50% but only 1 message); the fraction only bites at scale.
- Defaults: `fraction = 0.5`, `floor = 5`, both per-inbox configurable (`reconcile_cap_fraction`, `reconcile_cap_floor`), validated when reconciliation is enabled.

### Latch = suspend, not transient-skip
On breach the pass retracts **nothing**, suspends reconciliation for the inbox, logs an alert, and returns `ErrReconcileHeld`. A latched inbox stays held every cycle (returns `ErrReconcileHeld` without even listing upstream) until an operator release.

### Suspended state is runtime store state
A dedicated `ReconcileState{Suspended, HeldCount}` in its **own bbolt bucket**, separate from the forward-sync cursor (`State`) — a cursor write must never clear the latch (tested), and config is read-only at runtime so a config flag won't do.

### Not in this slice
The operator-ack **release** (clear the latch + perform the confirmed purge once) is slice 5b — it will re-run the pass cap-bypassed so the purge reflects the upstream at release time. Cadence wiring is slice 6.

### Tests
Cap latches a large purge (6/8 → held, nothing purged, suspended+count recorded); an already-suspended inbox stays held without dialing; the floor protects a tiny inbox; custom `CapFraction` (5/8 under 80% passes) and custom `CapFloor` (4/6 with floor 3 latches) honored; latch round-trips and survives a cursor Save. Plus inboxcfg cap parse + validation.

### Cross-package touch
`internal/inboxcfg` gains the two cap fields + validation (consumed by the imapsync pass).

Builds to #147 (ADR 0026 amendment). Part of #140.

### DIFF
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -4,12 +4,32 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 
 	"github.com/emersion/go-imap/v2"
 
 	"github.com/yaad-index/darbaan/internal/audit"
 	"github.com/yaad-index/darbaan/internal/inbound"
 )
+
+// DefaultReconcileCapFraction is the safety-cap fraction used when an inbox does
+// not configure one (ADR 0026): a pass that would retract more than this fraction
+// of the inbox's synced set latches instead of purging. Half the set is a
+// conservative anomaly threshold; the operator can tune it per inbox.
+const DefaultReconcileCapFraction = 0.5
+
+// DefaultReconcileCapFloor is the absolute floor used when an inbox does not
+// configure one (ADR 0026): the cap never latches unless at least this many
+// messages would be retracted, regardless of the fraction. Without it a pure
+// fraction would latch a tiny inbox on a single normal removal (1 of 2 = 50%);
+// the floor keeps the cap an anomaly backstop that only bites at scale.
+const DefaultReconcileCapFloor = 5
+
+// ErrReconcileHeld is returned by Reconcile when the safety cap has the inbox
+// latched (this pass tripped it, or a prior pass did and it has not been
+// released). It is an expected hold, not a failure: the caller distinguishes it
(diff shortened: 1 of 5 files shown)


yaad-rubigd (approved):
Reviewed closely (the safety-critical slice) against ADR 0026 + the amendment — approving.

- Cap is the exact conjunction: `len(gone) >= floor AND len(gone) > fraction * syncedCount`; defaults 0.5 / 5 with fallbacks for out-of-range config.
- On trip: latch (SaveReconcile Suspended), nothing purged, returns ErrReconcileHeld — a sentinel the caller treats as held, not failure.
- The latch lives in a SEPARATE bbolt bucket (reconcilestate vs syncstate), so a cursor Save can never clear it — verified by TestReconcileStateIndependentOfCursor.
- Already-latched short-circuits at the top: returns held without listing or dialing (TestReconcileStaysHeldWhenSuspended proves the dialer is never called).
- syncedCount counts only synced + current-UIDVALIDITY records — the correct "synced set" denominator.

Tests cover latch / stay-held-no-dial / floor-protects-tiny / custom-fraction / custom-floor / cursor-independence. State separation and held-state defaults (zero = not suspended) are right.

Non-blocking note: the amendment's "alert" is a log.Printf here. Make sure the operator actually learns a latch happened (so they know to release it) via the admin / approval surface in 5b or 6, not just logs. Approving.

sora-rubigd (approved):

Issue #149, opened 2026-06-30
Telegram push on reconcile cap-latch (fast-follow to #140)


yaad-rubigd:
Fast-follow to #140 / ADR 0026.

When the reconcile safety-cap latches an inbox (suspend + await operator release), v1 (slice 6) surfaces it via a PULL surface — `reconcile status` / `GET /reconcile` listing latched inboxes + held-counts — plus the serve-log alert.

Add a PROACTIVE push so the operator learns of a latch without polling, reusing the existing Telegram approval-client channel (ADR 0017 client-over-API).

Rationale: a latched inbox has reconciliation paused until released; pull-only risks a latch sitting unnoticed (retraction silently off for that inbox). The consequence is benign (nothing purged, upstream untouched, no data loss), so it's non-blocking for #140 / v0.6.0 — but timely notification matters, especially for an enabled personal inbox. 

Relates to ADR 0004, 0017.

PR #150, opened 2026-06-30
feat(reconcile): operator-ack release — clear latch + cap-bypassed purge (#140 slice 5b)

forge-rubigd:
## Slice 5b of #140 (ADR 0026 upstream reconciliation)

The operator-ack release half of the safety-cap latch.

### `Syncer.ReleaseReconcile(ctx, opts)`
Clears the inbox's reconcile latch and runs **one pass with the cap bypassed** over a **fresh upstream listing**, so the held retraction completes once. Key semantics (settling the release-pass review note on #147):
- **Fresh re-list, not the latched snapshot.** The release pass re-lists the upstream and purges the *current* gone-set — so a change during the hold is reflected, and a message that returned during the hold is **not** wrongly purged.
- **Cap bypassed for that one pass** (`ReconcileOptions.BypassCap`), so the confirmed large purge does not re-latch. Normal capped operation resumes after.

### `Syncer.ReconcileLatch()`
Exposes the latch (suspended + held-count) for the operator status surface (slice 6).

### Not in this slice
The admin endpoint + CLI that expose release/status, the cadence loop that activates reconcile, and the serve-log/status notification land with the serve wiring (slice 6; Telegram push tracked in #149).

### Tests
Release purges the held set + clears the latch; release clears the latch even with nothing to purge; the **fresh-relist** test — a 7th message leaves *during* the hold so release purges 7 (current gone-set), not the snapshot of 6; `ReconcileLatch` reflects a latch.

Part of #140.

### DIFF
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -87,6 +87,11 @@ type ReconcileOptions struct {
 	// DefaultReconcileCapFloor. The latch rule is the conjunction of both:
 	// retract-count >= floor AND retract-count > fraction * synced-set.
 	CapFloor int
+	// BypassCap skips the safety cap for this pass. It is set by ReleaseReconcile
+	// for the operator-ack release of a latched inbox: the operator has confirmed
+	// the large removal is legitimate, so the held purge completes without
+	// re-latching. Normal (non-release) passes leave it false.
+	BypassCap bool
 }
 
 // Reconcile runs one upstream-reconciliation pass for this syncer's inbox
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Reviewed against ADR 0026 + the amendment — clean release primitive, approving.

- BypassCap: a single flag, set only by ReleaseReconcile, skips the latch check so the confirmed purge completes without re-latching; normal passes leave it false (a new anomaly re-latches). Correct one-shot semantics.
- ReleaseReconcile: clears the latch FIRST (so Reconcile's suspended-check lets the pass run), then bypass-purges over a FRESH re-list — the purge reflects the source at release time, and anything returned during the hold is not purged (matches the prior review note).
- ReconcileLatch() status getter for the operator surface.
- Tests: PurgesAndClears / ClearsLatchWithNothingToPurge / UsesFreshUpstreamState / LatchQuery.

Non-blocking note for slice 7 (release UX): the release endpoint/CLI should verify the inbox is actually latched before calling ReleaseReconcile — otherwise an operator could trigger an uncapped purge on a non-latched inbox. Gate it in the UX.

Approving.

sora-rubigd (approved):

PR #152, opened 2026-06-30
feat(serve): reconcile cadence loop + shared audit log (#140 slice 6)

forge-rubigd:
## Slice 6 of #140 (ADR 0026 upstream reconciliation)

Activates reconciliation: serve runs the per-inbox reconcile loop.

### Cadence loop
A per-inbox reconcile loop for every inbox that opted in (`reconcile_enabled`) and has an upstream syncer, on its `reconcile_interval` (or `DefaultReconcileInterval` = 1h). **Off by default** — an upgrade never starts retracting on deploy.
- `reconcileTargets` (unit-tested) selects enabled inboxes with a syncer, defaults the interval, and carries the cap fraction/floor + audit log into the pass options. `reconcile_enabled` without an upstream is logged and skipped.
- `runReconcileLoop` mirrors the sync loop: an initial pass at startup, then on the ticker; errors and a held cap are logged, never fatal.

### Shared audit log
The audit log is now opened once in serve and shared by the message store and the reconcile pass (which audits each retraction per ADR 0026), instead of being private to `openStore`. `openStore` takes the log; serve owns its lifecycle (closed last, after the stores).

### Not in this slice
The operator release/status surface (admin API + CLI) is the next slice; until then a held inbox is visible in the serve log. Telegram push tracked in #149.

### Tests
`reconcileTargets`: only enabled inboxes with a syncer are targets; interval defaults when unset; cap fraction/floor + audit ride in the options; enabled-without-syncer and disabled are excluded.

Part of #140.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -122,19 +122,16 @@ func (c *CLI) router() *policy.Router {
 	return policy.NewRouter(c.ApprovalStrict, c.ApprovalLight)
 }
 
-// openStore opens the audit log and the message store per config and returns
-// the store plus a close func that closes both (store first, then audit).
-func (c *CLI) openStore() (sluice.MessageStore, func(), error) {
-	al, err := audit.New(c.AuditType, c.AuditDB)
-	if err != nil {
-		return nil, nil, err
-	}
+// openStore opens the message store per config, writing best-effort audit
+// entries to al. al is created and closed by the caller (it is shared with the
+// reconcile pass, which audits each retraction — ADR 0026), so the returned close
+// func closes only the store.
+func (c *CLI) openStore(al audit.AuditLog) (sluice.MessageStore, func(), error) {
 	ms, err := sluice.New(c.StoreType, c.SluiceDB, al)
 	if err != nil {
-		_ = al.Close()
 		return nil, nil, err
 	}
-	return ms, func() { _ = ms.Close(); _ = al.Close() }, nil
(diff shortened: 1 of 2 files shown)


yaad-rubigd (approved):
Reviewed (the activation slice) against ADR 0026 — approving.

- reconcileTargets: only reconcile_enabled inboxes WITH an upstream syncer get a loop (off-by-default honored); disabled + no-upstream excluded; interval = config or 1h default; cap fraction/floor + shared audit wired into opts. Unit-tested for all those cases.
- runReconcileLoop: own ticker (interval-independent of the forward poll), an initial pass at startup, ctx-cancellable. The startup pass safely skips until the forward sync establishes the cursor (guard 2 fires when loaded UIDVALIDITY is 0), so reconcile never runs ahead of sync.
- runReconcile: ErrReconcileHeld → logged as held, the loop continues (not a crash); other errors logged best-effort; retraction count logged.
- Audit opened once in serve and shared by the sluice store + the reconcile pass, so retractions are audited. Verified concurrency-safe: Append does read-prev-hash + compute + put in a single bbolt Update txn, and bbolt serializes writers — so the new parallel reconcile + sluice writers cannot fork the hash chain. No mutex needed.
- Operator warning when reconcile is enabled on an inbox with no upstream — good touch.

Activation is off-by-default and the cap is already in ahead of it. Approving.

sora-rubigd (approved):

PR #153, opened 2026-06-30
feat(admin): reconcile status + release operator UX (#140 slice 7)

### LINKED ISSUE #140: Upstream reconciliation: retract local copies of messages removed from the synced mailbox

forge-rubigd:
Slice 7 of #140 (ADR 0026 upstream reconciliation) — the final slice: the operator pull surface for the safety-cap latch.

Admin service: SetReconcileControls wires release/status callbacks over serve's per-inbox syncers; ReconcileStatusAll / ReleaseReconcile, returning an unavailable sentinel when no inbox has an upstream.

Admin API + client: GET /reconcile (each inbox's latch) + POST /reconcile/{inbox}/release (clear the latch + one cap-bypassed pass = the confirmed purge once), mirrored in the client.

CLI: darbaan reconcile status (lists held inboxes + counts) and darbaan reconcile release <inbox> (confirm the large retraction and resume).

Serve wiring installs the controls over its syncers (release audits via the shared log), skipped when no inbox has an upstream — so a latched inbox is always operator-releasable before reconcile runs in prod.

Notification: agreed v1 is pull (reconcile status) + serve-log alert; push tracked as #149.

Tests: unavailable sentinel when unwired; HTTP round-trip for status + release; unwired surfaces as a client error.

Closes #140.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -101,6 +101,7 @@ type CLI struct {
 	Serve      ServeCmd      `cmd:"" help:"Run the SMTP + IMAP faces and the admin API."`
 	Queue      QueueCmd      `cmd:"" help:"Inspect and decide held messages (via the running serve's admin API)."`
 	Holds      HoldsCmd      `cmd:"" help:"Inspect and decide inbound messages held for a human (ADR 0021)."`
+	Reconcile  ReconcileCmd  `cmd:"" help:"Inspect and release inboxes held by the reconcile safety cap (ADR 0026)."`
 	Filter     FilterCmd     `cmd:"" help:"Inspect the inbound filter without starting serve (ADR 0021/0022)."`
 	Telegram   TelegramCmd   `cmd:"" help:"Run the Telegram approval client (a separate admin-API client process)."`
 	DkimPubkey DkimPubkeyCmd `cmd:"" name:"dkim-pubkey" help:"Print the DKIM public-key record to pin to the agent."`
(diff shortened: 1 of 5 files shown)


yaad-rubigd (approved):
Reviewed against ADR 0026 + the amendment — the operator UX is complete and correct, approving.

- GET /reconcile (per-inbox suspended + held-count) + POST /reconcile/{inbox}/release, both auth'd; CLI `reconcile status` / `release <inbox>`; matching client methods; serve sets the controls over the syncers; ErrReconcileUnavailable → 503 when no upstream inbox is configured.
- statusFn iterates the syncers → ReconcileLatch; releaseFn → ReleaseReconcile with the shared audit log.

Non-blocking note (re-raising the slice-5b point): release calls ReleaseReconcile without first checking the inbox is actually latched — so `reconcile release <inbox>` on a NON-held inbox does a cap-bypassed pass (an uncapped purge of whatever is currently gone). Practically near-no-op since the cadence keeps the gone-set small, but it's a footgun and unclear UX — releasing something that isn't held should no-op or error "not held". Recommend a latched-precheck; fine as a fast-follow, not blocking. Approving.

sora-rubigd (approved):

PR #155, opened 2026-06-30
fix(reconcile): release refuses a non-held inbox (#154)

### LINKED ISSUE #154: reconcile release should no-op/error on a non-latched inbox
Fast-follow to #140 / ADR 0026.

`reconcile release <inbox>` (admin `POST /reconcile/{inbox}/release` → ReleaseReconcile) currently clears the latch and runs a cap-bypassed pass WITHOUT first checking the inbox is actually latched. So releasing a NON-held inbox does an uncapped pass — it purges whatever is currently gone with no cap.

Practically near-no-op (the cadence keeps the gone-set small on a healthy inbox), but it's a footgun and unclear UX. Add a precheck: if the inbox is not suspended, no-op or return a clear "inbox not held" error instead of performing a cap-bypassed purge.

Non-urgent; surfaced during the slice-7 review.

forge-rubigd:
## #154 — release-guard before the v0.6.0 cut

`ReleaseReconcile` runs the pass **cap-bypassed**, so releasing an inbox the safety cap never latched would purge the current gone-set with no cap — a footgun once reconcile is enabled on a real inbox.

### Fix
Release now checks the latch first and refuses unless the inbox is actually suspended.
- **imapsync**: a `Suspended` precheck at the top of `ReleaseReconcile` → `ErrReconcileNotHeld` (the real guard — the cap-bypass can't fire on a non-held inbox even off the admin path).
- **admin**: `ErrReconcileNotHeld` → **409 Conflict** on `POST /reconcile/{inbox}/release`; serve translates the syncer's error to the admin sentinel.

### Tests
- imapsync: a non-held inbox with a large removal (6 of 8 gone) is **not** purged by release — `ErrReconcileNotHeld`, store intact (the footgun is closed).
- admin: releasing a non-held inbox surfaces as a client error (409).

Closes #154.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -756,7 +756,11 @@ func (*ServeCmd) Run(cli *CLI) error {
 				if !ok {
 					return 0, fmt.Errorf("inbox %q has no upstream to reconcile", name)
 				}
-				return syn.ReleaseReconcile(ctx, imapsync.ReconcileOptions{Audit: al})
+				n, rerr := syn.ReleaseReconcile(ctx, imapsync.ReconcileOptions{Audit: al})
+				if errors.Is(rerr, imapsync.ErrReconcileNotHeld) {
+					return 0, admin.ErrReconcileNotHeld // → 409 at the API
+				}
+				return n, rerr
 			},
 			func() ([]admin.ReconcileStatus, error) {
 				out := make([]admin.ReconcileStatus, 0, len(syncers))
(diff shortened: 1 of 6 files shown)


yaad-rubigd (approved):
Reviewed — the release-guard is exactly right, approving.

- The Suspended-precheck is at the top of ReleaseReconcile (the imapsync primitive), so a non-held inbox returns ErrReconcileNotHeld BEFORE any clear/bypass-purge — the cap-bypass can't fire off any path, not just the admin one. Defense at the primitive, the right place.
- admin maps it to 409 (clear "nothing to release").
- TestReleaseRejectsNonHeld proves the footgun is closed: a non-latched inbox with 6/8 gone is not purged by release, and the unset latch is left untouched. Plus the 409-over-HTTP test.

Approving — last one before the cut.

sora-rubigd (approved):

PR #157, opened 2026-07-01
refactor: structured logging with log/slog (#151)

### LINKED ISSUE #151: Switch logging to structured logging (log/slog)
darbaan currently logs ad-hoc via `fmt.Fprintf(os.Stderr, ...)` and `log.Printf`. Switch to structured logging using the standard library **`log/slog`** (no external dependency).

Scope:
- Introduce a `*slog.Logger` behind a small seam so it stays injectable/testable.
- Thread it through serve / imapsync / reconcile / listener / admin / signer.
- Replace the `fmt.Fprintf(os.Stderr, ...)` + `log.Printf` calls with leveled, structured calls; attach fields (inbox, upstream_uid, event, ...) as slog attributes instead of formatted strings.
- Config for level + handler (text vs json).
- Keep secrets out of logs (existing discipline).

Stdlib `log/slog`, no third-party dependency. Non-urgent tech-debt; no behavior change. Note: the reconcile cap-latch log becomes a structured "reconcile held" event here — a clean hook for the latch push notification (#149).

forge-rubigd:
## #151 — structured logging with log/slog

Migrates operational logging from stdlib `log.Printf` + `fmt.Fprintf(os.Stderr)` to leveled, structured `slog`. Stdlib only, no third-party dep, **no behavior change** (all existing tests pass unchanged).

### The seam
`main` builds a `*slog.Logger` from two new flags and installs it as the slog default:
- `--log-level` (debug|info|warn|error, default info)
- `--log-format` (text|json, default text)

Every subcommand — and every component that falls back to `slog.Default()` — logs through it.

### Injection
- **imapsync Syncer** carries a per-inbox logger (`logger.With("inbox", name)`), so every sync/reconcile record is inbox-tagged. `SetLogger` makes it injectable/testable.
- **Telegram client** is tagged `component=telegram`, also via `SetLogger`.
- **Leaf/best-effort lines** (sluice + inbound orphan-sweep / best-effort audit-append fail; the IMAP handler's best-effort errors) use the configured `slog.Default()` rather than rippling the store/session constructors + every store test — matching the agreed call for non-per-inbox lines.

### The #149 hook
The reconcile cap-latch log is now a structured **`reconcile held`** event (attrs: `candidates`, `synced`, `cap_fraction`, plus `inbox` from the injected logger) — the clean event the #149 latch-push will consume. Tested: a capturing handler asserts the structured record + attrs.

### Fields to attrs
`inbox`, `id`, `event`, `count`, `err`, etc. Secrets are never logged (the Telegram token stays out; only the bot username + operator id appear).

### Not converted
The two `queue ls` / `holds ls` empty-state messages stay as CLI output (not operational logs).

Closes #151.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -16,7 +16,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"log"
+	"log/slog"
 	"net"
 	"net/http"
 	"os"
@@ -98,6 +98,9 @@ type CLI struct {
 	ApprovalStrict []string `name:"approval-strict" default:"manual" help:"Approver chain for the strict path."`
 	ApprovalLight  []string `name:"approval-light" default:"manual" help:"Approver chain for the light path."`
 
+	LogLevel  string `name:"log-level" default:"info" enum:"debug,info,warn,error" help:"Log verbosity."`
+	LogFormat string `name:"log-format" default:"text" enum:"text,json" help:"Log handler: text (human-readable) or json (structured ingest)."`
+
 	Serve      ServeCmd      `cmd:"" help:"Run the SMTP + IMAP faces and the admin API."`
 	Queue      QueueCmd      `cmd:"" help:"Inspect and decide held messages (via the running serve's admin API)."`
 	Holds      HoldsCmd      `cmd:"" help:"Inspect and decide inbound messages held for a human (ADR 0021)."`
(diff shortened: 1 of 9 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approved. Clean stdlib slog migration: setupLogging builds the handler from the two flags and installs it as default before Run, so constructed components capture it; the Syncer/Telegram loggers are injectable and default to slog.Default(); leaf and best-effort lines correctly stay on the package default rather than rippling logger injection through the store factories. Levels are well-chosen and the per-cycle held line is demoted to Debug so a latched inbox does not flood. The 'reconcile held' structured event carries candidates/synced/cap_fraction with the inbox from the injected logger, and the capturing-handler test locks both the event and injectability. No secrets in any log call. CI green.

PR #158, opened 2026-07-01
feat(admin,telegram): Change-sender at the approval gate (#139)

### LINKED ISSUE #139: Multi-inbox slice 5: Telegram approval-gate Change-sender (deferred)
Deferred final slice of the multi-inbox work (ADR 0023). The inbound + outbound multi-inbox path shipped (slices 1–4); this is the approval-gate sender-picker UI, intentionally held so we can deploy and test the rest first.

## What
The Telegram approval message gains an **Approve** + **Change** control. **Change** lists the configured inbox **identities**; tapping one **approves-and-sends from that identity**, routed through that inbox's backend Sender (the per-inbox Sender map from slice 4b). The default selection is the From the message was submitted/stamped with.

## Why deferred
Outbound already works without it: submission routes the From to an inbox and stamps it, and approve delivers via that inbox's Sender. The Change button is operator convenience for re-picking the sending identity at approval time, not required for correct routing. Holding it lets us deploy + test the inbound/outbound multi-inbox in production for a while before adding more surface.

## Depends on
- Slice 4a (From→inbox routing + sluice Inbox stamp) — merged.
- Slice 4b (per-inbox Senders + send dispatch) — merged.
The identity list is exactly the configured inbox identities.

## Acceptance
- Approve + Change buttons on the held-message gate message.
- Change → list of configured inbox identities → selecting one approves and sends from that identity via its backend.
- N=1 unchanged (single identity → no meaningful change vs plain Approve).

Refs: ADR 0023 (multi-inbox), ADR 0025 (deferred this as the sibling of the long-body work).

forge-rubigd:
## #139 — multi-inbox slice 5: Change-sender at the approval gate (ADR 0023)

The held-message Telegram gate gains **Approve** + **Change**. Change lists the configured inbox identities; selecting one **approves-and-sends from that identity** via that inbox's backend Sender (slice 4b). N=1 (≤1 identity) omits Change — plain Approve. Motivated by a real incident: a reply went from the wrong identity because the gate offered no switch at approval time.

### admin
- `ApproveAs(id, inbox)`: approve, then send **from that inbox's identity** — the envelope MAIL FROM **and** the `From:` header are rewritten to the identity, and the send routes through that inbox's Sender.
- **Body preserved byte-for-byte**: the From rewrite is surgical (textproto — replace the header, re-append the original body untouched), so MIME/attachment bodies (ADR 0025) are unchanged, no re-encode. No new dependency (go-message already present).
- **Persist-original**: the stored record keeps the message as submitted (send-time rewrite only). The chosen identity is recorded in a structured `approved-as` slog event (message_id, inbox, identity) — now that #157 landed.
- **No wrong-identity re-send path**: `decide()` rejects any non-Pending message and `Approve` is one-shot, so a re-send must re-enter the gate (another Change opportunity), never silently auto-send the original.
- `GET /inboxes` lists the identities; `POST /queue/{id}/approve-as/{inbox}`; unknown inbox → `ErrUnknownInbox` (400).

### telegram
- `[Change sender]` on the gate keyboard (only when there's more than one identity) opens the identity picker; picking one calls approve-as, **Back** returns to Approve. Identities are fetched from serve at startup — serve is the single source of truth, no config drift with the telegram process.

### Tests
rewriteFrom (+ MIME body preserved byte-for-byte); ApproveAs sends via the chosen sender with envelope+header From rewritten and the stored record unchanged; unknown-inbox error; Inboxes list; http round-trip (client→server); parseApproveAs (incl. inbox names containing `:`); the identity picker + N=1-omits-Change keyboard.

Closes #139.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -749,6 +749,16 @@ func (*ServeCmd) Run(cli *CLI) error {
 	}
 	svc.SetSenders(senders)
 
+	// The configured inbox send identities feed the Change-sender picker + the
+	// ApproveAs From-rewrite (ADR 0023 slice 5). Only inboxes with an identity.
+	identities := make(map[string]string, len(inboxes))
+	for _, in := range inboxes {
+		if in.Identity != "" {
+			identities[in.Name] = in.Identity
+		}
+	}
+	svc.SetInboxIdentities(identities)
+
 	// One local service hosts both agent faces (ADR 0001): SMTP submission
 	// (outbound trap) and IMAP read (the agent reads its bounces). The submission
 	// face routes each From to an inbox (ADR 0023): matched Identity → that inbox,
--- a/internal/admin/change_sender.go
+++ b/internal/admin/change_sender.go
@@ -1,0 +1,73 @@
+package admin
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"io"
+	"sort"
+
+	"github.com/emersion/go-message/textproto"
+
+	"github.com/yaad-index/darbaan/internal/inbound"
+	"github.com/yaad-index/darbaan/internal/sluice"
+)
+
+// ErrUnknownInbox is returned by ApproveAs when the requested inbox is not a
+// configured inbox with a send identity.
+var ErrUnknownInbox = errors.New("admin: unknown inbox (no configured send identity)")
+
+// InboxIdentity pairs a configured inbox name with its send identity (ADR 0023).
+type InboxIdentity struct {
+	Name     string `json:"name"`
+	Identity string `json:"identity"`
+}
+
+// SetInboxIdentities installs the per-inbox send identities (ADR 0023 slice 5):
+// the identities the operator may re-pick at approval time, and the address
+// ApproveAs rewrites From to. serve builds it from the configured inboxes; only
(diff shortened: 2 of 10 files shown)


yaad-rubigd (approved):
Approved. Reviewed the full diff. The From-rewrite is header-only via textproto with the body re-appended byte-for-byte, so MIME/attachment structure is preserved (ADR 0025), and there is a test asserting a From: line embedded in the body is untouched. approve() unifies the two paths cleanly: empty target leaves ApproveID behavior unchanged; a chosen target rewrites both the envelope MAIL FROM and the header From, routes via that inbox's Sender, and leaves the persisted record as submitted (send-time only) with the chosen identity captured in a structured approved-as event. Callback prefixes are non-overlapping (approve_as is not caught by approve, chg_back not by change), parseApproveAs uses the last colon field so an inbox name containing a colon still parses, and identities are fetched once before the poll loop starts so the lock-free read is safe with a best-effort fallback that just omits the picker. Unknown target returns a 400. Tests cover rewrite + body-preservation, send-via-chosen-sender, record-unchanged, unknown-inbox, list, http round-trip, parse, and both keyboards. CI green.

sora-rubigd (approved):

PR #161, opened 2026-07-01
docs(adr): ADR 0027 — multi-agent tenancy

forge-rubigd:
Drafts ADR 0027, the design record for darbaan multi-agent tenancy. Supersedes the deferred multi-agent sketch in ADR 0010 and builds on the multi-inbox unit from ADR 0023.

## What it decides
- **Config**: an `agents:` list beside `inboxes:`. Each agent = a login name, a password from a secret env var (`DARBAAN_AGENT_<NAME>_PASSWORD`, never in config per 0012), a `grants:` list of `{inbox, access:[read,send]}`, and a `default_inbox`. Sharing = list an inbox under multiple agents; privacy = omission.
- **Auth = identity**: username-to-agent resolution, constant-time password compare (generalizes today's single-cred check to a map); the resolved agent's grants become the session principal.
- **Read scoping (IMAP)**: `LIST` shows only granted-read inboxes; `SELECT`/`STATUS` on an ungranted inbox returns no-such-mailbox.
- **Send scoping (SMTP)**: a submit whose From resolves to an inbox the agent lacks `send` on is rejected at submit, fail-closed (0003). SMTP AUTH is already required today, so the authenticated principal is known at MAIL FROM.
- **Principal / mail-owner decoupling** (the one real refactor): the store owner key becomes a stable per-inbox mail-owner (the inbox `name`); the connecting agent only gates access.
- **Bounce scoping**: Darbaan-generated bounces stay keyed to the originating agent (`UpstreamUID == 0`), never folded into the shared inbox mail-owner — a read-only co-agent never sees another agent's bounce. The read face unions two owner key-spaces: synced upstream mail (inbox-mail-owner, grant-gated) + own bounces (originating-agent id).
- **Per-agent `INBOX`**: each agent's `default_inbox` shows as `INBOX`; the send catch-all is per-agent.
- **Backward compat**: no `agents:` means a single implicit agent, read+send on all inboxes, existing owner value preserved; opt-in like `inboxes:` in 0023.
- **Audit**: rows gain the acting agent id (implicit agent id = the configured agent username).
- **Migration**: the sole data-touching case (an existing multi-inbox store adopting `agents:`) runs a one-time owner rekey over synced records only (`UpstreamUID != 0`); bounces untouched.

Non-goals: no roles/RBAC, no per-message ACLs; bounce-signing key and the operator approval gate both stay global.

## Review folded in
- Env-var normalization collision: validator rejects colliding normalized names.
- `default_inbox` must have `read` (else un-`SELECT`able); mandatory, inferred when exactly one read grant, else config error.
- Migration promoted from Open Decisions to its own section.
- Implicit-agent audit id pinned to the configured agent username.
- SMTP AUTH-required stated affirmatively (already the case).
- Shared-inbox flag/keyword state documented as an intended consequence.
- Bounce scoping pinned (load-bearing for shared-inbox privacy): two separate key-spaces; the rekey skips bounces.

## Sequencing
ADR-first PR (design, hard-gate) ahead of the four implementation slices; slice 4 finalizes this record (status to Accepted, README).

Hard-gate: touches `adr/*`, so it needs the operator's explicit approve on HEAD in addition to review.

### DIFF
--- a/adr/0027-multi-agent-tenancy.md
+++ b/adr/0027-multi-agent-tenancy.md
@@ -1,0 +1,316 @@
+# ADR 0027: Multi-agent tenancy — per-agent logins, grants, and per-principal mailbox naming
+
+**Status:** Proposed (2026-07-01)
+
+## Context
+
+ADR 0010 chose **multi-mailbox + single-agent** for v1 and deferred multi-agent,
+sketching it so v1 would not foreclose it. ADR 0023 then realized multi-inbox: N
+inboxes in one Darbaan, each with its own backend, identity, filters, and
+visibility — but still **one agent principal that sees every inbox** read+send.
+Both ADRs left an inbox id threaded through storage and the audit log
+specifically so a per-agent permission model could be keyed on it later without
+reshaping data.
+
+This ADR is that model. It **supersedes the deferred multi-agent sketch in ADR
+0010** and builds directly on ADR 0023's inbox as the first-class unit. The
+motivating shape: several agent principals share one Darbaan, each seeing **only**
+the inboxes it is granted, in the direction(s) it is granted — one agent may read
+and send a shared team inbox while another only reads it, and a third sees a
+private inbox neither of the first two can address. Privacy is by **omission**: an
+inbox an agent is not granted does not exist for that agent.
+
+## Decision
+
+**An agent becomes a first-class, named config principal with its own login
+credentials and an explicit per-inbox grant list. Authentication resolves the
+login to an agent; that agent's grants are the session principal and gate every
+read and send. The mail-owner store key is decoupled from the login principal so
+several agents can share one inbox.**
+
+### Config shape
+
+A new `agents:` list sits beside `inboxes:` (ADR 0023). Each agent is a login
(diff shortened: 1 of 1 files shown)


yaad-rubigd (approved):
Approved (re-review of the revised HEAD). All blocking flags resolved cleanly: normalized env-var collision is rejected by the config validator; default_inbox requires read and the mandatory/inferred rule (one read grant infers, two require an explicit choice) resolves to exactly one default; the owner-key rekey is promoted to its own Migration section and correctly scoped to synced records (UpstreamUID != 0) only; the implicit agent id is the configured username; SMTP AUTH is stated as required for the send path; and the shared-IMAP-state consequence is documented. The bounce-scoping is exactly right: two distinct owner key-spaces, synced mail by inbox-mail-owner (grant-gated, shared among co-readers) and locally-generated bounces (UpstreamUID == 0) by originating-agent id (private, never visible to a co-reader), with the rekey deliberately leaving bounce records alone. Design is sound and least-privilege throughout. Good to build the slices.

sora-rubigd (approved):

PR #162, opened 2026-07-01
feat(agents): multi-agent config schema + per-agent auth (ADR 0027 slice 1)

forge-rubigd:
Slice 1 of 4 implementing ADR 0027 (multi-agent tenancy). Schema + authentication + backward-compat only; grant enforcement lands in later slices.

## What this slice does
- **New `agentcfg` package**: parses the `agents:` list beside `inboxes:`. Each agent = a login name, a `grants:` list of `{inbox, access:[read,send]}`, and a `default_inbox`. Passwords are never in config — read from `DARBAAN_AGENT_<NAME>_PASSWORD` (ADR 0012).
- **Validation** (mirrors the `inboxcfg` pattern):
  - unique agent names;
  - unique password-env names — the many-to-one mangle (`a-b` and `a_b` both fold to `..._A_B_...`) must not silently share a secret, so a collision is rejected at load;
  - well-formed grants: non-empty, each inbox configured, granted at most once, access ⊆ {read, send};
  - exactly one resolvable `default_inbox` **with read access** — explicit field, or a `default: true` grant, or inferred from a lone read grant; ambiguous (≥2 read grants, no default) is rejected, and a send-only default is rejected (an un-`SELECT`able INBOX).
- **Per-agent authentication**: a new `Auth`/`Principal` type in the listener replaces the single `Credential`. IMAP `LOGIN` and SMTP `AUTH` resolve the presented username to an agent and compare its password constant-time; an unknown username still runs a fixed comparison, so auth fails identically whether the username or password was wrong (no enumeration oracle). The authenticated agent name is stamped on the submission (`Submission.Agent`).
- **Backward compat**: with no `agents:` list, a single implicit agent (`DARBAAN_AGENT_PASS`) is granted every inbox read+send — the current single-agent path, unchanged.

## What this slice deliberately does NOT do
- No grant **enforcement** yet: `LIST`/`SELECT` are not grant-gated and send is not scoped (slices 2/3).
- No principal / mail-owner decoupling and no per-agent `INBOX` (slice 2).
- No audit acting-agent id (slice 4).

A multi-agent config authenticates per-agent but is not yet isolated; the feature is not deployed until all four slices land + a release is cut, so the intermediate state never ships.

## Tests
- `agentcfg`: parse, all validation error paths (dup name, env collision, grant errors), and every `default_inbox` resolution + error case.
- `listener`: per-agent `Verify` (own creds only), wrong-password, unknown-username, empty-auth; the existing SMTP acceptance test now asserts the authenticated agent name is stamped on the queued message.

Cross-package touch: `cmd/darbaan/main.go` wiring (build principals from config / implicit agent, pass `Auth` to both faces) and the two `listener` test files updated for the `Credential`→`Auth` signature change.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/alecthomas/kong"
 
 	"github.com/yaad-index/darbaan/internal/admin"
+	"github.com/yaad-index/darbaan/internal/agentcfg"
 	"github.com/yaad-index/darbaan/internal/audit"
 	"github.com/yaad-index/darbaan/internal/backend"
 	"github.com/yaad-index/darbaan/internal/bounceguard"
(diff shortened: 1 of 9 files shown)


yaad-rubigd (approved):
Approved. Slice 1 is clean and correctly scoped to schema + auth only, no premature enforcement. agentcfg validation is thorough: unique names, env-collision rejection via the shared PasswordEnv mangle (single source of truth for runtime lookup and the load-time guard, so they can't diverge), well-formed grants, and DefaultInbox resolution covering explicit/marked/inferred with the requires-read and ambiguity rules we asked for. auth.go does constant-time compare with a dummy-password path for unknown usernames (no enumeration oracle) and SingleAuth preserves the back-compat single-agent path; resolvePrincipals falls back to the implicit DARBAAN_AGENT_PASS agent when agents: is absent, unchanged. Login sites wire cleanly to Verify. Tests cover parse, no-agents, env collision, validation failure modes, default resolution, and cross-agent auth isolation. Good foundation for slices 2-4.

sora-rubigd (approved):

PR #164, opened 2026-07-01
feat(agents): IMAP read-scoping + per-agent INBOX (ADR 0027 slice 2a)

forge-rubigd:
Slice 2a of 4 (slice 2 split into 2a/2b per review). Read-face grant enforcement + per-agent mailbox naming. **No stored data is touched** — that (owner decoupling + one-time rekey + bounce two-key-space union) is slice 2b.

## What this slice does
- **Grants → session**: `Principal` now carries `DefaultInbox` + read/send grant sets; `Auth.Verify` returns the matched principal so the IMAP session knows who is connected.
- **IMAP read-scoping**:
  - `LIST` returns only the inboxes the connecting agent has read on;
  - `SELECT`/`STATUS` on an inbox the agent lacks read on returns no-such-mailbox — indistinguishable from a name that doesn't exist (privacy by omission);
  - the agent's `default_inbox` is exposed as `INBOX` (per-principal naming); every other granted inbox by its own name; the `default_inbox` is reachable **only** as `INBOX`, never by its own name.
- **Back-compat**: an unrestricted principal (nil grants — the implicit single agent, and the listener tests) reads every inbox and maps the global default to `INBOX`, exactly as today.

## What this slice deliberately does NOT do
- No principal / mail-owner decoupling, no owner rekey, no bounce two-key-space union (slice 2b — the data-touching, focused-review half).
- No SMTP send-scoping (slice 3).

Because there's no owner decoupling yet, a multi-agent config's reads still key on the agent name, so shared-inbox mail isn't returned until 2b — but the gating logic (LIST/SELECT/STATUS, INBOX naming) is complete and tested here, and implicit-agent deployments are fully unaffected. The feature isn't deployed until all slices land + a release is cut.

## Tests
- `TestIMAPReadScoping`: an agent granted read on work+personal (not the default inbox), default_inbox=work — asserts LIST shows exactly {INBOX, personal}, INBOX resolves to work, `personal` reachable by name, and `default`/`work`/STATUS-on-`default` all no-such-mailbox.
- `Auth.Verify` tests updated for the principal return; the existing multi-inbox + submission tests still pass unchanged (back-compat).

Cross-package touch: `cmd/darbaan/main.go` builds each agent's grant sets + resolved default_inbox into the `Principal` (implicit path unchanged).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -257,7 +257,24 @@ func (c *CLI) resolvePrincipals(inboxes []inboxcfg.Inbox) ([]listener.Principal,
 		if pass == "" {
 			return nil, fmt.Errorf("agent %q: set %s (the password is supplied out-of-band, never in config — ADR 0012)", a.Name, env)
 		}
-		principals = append(principals, listener.Principal{Name: a.Name, Password: pass})
+		def, err := agentcfg.DefaultInbox(a)
+		if err != nil {
+			return nil, err
+		}
+		reads := make(map[string]bool, len(a.Grants))
+		sends := make(map[string]bool, len(a.Grants))
+		for _, g := range a.Grants {
+			if g.HasAccess(agentcfg.AccessRead) {
+				reads[g.Inbox] = true
+			}
+			if g.HasAccess(agentcfg.AccessSend) {
+				sends[g.Inbox] = true
+			}
+		}
+		principals = append(principals, listener.Principal{
+			Name: a.Name, Password: pass,
+			DefaultInbox: def, Reads: reads, Sends: sends,
+		})
 	}
 	return principals, nil
 }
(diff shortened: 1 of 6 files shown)


yaad-rubigd (approved):
Approved. Slice 2a is clean read-scoping + per-agent INBOX with no data touch, exactly as scoped. Principal now carries DefaultInbox/Reads/Sends with nil = unrestricted (back-compat single-agent unchanged). resolveMailbox/mailboxNames gate every LIST/SELECT/STATUS by canRead, so an ungranted inbox is indistinguishable from non-existent (privacy by omission); fetch/keyword are transitively gated since you can't SELECT what you can't resolve. Per-session mailboxName maps the connecting agent's default_inbox to INBOX and exposes it only as INBOX (not by its own name), correct. SMTP Auth captures the principal name; send-enforcement rightly deferred to slice 3. Test is thorough and asserts the ungranted-is-no-such-mailbox and default_inbox-only-as-INBOX behaviors. Good.

sora-rubigd (approved):

PR #165, opened 2026-07-01
feat(agents): principal/mail-owner decoupling + bounce two-key-space (ADR 0027 slice 2b)

forge-rubigd:
Slice 2b of 4 — the data-touching half of read scoping. Builds on 2a (#164, merged). Owner decoupling + one-time rekey + the bounce two-key-space union.

## Owner decoupling
In multi-agent mode an inbox's **synced upstream mail is keyed by the inbox name** (its mail-owner), not a login — so several agents sharing an inbox read the same records. The single-agent / back-compat path is **unchanged** (mail stays keyed by the implicit agent; no rekey).

Wired through a `mailOwner(inbox)` resolver: the syncer owner, the IMAP read face, and the admin held queue all key synced mail by it.

## Migration (one-time rekey)
`InboundStore.RekeyOwnersToInbox()` rewrites each **synced** record's owner to `NormInbox(inbox)`, **only** for `UpstreamUID != 0`. Locally-generated records (bounces, `UpstreamUID == 0`) keep their originating-agent owner. The synced **dedup index moves with each record**, so a later re-sync still dedups instead of duplicating. Idempotent (a record already inbox-owned is skipped), so it runs safely at every multi-agent startup; the single-agent path never calls it.

## Bounce two-key-space (the privacy contract)
A bounce is keyed to its **originating agent** (Owner) AND the **inbox the failed submission was for** (bounce now carries `Inbox`, threaded from the submission). The read face unions, per selected inbox X:
- **synced mail** — `List(mailOwner(X), X)`, shared by every reader of X;
- **the connecting agent's own bounces** — `List(agent, X)`, private to that agent.

So a read-only co-agent on a shared inbox **never** sees another agent's bounce. When the two owners coincide (single-agent) one List already holds both. Per-message store operations (SetSeen, keyword writes, content fetch) key on **each message's own `m.Owner`**, since a selected inbox now mixes the two owner spaces.

## Guard
An agent name may not equal an inbox name — they share the store's owner namespace once mail is keyed by inbox, so an overlap could leak an agent's bounces to co-readers. Rejected at config load.

## Tests
- `TestRekeyOwnersToInbox`: synced rekeyed to inbox owner, bounce left at its agent owner, dedup index moved (re-sync dedups), idempotent second run.
- `TestGenerateKeysOwnerAndInbox`: a bounce carries both owner and submission-inbox.
- `TestIMAPBouncePrivacyOnSharedInbox`: two agents share inbox "work" — each sees the shared synced message + only its own bounce, never the other's.
- `TestValidateAgentInboxNameCollision`: agent-name == inbox-name rejected.
- All existing tests pass unchanged (back-compat: single-owner path, no rekey).

## Cross-package touch
- `internal/inbound` (store): new `RekeyOwnersToInbox` on the interface + bbolt.
- `internal/bounce`: `Bounce.Inbox` threaded from the submission.
- `internal/admin`: held queue keys by the per-inbox mail-owner (`SetInboundHolds` takes a resolver).
- `cmd/darbaan/main.go`: builds the `mailOwner` resolver, runs the startup rekey in multi-agent mode, threads it to sync / read / held-queue.

## Not in this slice
SMTP send-scoping (slice 3), audit acting-agent id + docs + ADR finalize (slice 4).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -221,45 +221,62 @@ func (c *CLI) resolveInboxes() ([]inboxcfg.Inbox, error) {
 	return inboxes, nil
 }
 
+// principalWiring is the resolved multi-agent auth for the serve loop (ADR 0027):
+// the login principals, the synced-mail owner key per inbox, and whether the
+// mail-owner is decoupled from the login (multi-agent mode).
+type principalWiring struct {
+	principals []listener.Principal
+	// mailOwner is an inbox's synced-mail store owner: its own name in multi-agent
+	// mode (so agents sharing an inbox read the same records), or the single
+	// implicit agent in back-compat mode (unchanged store keys).
+	mailOwner func(inbox string) string
+	// decoupled is true in multi-agent mode, where synced mail is keyed by inbox
+	// name and existing records are rekeyed once at startup.
+	decoupled bool
+}
+
 // resolvePrincipals resolves the configured agent principals (ADR 0027): the
 // `agents:` list, each a login with per-inbox grants and a password read from
 // DARBAAN_AGENT_<NAME>_PASSWORD (ADR 0012). With no `agents:` list it returns a
 // single implicit agent authenticated by DARBAAN_AGENT_PASS, granted every inbox
-// read+send — the back-compat single-agent path, unchanged. Grants are validated
-// here; their enforcement (read/send scoping) lands in a later increment.
-func (c *CLI) resolvePrincipals(inboxes []inboxcfg.Inbox) ([]listener.Principal, error) {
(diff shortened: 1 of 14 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approved (slice 2b). Full review complete: idempotent synced-only owner rekey (bounces keep originating owner, dedup index moves), bounce two-key-space with the co-reader-never-sees-another-agent's-bounce invariant proven by TestIMAPBouncePrivacyOnSharedInbox, per-message ops key on m.Owner, agent==inbox-name collision guard, back-compat clean. (Re-posting; my earlier approve did not register.)

PR #166, opened 2026-07-01
feat(agents): SMTP send-scoping + per-agent send catch-all (ADR 0027 slice 3)

forge-rubigd:
Slice 3 of 4 implementing ADR 0027. SMTP send authorization + the per-agent send catch-all. Builds on 2b (#165, merged).

## Send-scoping
An authenticated submission's From is routed to an inbox (identity match, else the catch-all), and the connecting agent must be granted **send** on that inbox — otherwise the submission is **rejected at MAIL FROM**, fail-closed (ADR 0003), before the message is ever trapped in the sluice. A read-only grant cannot originate mail even for an inbox the agent can read.

An unrestricted principal (the implicit single agent / back-compat) may send as any inbox — unchanged.

## Per-agent send catch-all
A From matching no configured identity now routes to the **connecting agent's `default_inbox`** rather than a single global default (ADR 0027). The `Router` takes the catch-all target, chosen per-agent by the submission face; at N=1 it stays the global default inbox, so single-agent behaviour is unchanged. (This completes the per-agent-mailbox-naming addition; the read-side landed in 2a.)

The authenticated principal is threaded onto the SMTP session (mirroring the IMAP face in 2a) so both the send grant and the per-agent default are available at MAIL FROM.

## Tests
- `TestSubmitSendScoping`: an agent with send only on `work` — MAIL FROM as `work` accepted, as read-only `personal` rejected, and an unmatched From routed to its default (`work`) and accepted, stamped with that inbox.
- Existing submission/routing tests pass unchanged (back-compat).

## Cross-package touch
`cmd/darbaan/main.go`: the route closure takes the per-agent catch-all target (still `inboxcfg.Route`, now with the caller-supplied default).

## Remaining
Slice 4: audit acting-agent id + docs + ADR finalize (status → Accepted, README), then release + deploy.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -857,8 +857,10 @@ func (*ServeCmd) Run(cli *CLI) error {
 	// (outbound trap) and IMAP read (the agent reads its bounces). The submission
 	// face routes each From to an inbox (ADR 0023): matched Identity → that inbox,
 	// else the default catch-all, else refused at MAIL FROM.
-	route := func(from string) (string, bool) {
-		return inboxcfg.Route(inboxes, from, inbound.DefaultInbox)
+	// The catch-all target is chosen per-agent by the submission face (ADR 0027):
+	// the connecting agent's default_inbox, or the global default at N=1.
+	route := func(from, catchAll string) (string, bool) {
+		return inboxcfg.Route(inboxes, from, catchAll)
 	}
 	smtpSrv, err := listener.NewServer(listener.ServerConfig{
 		Addr:          cli.ListenerAddr,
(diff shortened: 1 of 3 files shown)


yaad-rubigd (approved):
Approved. Send-scoping is correct and fail-closed: Mail() resolves From to an inbox (per-agent catch-all = the connecting agent's default_inbox, global default at N=1), then rejects at MAIL FROM with 550 5.7.1 if the principal lacks send on the resolved inbox, before the message is ever trapped (ADR 0003). Read-only grant cannot originate mail. Router signature cleanly threads the catchAll target. Back-compat safe: nil Sends = unrestricted, empty DefaultInbox falls back to the global default catch-all, so the single-agent path is unchanged. Principal threaded onto the SMTP session at Auth. Test covers sendable-accept, read-only-reject, and unmatched-From routing to the agent's default with the enqueued Inbox stamped correctly. Good.

sora-rubigd (approved):

PR #174, opened 2026-07-04
feat(sync): STATUS triggers debounced on-demand inbound sync (ADR 0028)

forge-rubigd:
## What

Adds **on-demand inbound sync**: an IMAP `STATUS` for an opted-in inbox now runs an immediate, per-inbox **debounced** upstream pull *before* returning counts, so an agent can "sync now" — see its reply on the same round-trip — instead of waiting out the background poll interval (ADR 0019). `NOOP`/`CHECK` stay true no-ops; no keep-alive triggers a dial.

Motivation: after a send/approval or a test round-trip, new mail isn't visible on the IMAP face until the next scheduled poll. `STATUS` ("give me current status") is the honest, low-frequency place to refresh first and then report.

## Design (ADR 0028)

- **Trigger = `STATUS`, not `NOOP`.** `Status()` already re-lists the store fresh per call, so the only change is to run the pull up front; the existing MESSAGES/UNSEEN/UIDNEXT computation then reports the grown set. Works without a prior `SELECT`.
- **Mandatory per-inbox debounce.** After a pull completes, further triggers for that inbox are suppressed until a configurable window elapses (**silent-skip**: no dial, current counts returned). The debounce is **shared across sessions** (N agents sharing an inbox collapse to one dial) plus a **single-flight** guard; the window is measured from pull *completion*.
- **Best-effort.** A pull error is logged, never surfaced — `STATUS` still reports the store's current counts.
- **Counts-only (v1 boundary).** No unsolicited `EXISTS`/`EXPUNGE` into a selected session; a growth or a reconcile-retraction shrink surfaces via the count and the client's next `SELECT`, never as a mid-session sequence desync.
- **Opt-in per inbox**, default OFF, matching the `reconcile_enabled` idiom: `sync_on_status` + `sync_on_status_interval` (default **60s** window). Inert until an operator opts an inbox in; ignored for a bounce-only / sync-disabled inbox.

## Package touches

- `adr/0028-…` — the decision record (+ adr index). **hard-gate.**
- `internal/imapsync/ondemand.go` — `OnDemandSync` (per-inbox debounce + single-flight).
- `internal/inboxcfg` — `sync_on_status` / `sync_on_status_interval` schema + validation.
- `internal/listener/imap.go` — `Status()` triggers the pull; `SyncTrigger` type plumbed through `NewIMAPServer`.
- `cmd/darbaan/main.go` — builds the coordinator from opted-in inboxes + the runtime default; wires the trigger onto the read face.

## Testing

- `OnDemandSync`: pulls on first trigger, silent-skips within the window, pulls again after it; no-op for an unregistered inbox; no-window = every trigger pulls.
- `Status()`: triggers the pull (gets the resolved inbox name, not `INBOX`) and reflects the pulled mail in the count; survives a pull error; `NOOP` triggers nothing.
- inboxcfg: parse defaults-OFF, duration parsing, enabled-only validation.
- Local `go build` / `go vet` / `gofmt -l` / `go test -race ./...` all green.

## Gate

Touches `adr/*` → hard-gate: the maintainer's APPROVE on the current HEAD, on top of the standard 2-of-N reviews and green CI. The STATUS trigger, the 60s default window, and silent-skip-within-window were all settled before implementation.

### DIFF
--- a/internal/imapsync/ondemand.go
+++ b/internal/imapsync/ondemand.go
@@ -1,0 +1,80 @@
+package imapsync
+
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+// OnDemandSync coalesces caller-triggered ("sync now") upstream pulls per inbox
+// (ADR 0028). An IMAP STATUS for an enabled inbox calls Trigger; Trigger runs one
+// incremental Syncer.Sync, but at most once per the inbox's debounce window and
+// never concurrently with itself — so a burst of STATUS commands, or several
+// agents sharing one inbox (ADR 0027), collapse to a single upstream dial.
+//
+// The debounce is per inbox and shared across every session (it lives here, not
+// on a connection): N connected agents share one window rather than each getting
+// their own and multiplying upstream load by N. Only inboxes an operator opted in
+// (with an upstream syncer) are registered; Trigger is a cheap no-op for any
+// other inbox.
+type OnDemandSync struct {
+	mu      sync.Mutex
+	inboxes map[string]*onDemandInbox
+}
+
+// onDemandInbox is one registered inbox's syncer, debounce window, and the
+// coalescing state guarded by OnDemandSync.mu.
+type onDemandInbox struct {
+	syncer      *Syncer
+	minInterval time.Duration
+	last        time.Time // completion time of the last on-demand pull
+	inflight    bool      // a pull is running now (single-flight guard)
+}
+
+// NewOnDemandSync returns an empty coordinator; register inboxes with Register.
+func NewOnDemandSync() *OnDemandSync {
+	return &OnDemandSync{inboxes: map[string]*onDemandInbox{}}
+}
+
+// Register enables on-demand sync for an inbox with its syncer and debounce
+// window. It is called once per opted-in inbox at wiring time (not concurrently),
+// before any Trigger. A non-positive minInterval is treated as "no debounce"
+// (every Trigger that is not already in flight pulls); callers pass a sane
+// default instead.
+func (o *OnDemandSync) Register(inbox string, syncer *Syncer, minInterval time.Duration) {
+	o.inboxes[inbox] = &onDemandInbox{syncer: syncer, minInterval: minInterval}
+}
+
+// Trigger runs one incremental upstream pull for inbox if one is due — the
+// debounce window since the last on-demand pull has elapsed and no pull is
+// already in flight — and returns the count stored and whether a pull actually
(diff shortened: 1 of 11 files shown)


yaad-rubigd (approved):
Reviewed against the agreed design (ADR 0028). Approving.

**Architecture:** STATUS as the trigger (not NOOP/CHECK) is the right call — the ADR's reasoning (NOOP is a keep-alive clients spam every few seconds; STATUS is explicit + low-frequency) is sound, and NOOP/CHECK/Poll stay true no-ops. No library fork, periodic pull untouched, works without SELECT.

**Concurrency (ondemand.go):** per-inbox debounce shared across sessions (not per-connection, so multiple agents sharing an inbox don't multiply upstream load), single-flight guard, window stamped from pull COMPLETION (a slow pull doesn't immediately re-trigger), and the network dial runs without holding the lock. Clean.

**Behavior matches spec:** opt-in per inbox (sync_on_status default false), 60s default debounce (DefaultSyncOnStatusInterval), silent-skip within window, best-effort (pull error logged, never surfaced — STATUS still returns store counts), no-op for non-opted-in / no-syncer inboxes. Counts-only v1 (no unsolicited EXISTS) is a clean boundary.

Tests cover the debounce/coalesce/no-syncer paths. LGTM. Still needs the second requested review (2-of-N) plus the maintainer APPROVE on HEAD (adr/* hard-gate).

internal/imapsync/ondemand.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Register enables on-demand sync for an inbox with its syncer and debounce
+// window. It is called once per opted-in inbox at wiring time (not concurrently),
+// before any Trigger. A non-positive minInterval is treated as "no debounce"
+// (every Trigger that is not already in flight pulls); callers pass a sane
+// default instead.
+func (o *OnDemandSync) Register(inbox string, syncer *Syncer, minInterval time.Duration) {
+	o.inboxes[inbox] = &onDemandInbox{syncer: syncer, minInterval: minInterval}
+}
```
sora-rubigd, line 46: On error, `in.last = time.Now()` still starts the debounce window — intentional anti-storm behaviour (a flaky upstream won't be hammered), but worth knowing: a transient error means a 60s wait before the next on-demand attempt. Background poll is the fallback so this is fine; just not stated explicitly in the ADR.

sora-rubigd (approved):

fzerorubigd (approved):

PR #177, opened 2026-07-04
feat(sync): log successful on-demand inbound sync pulls (ADR 0028)

forge-rubigd:
Follow-up (b) from the v0.9.0 sync_on_status verification — closes an observability gap.

`OnDemandSync.Trigger` now emits an Info log **`on-demand inbound sync ran`** (with `inbox` + `pulled` count) when a pull actually fires, and only then — a debounced silent-skip logs nothing.

### Why

The on-demand **success** path was silent: `Syncer.Sync` logs the pull count only through the background loop's `runSync` wrapper, which the on-demand trigger bypasses; only a *failed* pull logged (the STATUS handler warns). So from prod logs you couldn't answer "did the on-demand path fire, or did the 5m background poll just run early?" — this line makes an on-demand pull unambiguous and distinct from the background `inbound sync pulled messages`.

Placed in `Trigger` after the `Sync` call (per review suggestion), logged only on the fire path (`err == nil`), so it never fires on the debounced skip.

### Testing

- New `TestOnDemandSyncLogsOnFire`: a fired pull logs `on-demand inbound sync ran` with `inbox=…` + `pulled=…`; a debounced skip logs nothing.
- Local `go build` / `go vet` / `gofmt -l` / `go test -race ./...` green.

No `adr/*` touch — standard 2-of-N.

### DIFF
--- a/internal/imapsync/ondemand.go
+++ b/internal/imapsync/ondemand.go
@@ -2,6 +2,7 @@ package imapsync
 
 import (
 	"context"
+	"log/slog"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Clean. The log fires only on the successful on-demand pull path (`if err == nil`), not on silent-skip or failure — exactly the distinction we needed, so an on-demand pull is now unambiguous versus the background poll's line and versus a skip that logs nothing. Placement after the Sync in OnDemandSync.Trigger is right, and the included TestOnDemandSyncLogsOnFire covers the fire-vs-skip behavior. LGTM. Needs the second requested review.

PR #178, opened 2026-07-04
test(sync): end-to-end on-demand inbound sync via STATUS (ADR 0028)

forge-rubigd:
Follow-up (a) from the v0.9.0 sync_on_status verification — hardens the suite.

A full-stack integration test wiring the **real** path with no stubs:

`client STATUS → listener.Status → imapsync.OnDemandSync (real debounce) → imapsync.Syncer → in-process upstream`

Asserts, in one test, the exact behavior the production wiring runs:
1. First `STATUS` pulls the one upstream message on demand (store empty, no background poll ran).
2. A 2nd upstream message + a `STATUS` **inside** the debounce window silent-skips → count unchanged.
3. A `STATUS` **past** the window pulls again → count reflects the 2nd message.

The existing unit tests cover the pieces (`OnDemandSync` debounce against a real Syncer; `Status()` triggering the callback); this stitches them into a single end-to-end proof.

Test-only; no `adr/*` touch — standard 2-of-N. Local `go test -race ./...` green.

### DIFF
--- a/internal/listener/imap_ondemand_e2e_test.go
+++ b/internal/listener/imap_ondemand_e2e_test.go
@@ -1,0 +1,127 @@
+package listener_test
+
+import (
+	"bytes"
+	"context"
+	"net"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/emersion/go-imap/v2"
+	"github.com/emersion/go-imap/v2/imapclient"
+	"github.com/emersion/go-imap/v2/imapserver"
+	"github.com/emersion/go-imap/v2/imapserver/imapmemserver"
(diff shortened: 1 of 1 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Real-component e2e — no stubs, wires the actual listener.Status → OnDemandSync (real debounce) → Syncer → upstream, matching main.go. The three assertions nail the behavior deterministically: pull-on-STATUS (count 1), silent-skip inside the 200ms window (count still 1), pull-past-window after the sleep (count 2). Exactly the production path, well covered. LGTM once CI is green.

PR #180, opened 2026-07-07
fix(telegram): retry Change-sender identity fetch in the poll loop (#160)

### LINKED ISSUE #160: telegram: identity fetch is startup-only with no wait-for-serve; Change-sender silently disabled on co-start
## Problem

The Telegram client fetches the configured inbox identities (for the Change-sender picker, #139) exactly once at startup, best-effort. If the telegram container starts before serve's admin endpoint is listening (a normal co-start race under docker compose up), the fetch fails with connection refused, is logged as a warning, and the client comes up with **Change-sender permanently disabled** until the telegram process is restarted after serve is up.

Observed on the v0.7.0 deploy: telegram logged `telegram fetch inbox identities failed; change-sender disabled` at startup because serve's admin came up ~0.3s later. The queue/holds polls self-recover (they retry every poll interval), but the identity fetch does not retry, so the picker stayed off until a manual restart of the telegram container.

## Impact

Change-sender (the outbound wrong-identity guardrail) silently unavailable after any simultaneous restart, with only a startup warning to signal it. The operator sees a plain Approve with no Change option and may not realize the picker should be there.

## Fix options (pick one)

1. **Retry the identity fetch** with backoff until it succeeds (mirrors how the pollers already tolerate a not-yet-ready serve).
2. **Lazy-fetch on first Change tap** (and cache), so it never depends on startup ordering.
3. **depends_on: service_healthy** on serve in compose + a serve health endpoint gate, so telegram only starts once admin is up.

Option 1 or 2 is self-contained in the telegram client and robust regardless of orchestration; option 3 is a compose-only mitigation. Leaning 1 or 2.

## Refs
Fast-follow to #139 (Change-sender) / v0.7.0.

forge-rubigd:
Closes #160. Fast-follow to #139 (Change-sender) / v0.7.0.

## Problem

The Telegram client fetched the configured inbox identities (for the Change-sender picker) once at startup, best-effort. If telegram came up before serve's admin endpoint was listening (a normal docker-compose co-start race — observed ~0.3s on the v0.7.0 deploy), the fetch failed with connection refused, logged a warning, and Change-sender stayed **permanently disabled** until the telegram container was manually restarted.

## Fix — Option 1 (retry in the poll loop)

Fold the fetch into `pollLoop` via `ensureIdentities`: retried each cycle until it succeeds, then a no-op. This mirrors exactly how `poll`/`pollHolds` already tolerate a not-yet-listening serve, so a co-start race self-recovers on the next tick.

**Why Option 1 over lazy-fetch-on-first-tap:** the "Change sender" button is rendered by `decisionKeyboard` at **message-post time** (poll goroutine), gated on `len(identities) >= 2`. On a co-start race the button never appears, so there is no Change tap to lazily fetch on — Option 2 would need a bigger UX change (always show the button, fetch on tap). Retry-in-poll-loop fits the existing structure. (Reviewers concurred on this before implementation.)

## Concurrency

Identities move from write-once-at-startup to written-by-the-poll-goroutine while read by the callback goroutine (`identityKeyboard`) and the poll goroutine (`decisionKeyboard`), so they are now guarded by a `sync.RWMutex`; reads go through `inboxIdentities()`. The old `telegram.go` "read-only after Run's initial fetch, no lock needed" comment is retired.

## Logging

A failed fetch now logs at **debug** (the loop retries) instead of a one-shot warning; success logs an info line. A persistent serve-down is already signalled loudly by `poll`/`pollHolds` erroring each cycle, so this avoids a duplicate per-cycle warning during the brief normal race.

## Testing

- `TestEnsureIdentitiesRetriesUntilReady`: first cycle (serve 503) → nothing loaded, decision keyboard omits Change; second cycle (serve ready) → loaded, Change offered; third cycle → no re-fetch after success.
- `TestInboxIdentitiesRaceClean`: concurrent poll-goroutine writes + callback-goroutine reads, clean under `-race`.

No `adr/*` touch — standard 2-of-N.

### DIFF
--- a/internal/telegram/identity_retry_internal_test.go
+++ b/internal/telegram/identity_retry_internal_test.go
@@ -1,0 +1,96 @@
+package telegram
+
+import (
+	"context"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/yaad-index/darbaan/internal/admin"
+)
+
+// newClientForAdmin builds a bare Client pointed at a test admin server, with the
+// bot/handler machinery omitted (these tests exercise only the identity fetch).
+func newClientForAdmin(url string) *Client {
+	return &Client{
+		admin:  admin.NewClient(strings.TrimPrefix(url, "http://"), "tok"),
+		logger: slog.Default(),
+	}
+}
+
+// #160: the Change-sender identity fetch is retried by the poll loop until it
+// succeeds. A co-start race where serve's admin endpoint isn't listening yet
+// leaves Change-sender off for that cycle but self-recovers on the next, instead
+// of staying off until a manual restart.
+func TestEnsureIdentitiesRetriesUntilReady(t *testing.T) {
+	var calls atomic.Int32
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		// First cycle: serve not ready yet (stands in for connection-refused).
+		if calls.Add(1) == 1 {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		}
+		_, _ = w.Write([]byte(`[{"name":"work","identity":"w@x"},{"name":"personal","identity":"p@x"}]`))
+	}))
+	defer srv.Close()
+	c := newClientForAdmin(srv.URL)
+
+	// Cycle 1: serve not ready → nothing loaded, Change-sender omitted from the
+	// decision keyboard (Approve + Reject rows only).
+	c.ensureIdentities(context.Background())
+	assert.False(t, c.identitiesLoaded)
+	assert.Empty(t, c.inboxIdentities())
+	assert.Len(t, c.decisionKeyboard("42").InlineKeyboard, 2)
+
+	// Cycle 2: serve ready → identities loaded, Change-sender now offered (Approve +
+	// Change + Reject rows).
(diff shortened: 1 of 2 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve — the retry-in-poll-loop approach is the right fit and the implementation is clean.

**Correctness / concurrency (the load-bearing part):** all four `c.identities` sites are accounted for — the one-shot write in Run is removed, both keyboard reads (`decisionKeyboard`, `identityKeyboard`) route through `inboxIdentities()` under RLock, and the sole write in `ensureIdentities` is under Lock. No unguarded access remains. The wholesale-replace (never mutate-in-place) plus single-writer (only pollLoop calls `ensureIdentities`) invariant makes the returned slice header safe to range lock-free, as the comment claims. `identitiesLoaded` correctly makes it a no-op after first success, and the network call is done outside the lock so it never blocks readers.

**Logic:** `decisionKeyboard` gated on `len(inboxIdentities()) >= 2` means the button is simply omitted until identities land, then appears on every message posted after — the intended self-recovery, mirroring how poll/pollHolds already tolerate a not-yet-listening serve.

**Tests:** the retry-503 → ready → no-refetch sequence pins both the bug and the no-re-fetch invariant; the 1-writer/4-reader test is race-clean under -race and matches the real goroutine shape.

**Non-blocking nits:** (1) the test uses a 503 as a stand-in for connection-refused — fine since the retry keys off `err != nil` and any error source is equivalent, just noting it simulates the symptom rather than the exact dial-refused cause; (2) 'fetched ... lazily' in the struct comment reads a touch loose (it's eager-each-cycle-until-success, not on-demand) but the next clause clarifies. Neither needs a change.

Ship it.

PR #182, opened 2026-07-08
feat(telegram): push a proactive alert on reconcile cap-latch (#149)

### LINKED ISSUE #149: Telegram push on reconcile cap-latch (fast-follow to #140)

forge-rubigd:
Closes #149. Fast-follow to #140 / ADR 0026.

## Problem

When the reconcile safety cap latches an inbox (suspends reconciliation until an operator release), the operator learned of it only by **polling** — `reconcile status` / `GET /reconcile` + the serve-log warning. A latch sitting unnoticed means retraction is silently off for that inbox (benign — nothing purged, upstream untouched — but timely notification matters, especially for an enabled personal inbox).

## Fix — `pollReconcile`, the reconcile mirror of `pollHolds`

Add `pollReconcile` to the telegram poll loop: list the reconcile status via the admin API (the `reconcile held` warn-log in `reconcile.go` was already the documented hook for this), and push a Telegram alert for each inbox the cap has **newly** suspended. This is the ADR-0017-consistent shape — telegram is a client-over-API that polls; serve doesn't push to it — reusing the exact `poll`/`pollHolds`/`prunePosted*` structure.

**Dedup / re-notify (the case that matters most):** per-inbox dedup via `postedLatches`, pruned when an inbox is no longer suspended — so a latch alerts **once**, a release clears the entry, and a **fresh latch after release re-notifies**. Covered directly in `TestLatchDedupReNotifyCycle`.

## Notification-only, by design

The alert names the inbox + held-count and points at `darbaan reconcile release <inbox>`, but offers **no one-tap release button**. Releasing runs a cap-bypassed purge the operator should decide only after checking *why* the inbox latched (legitimate source removal vs. anomaly). The latch exists to force that deliberation; a one-tap release would undercut it. (Reviewers concurred on notification-only before implementation.)

## Impl note

Reconciliation is opt-in / off by default (ADR 0026), so `ReconcileStatus` returns 503 (unavailable) in the common config. `pollReconcile` logs a poll failure at **debug**, not error, so it never spams every cycle when reconcile is off — same reasoning as #160 (a genuine serve-down is already loud via `poll`/`pollHolds`).

## Testing

- `TestFormatLatch`: message carries inbox, held-count, "nothing has been purged", the CLI release pointer; no button.
- `TestLatchDedupReNotifyCycle`: post-once → clear-on-release → re-notify-on-re-latch.
- `TestPruneLatchedIsPerInbox`: keeps still-suspended, prunes released, never adds entries.

Single-package (telegram) change; no `adr/*` — standard 2-of-N.

### DIFF
--- a/internal/telegram/reconcile.go
+++ b/internal/telegram/reconcile.go
@@ -1,0 +1,94 @@
+package telegram
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-telegram/bot"
+
+	"github.com/yaad-index/darbaan/internal/admin"
+)
+
+// pollReconcile lists each inbox's reconciliation latch (ADR 0026) and pushes a
+// proactive alert for each inbox the safety cap has newly suspended, so the
+// operator learns of a latch without polling the status surface (#149). It is the
+// reconcile mirror of pollHolds: alert once per latch, and prune the dedup set
+// when an inbox is no longer suspended so a re-latch after an operator release
+// notifies again.
+//
+// Reconciliation is opt-in and off by default (ADR 0026), so the status call is
+// unavailable (503) in the common config; a poll failure is logged at debug — not
+// error — so it never spams every cycle when reconcile is off. A genuine serve
+// outage is already surfaced loudly by poll/pollHolds.
+func (c *Client) pollReconcile(ctx context.Context) {
+	statuses, err := c.admin.ReconcileStatus(ctx)
+	if err != nil {
+		c.logger.Debug("telegram poll reconcile status failed", "err", err)
+		return
+	}
+	c.pruneLatched(statuses)
+	for _, st := range statuses {
+		if !st.Suspended || c.seenLatch(st.Inbox) {
+			continue
+		}
+		if err := c.notifyLatch(ctx, st); err != nil {
+			c.logger.Warn("telegram notify reconcile latch failed", "inbox", st.Inbox, "err", err)
+			continue
+		}
+		c.markPostedLatch(st.Inbox)
+	}
+}
+
+func (c *Client) notifyLatch(ctx context.Context, st admin.ReconcileStatus) error {
+	_, err := c.bot.SendMessage(ctx, &bot.SendMessageParams{
+		ChatID: c.operatorID,
+		Text:   formatLatch(st),
+	})
+	return err
+}
+
+// formatLatch is the operator-facing latch alert. It names the inbox and the
+// held-count and points at the CLI release — deliberately NOT a one-tap button,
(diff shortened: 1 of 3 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Reviewed the full diff, traced the state machine, and verified against the codebase.

**What I checked**
- `admin.ReconcileStatus{Inbox, Suspended, HeldCount}` exists (internal/admin/service.go:53) and `Client.ReconcileStatus` is the GET /reconcile client — types line up, build/vet/race + CI green.
- **pollReconcile ordering is correct:** prune-then-alert. `pruneLatched` drops dedup entries for inboxes no longer in the suspended set, then the loop alerts only `Suspended && !seenLatch`. Post-once holds; release clears; re-latch re-notifies.
- **Error path is right:** `notifyLatch` failure -> Warn + `continue` **without** `markPostedLatch`, so a failed send retries on the next tick instead of being deduped as if delivered. This is the easy thing to get wrong; it is correct here.
- **Startup latch surfaces:** empty `postedLatches` on boot => a standing latch is unseen => alerts on first poll. Desirable — the operator learns of a pre-existing latch after a restart.
- **Concurrency:** `postedLatches` is touched only from the single pollLoop goroutine (unlike `posted`/`postedHolds` which also see callback handlers); the mutex is defensive-consistent with the existing pattern, no check-then-act race given the single writer.
- **Design intent preserved:** notification-only, no one-tap Release — the cap-latch must force deliberation before a cap-bypassed purge (ADR 0026). Message names inbox + held-count + CLI pointer + the non-destructive reassurance. 503-when-reconcile-off at debug, not error — no per-cycle spam. Matches the shape both reviewers signed off on.

Tests cover the three cases that matter: dedup/re-notify cycle (the flagged one), per-inbox prune independence, and message content/no-button. Clean, ADR-0017-consistent, well-scoped. LGTM.

Issue #59, opened 2026-06-26
Admin API: per-client scoped/revocable tokens as approval interfaces multiply


yaad-rubigd:
Surfaced in the #58 (ADR 0017) review, non-blocking. Today a single shared admin-API token grants full admin capability over the API. As approval interfaces multiply (CLI, Telegram bot, web admin UI, and the pre-screener client), that one shared token becomes a concentrated weak point — any holder has full approve/reject power and there is no way to revoke one client without rotating for all.

Add per-client, scoped, independently-revocable tokens to the admin-API contract, so each interface holds its own least-privilege credential. Consider alongside #50 (clients-over-API) and the first Telegram client. Ref: ADR 0017 (#58), ADR 0002. Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd.

PR #184, opened 2026-07-09
docs(adr): ADR 0029 — admin API per-client scoped tokens (#59)

forge-rubigd:
Design record for #59. **ADR-first / hard-gate** — no code in this PR; implementation follows in slices after this is approved.

## What it decides

Per-client, scoped, independently-revocable admin-API tokens, replacing the single shared full-capability token (`DARBAAN_ADMIN_TOKEN`) as approval interfaces multiply. Extends ADR 0017 (admin API contract); respects ADR 0002 (least authority) and ADR 0012 (secrets out-of-band).

**Shape (A)** — config-defined, mirroring ADR 0027 multi-agent:
- `admin_clients:` list, each = name + a least-privilege `scopes:` set; token out-of-band via `DARBAAN_ADMIN_TOKEN_<NAME>`, never in config.
- `auth` resolves bearer → client → scopes **constant-time** (no early-exit oracle, ADR 0027 `Verify` discipline), then authorizes the route's required scope: `401` unknown token vs `403` wrong scope.
- Opt-in back-compat: **no `admin_clients:` → the single `DARBAAN_ADMIN_TOKEN` is an implicit full-scope root**, unchanged. Retained as break-glass when clients are configured.
- Revocation = drop the client's token + restart; every other client is unaffected (the "rotate for all" problem is gone). Restart-scoped, not live.

## Addressing the review asks

- **Complete route → required-scope map** — all 11 admin routes mapped explicitly in the ADR, with `reconcile:read` called out for the Telegram client's #149 cap-latch alert (`GET /reconcile` each poll; deliberately *not* granted `reconcile:release`).
- **The (A)/(B) fork is explicit** — a dedicated "Alternative considered — runtime token management (rejected)" section lays out the runtime mint/revoke API with its trade-offs (it re-creates the concentrated surface #59 shrinks; bootstrapping; a mutable secret store), so the fork can be flipped at the hard-gate if there's a hard live-no-restart requirement.

## Gate

Touches `adr/*` → hard-gate: needs the maintainer's APPROVE on HEAD alongside 2-of-N. Slices proposed in the ADR: (1) `admincfg` parse/validate + scope map, (2) enforcement in `auth`, (3) wiring + docs + finalize.

### DIFF
--- a/adr/0029-admin-per-client-scoped-tokens.md
+++ b/adr/0029-admin-per-client-scoped-tokens.md
@@ -1,0 +1,175 @@
+# ADR 0029: Admin API — per-client scoped, independently-revocable tokens
+
+**Status:** Proposed (2026-07-09)
+
+## Context
+
+The operator admin API (ADR 0017) is a loopback-only, token-authenticated HTTP
+surface that `serve` exposes; each approval or admin interface — the `darbaan`
+CLI, the Telegram approval client, a future web admin UI, the agent pre-screener
+— is a separate process that reaches the core only through it, holding **only**
+an admin-API token (never the mail credentials).
+
+Today that is a **single shared token** (`DARBAAN_ADMIN_TOKEN`): the `auth`
+middleware constant-time-compares the presented bearer against the one configured
+token, and every route is equally gated by it. So every interface holds the
+**same, full-capability** credential.
+
+As interfaces multiply, that shared token is a concentrated weak point:
+
+- **No least privilege.** A read-only pre-screener that only needs to *list* the
+  queue holds the same token that can *approve*, *reject*, *expose*, *drop*, and
+  *release reconciliation*. Any holder has full power.
+- **No independent revocation.** There is no way to revoke one interface's
+  credential without rotating the single token — which breaks *every* other
+  interface at once.
+
+This ADR closes that gap while keeping the admin API's other invariants (ADR 0017
+loopback-only; ADR 0002 least authority + operator-gated; ADR 0012 secrets
+out-of-band). It was surfaced non-blocking in the ADR 0017 review (#59).
+
+## Decision
+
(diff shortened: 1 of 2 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
APPROVE (ADR-only). Both review asks are in and the design is darbaan-idiomatic and sound.

**What's right**
- Complete route→scope map, all 11 routes explicit. `reconcile:read` for the Telegram #149 alert deliberately **without** `reconcile:release` is exactly the right cut: the latch-alert credential can read the cap state but cannot run the cap-bypassed purge.
- (B) documented as rejected with the real trade-offs (re-creates the concentrated surface, bootstrap authority, mutable secret store), plus the 'flip the fork as its own follow-up ADR' escape hatch. Matches the acked (A) shape.
- Back-compat as implicit full-scope root, constant-time no-oracle resolution, 401-vs-403 split, `admincfg` beside `agentcfg` — all consistent with 0017/0002/0012/0027.

**Two non-blocking notes for the implementation slices (not ADR blockers)**
1. Slice 2 constant-time: `subtle.ConstantTimeCompare` short-circuits when the two inputs differ in length, so a raw per-token compare can leak token length through the loop. Hash both sides to a fixed width (e.g. SHA-256) before comparing so every comparison is over 32 bytes regardless of input length — keeps the miss path genuinely equalized.
2. Minor clarity: when `admin_clients:` **is** set, spell out whether the CLI is expected to use the retained `DARBAAN_ADMIN_TOKEN` root or a full-scope named client, so the intended least-privilege path for the CLI is unambiguous.

Design approved; the notes are guidance for slices 2/3. Still gated on the maintainer's APPROVE on HEAD + 2-of-N.

PR #185, opened 2026-07-09
feat(admincfg): admin_clients schema + scope vocabulary (ADR 0029 slice 1)

forge-rubigd:
Slice 1 of 3 for ADR 0029 (per-client scoped admin tokens, #59). Schema + validation only — **no enforcement** (that's slice 2).

## What

New `internal/admincfg`, mirroring `internal/agentcfg`:

- **Scope vocabulary** — `queue:read`/`queue:decide`, `holds:read`/`holds:decide`, `reconcile:read`/`reconcile:release`, `inboxes:read` as constants, plus `AllScopes()` / `ValidScope()`.
- **`RouteScopes`** — the authoritative route→required-scope map for all 11 admin routes (ADR 0017), keyed by ServeMux pattern. Pinned complete + consistent by `TestRouteScopesComplete` (every route present, every value a known scope). `reconcile` splits read vs release, so the Telegram #149 cap-latch alert (`GET /reconcile`) needs only `reconcile:read`, never `reconcile:release`.
- **`Client{Name, Scopes}`** + `Parse` (the `admin_clients:` list) + `Validate` (unique names, token-env collision guard, non-empty known-scope sets, no dup scopes) + `TokenEnv` (`DARBAAN_ADMIN_TOKEN_<NAME>`, shared `inboxcfg.EnvPrefix` mangle, secret out-of-band per ADR 0012) + `HasScope`.

## Deferred to later slices (per the ADR)

- **Slice 2** — enforcement: `auth` resolves a presented token → client → scopes (constant-time, with the fixed-width-hash-before-compare refinement from review) and authorizes the route's required scope (401 vs 403); the implicit full-scope root (back-compat when `admin_clients:` is unset). A slice-2 test will also cross-check that every route registered in `admin/http.go` has a `RouteScopes` entry (this slice pins the map's contents; slice 2 pins map↔routes).
- **Slice 3** — `serve` wiring + config example + README + ADR finalize.

## Testing

- Parse (+ empty→nil back-compat), `TokenEnv` mangle, `Validate` (dup name, token-env collision, empty/unknown/dup scope, blank name), scope vocabulary, `RouteScopes` completeness.
- Full `go build` / `go vet` / `gofmt -l` / `go test ./...` green.

No `adr/*` touch (the ADR merged separately) — standard 2-of-N.

### DIFF
--- a/internal/admincfg/admincfg.go
+++ b/internal/admincfg/admincfg.go
@@ -1,0 +1,180 @@
+// Package admincfg is the admin-API scoped-token config schema (ADR 0029): the
+// `admin_clients:` list that sits beside `inboxes:` (ADR 0023) and `agents:`
+// (ADR 0027). Each admin client is a named credential with a least-privilege set
+// of scopes; its token is supplied out-of-band via env, never in config
+// (ADR 0012), and gates only the admin API (ADR 0017) — it is unrelated to the
+// ADR 0027 agent grants that gate mail read/send.
+//
+// A config with no `admin_clients:` list yields the back-compat behavior: the
+// single DARBAAN_ADMIN_TOKEN is an implicit full-scope root, so an existing
+// single-token deployment is unchanged (opt-in, exactly like `inboxes:`).
+//
+// This increment is schema + validation + the scope vocabulary and the
+// authoritative route→scope map. ENFORCEMENT — resolving a presented token to a
+// client and authorizing the route's required scope in the admin auth middleware
+// — lands in the next increment; here the schema is parsed and validated, but
+// access is not yet gated.
+package admincfg
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/yaad-index/darbaan/internal/inboxcfg"
+)
+
+// The scope vocabulary (ADR 0029): capability-based, coarse, one required scope
+// per admin route. A read scope grants the list/show endpoints of a resource; a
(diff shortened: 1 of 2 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
APPROVE. Clean, defensive schema with thorough tests, correctly no-enforcement for this slice.

**Strong points**
- `TokenEnv` as the single source of truth for the mangle (Validate's collision guard calls it) means the load-time check and the runtime lookup can't diverge. The many-to-one env-collision guard (`a-b` vs `a_b` → same `DARBAAN_ADMIN_TOKEN_A_B`) rejected at load is exactly right; it'd otherwise silently share one token across two clients.
- Token is never a struct field (env-only, ADR 0012).
- `TestRouteScopesComplete` pins all 11 routes present, every value a known scope, and the reconcile read/release split, so a new or removed admin route without a scope binding fails the build. `Validate` is exercised across every failure mode (dup name, env collision, empty/unknown/duplicate scope, blank name).

**One optional nit (non-blocking)**
The scope vocabulary lives in two hand-maintained lists: the `const` block and the `allScopes` map. A new `Scope*` constant not added to `allScopes` wouldn't be caught directly (only indirectly, if it were used as a `RouteScopes` value). Deriving one from the other, or a test asserting every `Scope*` const is present in `allScopes`, would close that gap. Trivial, fine to defer.

The constant-time hardening and the implicit full-scope root are correctly deferred to slice 2 as noted. Approving; CI green.

PR #187, opened 2026-07-09
feat(admin): enforce per-client scopes in the auth middleware (ADR 0029 slice 2)

forge-rubigd:
Slice 2 of 3 for ADR 0029 (#59). The enforcement layer — builds on slice 1's `admincfg` (#185).

## What

Replace the single-token check in the admin `auth` middleware with **scoped credential resolution**:

- **`Server.creds`** — a set of credentials, each = `SHA-256` of the full `"Bearer <token>"` header value + a scope set. **The token is hashed to a fixed width before `subtle.ConstantTimeCompare`** (per the slice-1 review note): `ConstantTimeCompare` short-circuits on unequal length, so a raw per-token compare would leak token length; 32-byte hashes equalize every comparison.
- **`resolve()`** — hashes the presented header and constant-time-compares against every credential with **no early exit** and no which/whether-matched oracle (ADR 0027 `Verify` discipline).
- **`scoped(required)`** — authorizes the route's scope: unrecognized token → **401**, recognized-but-missing-scope → **403** (the distinct diagnostic split).
- **`register()`** — wires each route with its scope from `admincfg.RouteScopes` and **panics on an unmapped route** — the map↔routes invariant (a new admin route must ship its scope), enforced at startup.
- **`SetScopedClients`** — adds least-privilege per-client credentials; called by the serve wiring in slice 3. **`NewServer`'s signature is unchanged**, so a single-token deploy is byte-for-byte the same behavior (the root credential has every scope). Break-glass root retained.

## Also — closes the slice-1 review nit

`admincfg.allScopes` is now **derived from `RouteScopes`** (a scope is valid iff some route requires it), so the vocabulary has one hand-maintained list and the const-block/allScopes drift raised in slice-1 review is gone.

## Testing

- `TestScopedEnforcement`: reader client reaches `GET /queue` (200) but not `POST approve` (403) or `GET /reconcile` (403 — scope denied *before* the handler's 503); root full access; unknown/missing token → 401.
- `TestRootTokenFullAccess`: single-token deploy reaches every route (back-compat).
- `TestEmptyScopedClientTokenSkipped`; `TestNewServerScopesEveryRoute` (map↔routes, no panic).
- Existing admin/telegram/cmd tests pass unchanged. Full `go build` / `go vet` / `gofmt -l` / `go test -race ./...` green.

Slice 3 next: `serve` wiring (`SetScopedClients` from config) + config example + README + ADR finalize (+ the CLI-credential doc clarity note). No `adr/*` here — standard 2-of-N.

### DIFF
--- a/internal/admin/http.go
+++ b/internal/admin/http.go
@@ -1,6 +1,7 @@
 package admin
 
 import (
+	"crypto/sha256"
 	"crypto/subtle"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


yaad-rubigd (approved):
APPROVE. Security-critical slice, and it's correct and well-tested. CI green.

**The auth resolution is right**
- Fixed-width 32-byte SHA-256 hashes before `subtle.ConstantTimeCompare` — length can't leak, since both sides are always 32 bytes (the short-circuit-on-unequal-length concern is gone).
- `resolve()` compares against every credential with no early exit and no break, so timing reveals neither which credential matched nor whether one did. Real tokens are pre-hashed to fixed width, so `sha256`'s own length-dependence only touches the attacker's own input, not a secret.
- `register()` panicking on an unmapped route enforces the map↔routes invariant at wiring time (fail-closed, loud) — a new route without a scope can't ship.
- 401 (unknown) vs 403 (recognized but missing scope) split is clean.

**Tests are thorough and threat-aware**
- `TestEmptyScopedClientTokenSkipped` is the one I most wanted to see: skipping empty-token clients closes the bypass where a client with `Token: ""` would otherwise match an empty `Bearer ` header. Verified it 401s.
- Reader gets 403 on approve and on `GET /reconcile` *before* the handler (so it's the scope denial, not the handler's 503) — confirms the gate precedes handler logic.
- Back-compat single-token full access covered.

The `allScopes`-derived-from-`RouteScopes` change cleanly closes the slice-1 drift nit — RouteScopes is now the single hand-maintained list.

**One non-blocking note for slice 3 (wiring/docs)**
`SetScopedClients` silently skips a client whose token env is unset. That's the safe direction (fails closed, no accidental grant), but at serve-wiring time a one-line startup log for each skipped client ("admin client X skipped: its token env is unset") would make a misconfigured least-privilege interface diagnosable instead of mysteriously 401ing. Fits naturally in slice 3.

Approving.

sora-rubigd (approved):

PR #188, opened 2026-07-09
feat(admin): wire scoped admin clients + docs + finalize ADR 0029 (#59)

### LINKED ISSUE #59: Admin API: per-client scoped/revocable tokens as approval interfaces multiply

forge-rubigd:
Final slice (3 of 3) for ADR 0029 (#59). Wiring + docs + ADR finalize. **Hard-gate** — touches `adr/0029` (finalize) + `README.md`, so it needs the maintainer's APPROVE on HEAD alongside 2-of-N.

## What

- **serve wiring** — `resolveAdminClients` parses `admin_clients:` from config, reads each token from `DARBAAN_ADMIN_TOKEN_<NAME>`, and calls `adminSrv.SetScopedClients`. **A client whose token env is unset is logged (`Warn`: client + env) and skipped** — a misconfigured least-privilege interface surfaces as a startup warning instead of a silent 401 at request time (per the slice-2 review note). No `admin_clients:` → nil → single-token root only (back-compat).
- **`config.example.yaml`** — commented `admin_clients:` example + scope vocabulary. **Spells out the CLI credential** (the other review note): `DARBAAN_ADMIN_TOKEN` stays the full-scope root the CLI uses by default (no config change needed to keep the CLI working), with a named full-scope client as the opt-in alternative to put the CLI on its own revocable token.
- **`README.md`** — brief pointer to per-client tokens.
- **ADR 0029 finalized** — status Proposed → **Accepted**; the Auth-resolution paragraph now states the **fixed-width-hash-before-`ConstantTimeCompare`** (it short-circuits on unequal length, so hashing removes the length leak).

## Testing

- `resolveAdminClients`: returns only clients whose token env is set (unset skipped + warned), nil on no `admin_clients:` (back-compat), rejects an unknown scope at resolve.

## Wraps ADR 0029

Slice 1 `admincfg` (#185) + slice 2 enforcement (#187) + this. `closes #59`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,11 @@ into the core:
 - **CLI** — `darbaan queue …` lists and decides held messages.
 - **Telegram** — approve or reject from your phone (setup in [INSTALL.md](INSTALL.md)).
 
+Each client authenticates with a bearer token. By default that's a single shared
+`DARBAAN_ADMIN_TOKEN`; optionally, give each client its own least-privilege,
+independently-revocable token via `admin_clients:` (see `config.example.yaml` and
+adr/0029).
+
 ## Install / run
 
 Darbaan is a single Go binary; the easiest deployment is Docker Compose. See
--- a/adr/0029-admin-per-client-scoped-tokens.md
+++ b/adr/0029-admin-per-client-scoped-tokens.md
@@ -1,6 +1,6 @@
 # ADR 0029: Admin API — per-client scoped, independently-revocable tokens
 
-**Status:** Proposed (2026-07-09)
+**Status:** Accepted (2026-07-09)
 
 ## Context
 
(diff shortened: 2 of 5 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
APPROVE (final slice). Clean completion of the series; CI running, gate on green before merge, and the adr/* hard-gate still needs the maintainer's APPROVE.

**Both remaining notes fully landed**
- `resolveAdminClients` logs an `slog.Warn` for each admin client whose token env is unset and skips it, so the misconfigured-interface case surfaces as a startup warning instead of a silent 401 at request time. Exactly the diagnosability fix.
- `config.example.yaml` spells out the CLI credential: the root token by default (no config change needed to keep the CLI working), with an opt-in named full-scope client if you want the CLI on its own revocable token.

**Also good**
- ADR finalized (Proposed→Accepted); the Auth-resolution paragraph now documents the fixed-width-hash-before-`ConstantTimeCompare` rationale, so the security property is on record.
- Wiring is correctly ordered (build server → `SetScopedClients` once at setup → serve); a malformed `admin_clients:` fails serve startup rather than at request time.
- `resolveAdminClients` tested across set / unset-skip / back-compat-nil / bad-scope-reject.

Nothing blocking. Closes #59 and completes the ADR 0029 series.

PR #191, opened 2026-07-17
fix(imapsync): don't hang the client on an unresolvable upstream UID (#190)

### LINKED ISSUE #190: imapsync: UID FETCH hangs the client on unresolvable upstream UID (stale mapping stalls a whole account)
## Symptom
A downstream IMAP client's `UID FETCH <n> (RFC822)` against **one fronted account's INBOX** hangs indefinitely (client-side read timeout), continuously for ~20h. Header fetches and `RFC822.SIZE` for the *same* UIDs return instantly, and the other fronted accounts are perfectly healthy — so it is isolated to body/content retrieval on one account.

## Root cause (from darbaan logs)
Every fetch attempt logs, at INFO:
```
handling UID FETCH command: imapsync: content for 3213 unavailable: upstream uid 7239 not found
```
The local→upstream UID mapping is **stale**: the local UID maps to an upstream UID that no longer exists (message moved/expunged, or a UIDVALIDITY change upstream that wasn't reconciled).

## The bug
When imapsync cannot resolve the content ("upstream uid not found"), it does **not** send any response back to the client — the `FETCH` just hangs until the client times out. It should return a clean result (an empty/`NIL` body, or a tagged `NO`), so the client can skip that message and continue. Instead a single unresolvable UID **stalls every fetch** for that account, and because clients fetch oldest-first, one poisoned UID at the head blocks the entire poll — nothing after it is ever retrieved.

## Impact
Fully silent: logged at INFO, no error surfaced anywhere. A downstream poller stalled ~20h before a human noticed the absence of mail. A restart did **not** clear it (mapping is persisted); unclear whether the hourly reconcile loop will.

## Asks
1. **Don't hang** on unresolved content — return an empty body or tagged `NO` for that UID so the client proceeds to the next.
2. **Repair stale mappings** — a UIDVALIDITY change / vanished upstream UID should re-seed or reconcile the local↔upstream map rather than leaving a dangling entry that poisons fetches indefinitely.
3. **Sync visibility (the "better logs" ask)** — raise unresolved-content / sync-anomaly events above INFO, and emit a per-sync-cycle summary per account (fetched / failed / skipped counts, and current local↔upstream UID watermarks). A stalled or desynced account should be obvious from the logs, not require reading imapsync line-by-line.

## Evidence
`darbaan-1` logs repeat this every polling cycle (~10 min):
```
... imapsync: content for 3213 unavailable: upstream uid 7239 not found
```

forge-rubigd:
Fixes #190 — a downstream IMAP client's `UID FETCH … BODY[]` against one fronted account hung indefinitely (~20h silent outage) when a message's local→upstream UID mapping was stale (upstream message expunged).

## Root cause
`FetchContent` returned an error for the unresolvable UID, but the read face (`fetchMessage`) had already written a partial FETCH response (UID/FLAGS) and returned **without completing it** — no tagged result reached the client, so it hung on the read. Clients fetch oldest-first, so one poisoned UID at the head stalled the whole poll; logged only at INFO, nothing surfaced.

## Fix (the three asks)
1. **Don't hang.** Classify unresolvable content with a sentinel `inbound.ErrContentUnavailable` (wrapping both the vanished-UID and mailbox-reset cases). The read face serves an **empty body** for that message and completes the FETCH, so one unresolvable UID can't stall the rest. The stored ENVELOPE / SIZE / FLAGS still serve — only the body is empty.
2. **Repair the stale mapping.** A content fetch that finds the upstream UID **gone** drops the dangling local mapping and audits the retraction (same sink as reconcile) — at the moment of detection, not waiting for the hourly pass. A mailbox reset (UIDVALIDITY change) is deliberately **not** dropped per-message — that stays reconcile's re-seed concern.
3. **Sync visibility.** Unresolved content now logs at **WARN** (was effectively invisible), and each sync cycle emits a **per-account summary** (stored count, local UID watermark, UIDVALIDITY) so a stalled or desynced inbox is obvious from the log.

## Tests
- `listener.TestIMAPFetchContentUnavailableServesEmpty` — a BODY[] on a poisoned UID completes with an empty body, and a fetch of the whole range still returns every message (the outage scenario: head-of-queue poison no longer stalls the rest).
- `imapsync.TestFetchContentDropsVanishedMapping` — expunge upstream, then FetchContent returns the sentinel, drops the record, and audits a `retract`.
- `imapsync.TestFetchContentStaleUIDValidity` extended — the mailbox-reset case is the sentinel and is **not** dropped on a single fetch.

Gate (no CI on this repo → verified locally): build, vet, gofmt, full test suite, `-race` on the touched packages, golangci-lint 0.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -415,7 +415,7 @@ func (cli *CLI) newSenders(inboxes []inboxcfg.Inbox) (map[string]backend.Sender,
 // backend host are skipped (no sync). The syncers share one sync-state store
 // (keyed per inbox internally); the returned stop func closes it. An empty map
 // means no inbox syncs (gated off, like the stub sender).
-func (cli *CLI) newSyncers(inboxes []inboxcfg.Inbox, store inbound.InboundStore, mailOwner func(inbox string) string) (map[string]*imapsync.Syncer, func(), error) {
+func (cli *CLI) newSyncers(inboxes []inboxcfg.Inbox, store inbound.InboundStore, mailOwner func(inbox string) string, al audit.AuditLog) (map[string]*imapsync.Syncer, func(), error) {
 	syncers := map[string]*imapsync.Syncer{}
 	var state imapsync.StateStore
 	stop := func() {
@@ -449,6 +449,10 @@ func (cli *CLI) newSyncers(inboxes []inboxcfg.Inbox, store inbound.InboundStore,
 		// Per-inbox structured logger (#151): every sync/reconcile record this
 		// syncer emits is tagged with its inbox.
 		syn.SetLogger(slog.Default().With("inbox", in.Name))
+		// On-demand retraction audit (#190): a content fetch that finds the upstream
+		// UID gone drops the stale mapping and records a "retract", same sink as the
+		// reconcile loop.
+		syn.SetAudit(al)
 		// Gmail label write-through (ADR 0020 20c): capability-gated, harmless on a
 		// non-Gmail backend (reports not-supported → WriteKeywords uses plain keywords).
 		syn.SetLabelStore(imapsync.RawGmailLabelStore(in.Backend.IMAPHost, in.Backend.IMAPUsername, pass, mailbox))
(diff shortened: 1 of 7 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Correct and complete fix for #190, all three asks addressed. The read-face change is the key: on ErrContentUnavailable the resolver serves an empty body (raw=nil, err=nil) so the FETCH stays well-formed and completes — one unresolvable UID can no longer stall the poll — while transient/connection errors still propagate. The sentinel cleanly covers both the expunged-UID and the UIDVALIDITY-reset cases. FetchContent now drops the stale mapping at the moment of detection (best-effort RemoveSynced + retract audit) so it can't re-poison future fetches, instead of waiting for the hourly reconcile. The per-cycle summary log (stored / watermark_uid / uidvalidity) plus WARN on unavailable content gives the visibility that was missing. Matches the verified root cause (the mapped upstream UID was a genuine gap; upstream healthy and fast). LGTM.

PR #198, opened 2026-07-17
feat(sync): per-account sync-health observability + stall alert (#195)

### LINKED ISSUE #195: sync observability: surface per-account sync health beyond logs (queryable state + repeated-failure signal)
## Motivation
The ~20h silent outage in #190 was invisible until a missing downstream escalation gave it away. #191 fixed the hang and added a first layer of visibility: unresolved content now logs at **WARN**, and each sync cycle emits a **per-account summary** (stored count, local UID watermark, UIDVALIDITY). Good — but it's still "visible only if you tail the log." A gate whose sync is silently degrading should surface that health without someone reading logs.

## Proposal
Build on #191's summary; make sync health queryable and make persistent failure loud.

1. **Per-account sync-health record** — track, per fronted account: last-successful-sync timestamp, consecutive-error count, last error (string/class), and the current UID watermark + UIDVALIDITY (already computed for the #191 summary). Expose it via the admin API (same localhost/trusted posture as the rest of the admin API, ADR 0016), so a health check can answer "is every fronted account syncing, and since when?" without log-scraping.

2. **Repeated-failure signal** — when an account's consecutive sync errors cross a threshold (N cycles, or a staleness duration since last success), emit a distinct high-severity structured event a downstream can alert on — so a persistent stall is *loud*, not just one WARN per cycle buried in the log. (The downstream fetch client already sends a rate-limited alert on fetch failure; this is the server-side complement, catching stalls that don't even reach a client.)

3. *(Optional)* a small admin-API `sync-status` (or `/healthz`-style) command returning the per-account records, for scripted monitoring.

## Non-goals
- Not a full metrics/Prometheus stack — just enough to answer "which account is stalled and since when" and to make a persistent stall loud.

## Refs
#190, #191 (first-layer visibility), ADR 0016 (admin API), ADR 0019 (lazy read face).
Roles: admin yaad-rubigd / coder forge-rubigd / reviewer sora-rubigd.

forge-rubigd:
Fixes #195. Builds on #191's per-cycle summary log — makes inbound-sync health **queryable** and makes a persistent stall **loud**, so a silently-degrading account surfaces without someone tailing the log (the failure mode behind the ~20h #190 outage).

## What
1. **Per-account health registry** (`cmd/darbaan/synchealth.go`) — the sync loop records each cycle's outcome per fronted account: last-successful-sync timestamp, consecutive-error count, last error, and the UID watermark + UIDVALIDITY. Concurrency-safe (one writer goroutine per account + the admin reader).
2. **Queryable via the admin API** — `GET /sync-status` (new `sync:read` scope; localhost/trusted posture per ADR 0016) + a `darbaan sync-status` CLI command (tabwriter output: state / last-success / errors / uidvalidity / watermark / last-error). Answers "is every account syncing, and since when?" without log-scraping. Empty set (no account syncs) is a valid 200.
3. **Loud repeated-failure signal** — when an account's consecutive sync errors cross `--sync-stall-threshold` (default 3), a distinct high-severity `account sync stalled` structured event is emitted (inbox, consecutive_errors, stalled_since, last_error) — the server-side complement to the client-side fetch alert (#191), catching a stall that never even reaches a client.

## Notes
- Watermark is read from the local sync state store (no upstream contact), so it's cheap to call each cycle; a failed read keeps the prior watermark rather than clobbering it.
- Mirrors the existing `reconcile-status` plumbing exactly (service reader-injection, route→scope map, client method, CLI subcommand). The RouteScopes completeness test is updated for the new route.

## Tests / gate
`synchealth_test.go` (threshold crossing, recovery, watermark-keep-on-zero, sort/clamp), admin `syncstatus_test.go` (roundtrip + empty-when-no-reader), imapsync `TestSyncerWatermark`, admincfg route/scope pins updated. Local gate: build/vet/gofmt/`-race`/golangci-lint 0 (no CI on this repo). Non-goal: not a metrics/Prometheus stack — just "which account is stalled and since when" + a loud stall. Ref #190/#191, ADR 0016/0019.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Set up Go
-        uses: actions/setup-go@v7
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
           # Track the version pinned in go.mod (currently Go 1.26) so CI and the
           # module stay in lockstep — single source of truth.
@@ -37,7 +37,7 @@ jobs:
           fi
 
       - name: Lint
-        uses: golangci/golangci-lint-action@v9
+        uses: golangci/golangci-lint-action@ba0d7d2ec06a0ea1cb5fa41b2e4a3ab91d21278a # v9.3.0
         with:
           # Pin the linter version so CI matches the local run (same reasoning as
           # go-version-file above); Dependabot bumps it deliberately.
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -29,13 +29,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Set up Buildx
-        uses: docker/setup-buildx-action@v4
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
 
       - name: Log in to GHCR
-        uses: docker/login-action@v4
+        uses: docker/login-action@af1e73f918a031802d376d3c8bbc3fe56130a9b0 # v4.4.0
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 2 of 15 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Reviewed the full diff. Approve.

**Registry (synchealth.go).** RWMutex-guarded; both record methods take the write lock before touching the shared map via at(), and snapshot() reads under RLock — safe with one sync goroutine per account writing and the admin reader. Threshold clamped to ≥1. recordSuccess keeping the prior watermark when uidValidity==0 is the right call (a watermark read that failed shouldn't clobber a good cursor), and runSync signals exactly that by passing uidValidity=0 on a Watermark() error. Watermark() reads local state only — cheap to call every cycle, no upstream contact.

**Admin surface.** GET /sync-status gated by the new sync:read scope in RouteScopes; empty set is a normal 200, not an error; nil reader (no upstream inbox) degrades to empty rather than failing. Mirrors the reconcile-status plumbing exactly. CLI sync-status renders ok/erroring/STALLED with never/no-accounts handled.

**Stall event — one deliberate-design note (non-blocking).** recordFailure returns stalled = consec >= threshold, so the loud "account sync stalled" ERROR fires on every cycle once past the threshold, not only on the crossing. For a persistent-stall alarm that's the correct behavior — you want it loud and continuous until fixed (the failure mode this closes was 20h of silence), and a downstream alerting on the ERROR wants the repeat so it doesn't self-resolve. Flagging only so it's a conscious choice, not an oversight.

CI green. Tests cover the registry, the admin endpoint, and the watermark path. Closes #195 and completes the observability layer over #191's per-cycle summary. Will update the branch past the merged Action-pin change before merging.

PR #202, opened 2026-07-22
feat(provenance): strip X-Darbaan-* at the content-write chokepoint (ADR 0030 slice 1)

forge-rubigd:
ADR 0030 slice 1 (of #200) — the Layer-1 anti-spoof floor, shippable on its own. No config, no trust logic yet.

## What it does

Unconditionally strips the entire `X-Darbaan-*` header namespace from inbound mail **before the body blob is persisted**, so no upstream (or attacker-forged) provenance header can reach a served message. A sender cannot pre-forge `X-Darbaan-Trust: trusted` — the whole namespace is deleted regardless of value, case, count, or position, ahead of any trust logic a later slice adds.

## Where the guarantee lives

- **`internal/provenance.Strip`** — re-serializes only the header block and preserves the body byte-for-byte (the same `textproto` read/rewrite the send-path From rewrite already uses). Deletes every field whose name begins with `X-Darbaan-` (case-insensitive).
- Enforced at a single store chokepoint, **`inbound.bboltStore.putBlob`**, that **both** blob-write sites call — `SetContent` (lazily-fetched pending mail) **and** `put` (a present `Add`). The invariant can't live in one caller when there are two, so the strip is structural: no write path can persist an un-stripped blob.

> **Review fix (thanks to the CHANGES_REQUESTED catch):** the first push stripped only in `SetContent`, leaving the present/`Add` path (`put` → `blobs.Put`) as a bypass. Both sites now route through `putBlob`; added a test for the `Add` path.

## Parse-failure handling

The chokepoint feeds `Strip` both untrusted upstream mail (always well-formed RFC 822) and darbaan's own locally-generated blobs (bounces), some of which aren't full messages. A blob whose header block won't parse has no RFC 822 header block a compliant consumer reads a trust header from, so it **passes through unchanged** — **unless** its header region carries a namespace line, in which case it is **refused** (not persisted) rather than risk a lenient consumer reading a smuggled look-alike.

## Scope

Layer-1 only. Config-driven trust **stamping** (`X-Darbaan-Trust` value from the authenticated source), the note, the body banner, and the serve-path backstop are the later slices, one PR each, per the ADR.

## Cross-package touch

Two packages: new `internal/provenance` (the strip helper + its tests), and `internal/inbound` (one call added at the `SetContent` chokepoint + an end-to-end test). The inbound change is exactly the one call site plus its import.

## Tests

- `provenance`: strip across case/count/position; a body line that *looks* like an `X-Darbaan-` header is left untouched (only the header block is edited); clean passthrough when no namespace header is present.
- `inbound`: end-to-end — a forged `X-Darbaan-Trust: trusted` on upstream mail never reaches the served blob (`SetContent` return nor a later `Get`), with the unrelated header + body preserved.

Green locally: `go build` / `go vet` / `gofmt` / `go test ./...` + golangci-lint v9.3.0 (0 issues).

### DIFF
--- a/internal/provenance/provenance_test.go
+++ b/internal/provenance/provenance_test.go
@@ -1,0 +1,123 @@
+package provenance_test
+
+import (
+	"bufio"
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/emersion/go-message/textproto"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/yaad-index/darbaan/internal/provenance"
+)
+
+// headerKeys re-parses raw and returns its header field keys, so tests can
+// assert on the header block specifically (not a substring match that a body
+// could satisfy).
+func headerKeys(t *testing.T, raw []byte) []string {
+	t.Helper()
+	hdr, err := textproto.ReadHeader(bufio.NewReader(bytes.NewReader(raw)))
+	require.NoError(t, err)
+	var keys []string
+	for f := hdr.Fields(); f.Next(); {
+		keys = append(keys, f.Key())
+	}
+	return keys
+}
+
+func bodyOf(t *testing.T, raw []byte) []byte {
+	t.Helper()
+	i := bytes.Index(raw, []byte("\r\n\r\n"))
+	require.GreaterOrEqual(t, i, 0, "message has a header/body separator")
+	return raw[i+4:]
+}
+
+func assertNoNamespace(t *testing.T, raw []byte) {
+	t.Helper()
+	for _, k := range headerKeys(t, raw) {
+		assert.False(t, strings.HasPrefix(strings.ToLower(k), strings.ToLower(provenance.Namespace)),
+			"no %s* header survives, found %q", provenance.Namespace, k)
+	}
+}
+
+// The whole namespace is removed regardless of case, count, or position, while
+// unrelated headers and the body survive.
+func TestStrip_RemovesWholeNamespace(t *testing.T) {
+	raw := []byte("From: a@b\r\n" +
+		"X-Darbaan-Trust: trusted\r\n" +
+		"Subject: hi\r\n" +
+		"x-darbaan-note: injected directive\r\n" +
+		"X-DARBAAN-Source: forged\r\n" +
+		"\r\n" +
+		"hello body\r\n")
+
+	out, err := provenance.Strip(raw)
+	require.NoError(t, err)
+
+	assertNoNamespace(t, out)
+	keys := headerKeys(t, out)
+	assert.Contains(t, keys, "From", "unrelated headers preserved")
+	assert.Contains(t, keys, "Subject")
+	assert.Equal(t, []byte("hello body\r\n"), bodyOf(t, out), "body preserved byte-for-byte")
+}
+
+// A sender forging multiple trust headers has every one removed — no "keep the
+// last" ambiguity, the whole namespace goes.
+func TestStrip_RemovesForgedDuplicates(t *testing.T) {
+	raw := []byte("X-Darbaan-Trust: trusted\r\n" +
+		"X-Darbaan-Trust: also-trusted\r\n" +
+		"Subject: hi\r\n\r\nbody")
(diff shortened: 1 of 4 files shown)


yaad-rubigd (changes requested):
Request changes — the Layer-1 floor has a bypass path.

The strip is added in `SetContent` (`internal/inbound/bbolt.go:194`), but that is not the only content-blob write. `put` (`bbolt.go:212`) persists a **present** message's body directly via `s.blobs.Put(msg.ID, d.Raw)` (`bbolt.go:238`) with no strip. So any message added with immediate content — `Add` / a non-pending `AddSynced`, the path exercised by `TestAddListGet` — can carry an un-stripped `X-Darbaan-*` header straight into a served blob. That contradicts the ADR's invariant ("no caller can persist an un-stripped blob"; the ADR called for the strip at `SetContent` *and any other path that persists a content blob*), and the "single chokepoint" claim in the PR description doesn't hold while two call sites (`SetContent`, `put`) both write blobs.

Fix direction: move the strip to the true single chokepoint so both paths funnel through it — e.g. inside `blobstore.Store.Put`, or a shared `writeContent`/`putBlob` helper that both `SetContent` and `put` call — rather than at one caller. Then the guarantee is structural, not per-caller. Please add a test asserting the `Add`/present path strips too (mirror of `TestSetContent_StripsProvenanceNamespace`).

Everything else checks out: the namespace strip is case/count/position-correct, body preserved byte-for-byte, the body look-alike line is left alone, and malformed-headers → error → not-persisted degrades safely. Just needs the second write path covered so the floor is genuinely non-bypassable.

internal/provenance/provenance_test.go
```diff
@@ -0,0 +1,103 @@
...
+func TestStrip_LeavesBodyContentAlone(t *testing.T) {
+	body := "X-Darbaan-Trust: trusted\r\nthis line is body text, not a header\r\n"
+	raw := []byte("Subject: hi\r\nX-Darbaan-Trust: trusted\r\n\r\n" + body)
+
+	out, err := provenance.Strip(raw)
+	require.NoError(t, err)
+
+	assertNoNamespace(t, out) // the real header is gone
+	assert.Equal(t, []byte(body), bodyOf(t, out), "body left untouched, including a look-alike line")
+}
```
sora-rubigd, line 103: Missing test for the `malformed headers → not persisted` path — the `Strip` error branch documented in the ADR and in the `Strip` docstring. A test that passes a message with no header/body separator (or a bare binary blob) to `Strip` and asserts `err != nil`, plus a corresponding `SetContent` variant that asserts the error surfaces and nothing is persisted, would cover the one documented behavior path that isn't exercised here. Non-blocking (edge case in practice), but it's called out explicitly in the design so worth a test.

sora-rubigd (dismissed):

sora-rubigd (approved):

yaad-rubigd (approved):
Approve (respin). The bypass is fixed structurally: both blob-write sites (`SetContent` and `put`) now route through the shared `putBlob` chokepoint, so the strip no longer depends on a caller remembering. `TestSetContent_StripsProvenanceNamespace` and the new `TestAdd_StripsProvenanceNamespace` cover both paths; `msg.Raw = clean` keeps the returned message consistent with the stored blob.

The fail-open-on-unparseable change is sound. Untrusted upstream mail is well-formed RFC 822 (MTA-normalized, `Received:` prepended), so it always takes the strip path; the parse-failure path is effectively darbaan's own generated non-message blobs (bounces). `hasNamespaceHeaderLine` conservatively refuses any unparseable blob that still shows a namespace line in its header region (`TestStrip_MalformedWithNamespaceRefused`), and an inert non-message passes through (`TestStrip_NonMessagePassesThrough`). CI green.

Non-blocking observation: `hasNamespaceHeaderLine` skips a leading-whitespace line as a folded continuation, so a pathological leading-WS namespace first-line on the parse-failure path wouldn't be flagged. Not reachable in practice (attacker mail is well-formed → strip path; own blobs aren't attacker-authored) and it assumes consumers use standard header parsing, which the ADR already requires. Fine to leave as-is.

PR #204, opened 2026-07-23
feat(provenance): stamp X-Darbaan-Trust from per-inbox config at the chokepoint (ADR 0030 slice 2)

forge-rubigd:
ADR 0030 slice 2 (of #200), building on the merged slice-1 `putBlob` chokepoint.

## What it does

The gate now stamps `X-Darbaan-Trust: trusted|untrusted|unknown` on every stored message, computed **strictly from the authenticated inbox** — never from `From` or any header/body content, so a message can't influence its own trust. The header is always present on a well-formed message.

## The design call (arch-fit)

The stamp lives in the store's `putBlob`, not at `imapsync.go:381` as the ADR's pre-slice-1 wording had it. A stamp applied *before* the write would be re-stripped by slice 1's unconditional namespace strip, so **strip→stamp must be one atomic step at the same chokepoint slice 1 settled**. `provenance.Sanitize` does both in a single header-block rewrite, so no caller can persist a blob that is un-stripped *or* un-stamped — more faithful to the ADR's invariant than the literal wording.

Both review criteria baked in:
1. **Trust keyed strictly off the authenticated inbox** on the stored record. A per-inbox resolver is injected into the store (`inbound.New` option), built from config, keyed on the normalized inbox name, defaulting to `unknown`. `putBlob` resolves `s.trustOf(inbox)` — the record's inbox, never message content.
2. **Strip and stamp are one atomic unit** in `Sanitize` (no early-return between them), and both write paths — `SetContent` (upstream/FetchContent) and `put` (local `Add`) — go through it, so "always present" holds on both. The local path yields the inbox's configured level (or `unknown`), deterministically.

## Config

A per-inbox `trust:` block on `inboxcfg.Inbox`:

```yaml
inboxes:
  - name: work
    trust:
      level: trusted        # trusted | untrusted; omit → unknown
      note: "..."           # reserved (slice 3), validated now
      body_banner: false    # reserved (slice 4)
```

Only `level` is wired this slice. `note`/`body_banner` are parsed + reserved so the schema is stable; `note` is validated now (length ≤ 512, no control chars / CRLF — header-injection guard) even though it isn't stamped until slice 3. Read-only at runtime (ADR 0012); no new env-secret surface.

## Out of scope (later slices)

`X-Darbaan-Note` (3), body banner (4), serve-path backstop for legacy blobs (5).

## Also folded in

ADR 0030 **Status: Proposed → Accepted** (merged + slice 1 shipped).

## Tests

- `provenance`: `Sanitize` strip-then-stamp (forged value replaced, exactly one trust header, stamp-when-absent, non-message passthrough).
- `inboxcfg`: trust block parse; `TrustHeaderValue` mapping; validation (unknown level, over-long note, CRLF/control-char note rejected).
- `inbound`: both write paths (SetContent + Add) strip a forged value and stamp; a stamp value driven by an injected resolver keyed on the inbox; existing pending→present test updated for the stamp.

### DIFF
--- a/adr/0030-gate-stamp-trust-provenance-headers-on-inbound-mail.md
+++ b/adr/0030-gate-stamp-trust-provenance-headers-on-inbound-mail.md
@@ -1,6 +1,6 @@
 # ADR 0030: Gate-stamp trust/provenance headers on inbound mail
 
-**Status:** Proposed (2026-07-22)
+**Status:** Accepted (2026-07-23)
 
 ## Context
 
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -41,6 +41,7 @@ import (
 	"github.com/yaad-index/darbaan/internal/inboxcfg"
 	"github.com/yaad-index/darbaan/internal/listener"
 	"github.com/yaad-index/darbaan/internal/policy"
+	"github.com/yaad-index/darbaan/internal/provenance"
 	"github.com/yaad-index/darbaan/internal/signer"
 	"github.com/yaad-index/darbaan/internal/sluice"
 	"github.com/yaad-index/darbaan/internal/telegram"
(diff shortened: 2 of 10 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve (2-of-2). Architecture-fit and code-fit both pass.

Verified:
- Trust resolved strictly from the authenticated inbox via the injected resolver (map keyed on NormInbox, default unknown); never From/headers/body, so a message cannot influence its own trust.
- strip -> stamp is one atomic pass in rewrite(): the X-Darbaan-* Del loop then a single hdr.Set(TrustHeader, trust), no early return between them, one header-block re-serialize with the body preserved byte-for-byte.
- Both persist paths (SetContent and put/Add) route through putBlob, so 'always present' holds on well-formed messages; nil resolver falls back to the unknown fail-safe.
- The slice-1 parse-failure floor survives the Strip->rewrite rename: an unparseable blob is passed through only when it has no namespace line, else it errors and is not persisted (hasNamespaceHeaderLine guard intact).
- Config validation fails closed at load: unknown trust.level rejected; note bounded (<=512) and control-char rejected (incl. CR/LF/tab); note/body_banner reserved+validated ahead of slices 3/4.
- Anti-spoof holds: a forged X-Darbaan-Trust is deleted by the strip loop before the real value is set.

PR #205, opened 2026-07-23
feat(provenance): stamp X-Darbaan-Note from per-inbox config (ADR 0030 slice 3)

forge-rubigd:
ADR 0030 slice 3 (of #200), building on the merged slice-2 stamp chokepoint.

## What it does

The gate now also stamps `X-Darbaan-Note: <note>` when the authenticated inbox has a `trust.note` configured — in the **same atomic sanitize pass** as `X-Darbaan-Trust`. No configured note → no `X-Darbaan-Note` header.

## Design

- `provenance.Sanitize` now takes a `Stamp{Trust, Note}` and, in the one header-block rewrite, stamps trust (when set) and note (when set). The resolver was extended to carry both fields (`provenance.Stamp`), still keyed **strictly on the authenticated inbox**, never message content. Renamed the store option to `WithProvenanceResolver`.
- **Header-injection guard (the important bit):** the note value is header-sanitized at stamp time — CR/LF and other control chars dropped before `hdr.Set` — so a templated value can never inject a second header. `Set` guarantees **one `X-Darbaan-Note` max**. This is belt-and-suspenders over the slice-2 load-time note validation.
- **Anti-spoof inherited:** an inbound forged `X-Darbaan-Note` is in the `X-Darbaan-*` namespace, so the slice-1 strip removes it before the gate stamps its own (or none) — confirmed in a test.

## Config

No new config surface — this wires the `trust.note` field that already landed (parsed + validated) in slice 2.

## Out of scope (later slices)

Body banner (4), serve-path backstop for legacy blobs (5).

## Tests

- `provenance`: `Sanitize` stamps the note alongside trust; empty note → no header; the CR/LF/control injection-strip yields exactly one `X-Darbaan-Note` with the control chars dropped and no injected field.
- `inbound`: the store stamps the configured note and **replaces** a forged inbound one; a no-note inbox strips a forged note and stamps none.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -340,26 +340,26 @@ func (c *CLI) resolveAdminClients() ([]admin.ScopedClient, error) {
 }
 
 // openInbound opens the inbound (served mailbox) store per config, wiring the
-// per-inbox trust resolver so the content-write chokepoint stamps X-Darbaan-Trust
-// by authenticated inbox (ADR 0030).
+// per-inbox provenance resolver so the content-write chokepoint stamps
+// X-Darbaan-Trust (and X-Darbaan-Note) by authenticated inbox (ADR 0030).
 func (c *CLI) openInbound(inboxes []inboxcfg.Inbox) (inbound.InboundStore, error) {
-	return inbound.New(c.InboundType, c.InboundDB, inbound.WithTrustResolver(trustResolver(inboxes)))
+	return inbound.New(c.InboundType, c.InboundDB, inbound.WithProvenanceResolver(provenanceResolver(inboxes)))
 }
 
-// trustResolver maps an inbox name to the X-Darbaan-Trust value to stamp
-// (ADR 0030), keyed on the normalized inbox name so it matches the store's
-// record scope, and defaulting to unknown for an unconfigured inbox. It reads
-// only config — never message content.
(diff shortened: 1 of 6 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Architecture-fit and code-fit both pass.

Verified:
- Note is resolved strictly from the authenticated inbox: provenanceResolver builds Stamp{Trust, Note} per inbox (Note from config trust.note), keyed on NormInbox; unconfigured inbox and nil resolver both fall back to unknown trust / no note. Never reads message content.
- Atomic pass preserved: rewrite() does the X-Darbaan-* Del loop, then Set(TrustHeader) and Set(NoteHeader) for the non-empty fields, one header-block re-serialize, body byte-for-byte.
- Injection guard: headerSafe() drops all control runes (<0x20 and 0x7f), so CR/LF/tab in a note value can't inject a second header; hdr.Set gives exactly one X-Darbaan-Note. Belt-and-suspenders over the slice-2 load validation.
- Note stamped only when configured (empty leaves the header unset), consistent with the ADR 'present only when configured'.
- Anti-spoof inherited: a forged inbound X-Darbaan-Note is removed by the namespace strip before stamping; parse-failure floor (hasNamespaceHeaderLine) unchanged.

PR #206, opened 2026-07-23
feat(provenance): optional fenced body banner for top-level text/plain (ADR 0030 slice 4)

forge-rubigd:
ADR 0030 slice 4 (of #200) — the optional body banner, deliberately tight to the bulletproof case.

## What it does

When the authenticated inbox sets `body_banner`, the gate prepends an advisory fenced banner to a **top-level `text/plain`** body, reproducing the `X-Darbaan-Trust`/`-Note` it stamped plus an explicit "This banner is advisory; the X-Darbaan-* headers are authoritative." line. Default **off**.

The banner is belt-and-suspenders for LLM visibility; the headers are the authoritative signal and ship on every message regardless of MIME shape, so skipping the banner loses **zero** trust information.

## Scope decision (consensus in review)

v1 banners **top-level `text/plain` only**. Every other shape — multipart, nested MIME, non-ASCII-superset charset, undecodable CTE — is **headers-only, no banner, never an error**. `multipart/alternative` is split into its own **slice 4b**: a partial edit on a malformed boundary would leave a body worse than just stamping headers, and the "which shapes can I safely edit" fallback deserves its own review + test surface rather than riding along here. A base64 / quoted-printable `text/plain` body IS handled — decoded, bannered, re-encoded for that one part.

## Idempotency (the trap)

The banner lives in the body, **outside** the namespace strip, so a re-materialized blob must not stack a second banner. The banner is a PEM-style ASCII fence (`-----BEGIN/END DARBAAN TRUST BANNER-----`); on re-run it is detected at the top of the body and stripped — **exactly the block plus the single blank-line separator it wrote**, no greedy whitespace trimming — before a fresh one is written. Re-running the chokepoint on an already-bannered body is a **byte-level no-op** (tested, including a body with its own leading blank lines).

## Where it lives

`provenance.Stamp` gains a `Banner bool`, resolved (like trust/note) strictly from the authenticated inbox's `body_banner` toggle. The banner is applied in the same `rewrite` pass as the header stamp; `maybeBanner` returns the body untouched for any non-bannerable shape.

## Out of scope

`multipart/alternative` text-part banner (slice 4b), serve-path backstop for legacy blobs (slice 5).

## Tests

`provenance`: banner content + header-still-authoritative; **re-run no-op** (and with leading blank lines); non-`text/plain` and unknown-CTE → headers-only; base64 decode→banner→re-encode with a single banner on re-run; banner-off leaves the body alone. `inbound`: the resolver's `Banner` flag reaches the stored blob end-to-end.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -353,7 +353,7 @@ func (c *CLI) openInbound(inboxes []inboxcfg.Inbox) (inbound.InboundStore, error
 func provenanceResolver(inboxes []inboxcfg.Inbox) inbound.ProvenanceResolver {
 	m := make(map[string]provenance.Stamp, len(inboxes))
 	for _, in := range inboxes {
-		m[inbound.NormInbox(in.Name)] = provenance.Stamp{Trust: in.TrustHeaderValue(), Note: in.Trust.Note}
+		m[inbound.NormInbox(in.Name)] = provenance.Stamp{Trust: in.TrustHeaderValue(), Note: in.Trust.Note, Banner: in.Trust.BodyBanner}
 	}
 	return func(inbox string) provenance.Stamp {
 		if s, ok := m[inbound.NormInbox(inbox)]; ok {
--- a/internal/inbound/inbound_test.go
+++ b/internal/inbound/inbound_test.go
@@ -234,6 +234,24 @@ func TestSetContent_StampsConfiguredProvenance(t *testing.T) {
 	assert.Equal(t, 1, strings.Count(string(filled.Raw), "X-Darbaan-Note:"), "exactly one note header")
 }
 
+// The resolver's Banner flag flows to the stored blob: a body_banner inbox gets
+// a fenced top-of-body banner on a text/plain message (ADR 0030 slice 4).
+func TestSetContent_BannersWhenConfigured(t *testing.T) {
+	s, err := inbound.New("bbolt", filepath.Join(t.TempDir(), "inbound.db"),
+		inbound.WithProvenanceResolver(func(string) provenance.Stamp {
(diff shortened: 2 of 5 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Architecture-fit and code-fit both pass — the body-mutating slice went tight per consensus and is well-bounded.

Verified:
- Default off and if-gated: banner runs only when s.Banner (from the per-inbox body_banner toggle) is true; otherwise the body is written back byte-for-byte (io.ReadAll then Write).
- MIME bounds are surgical: bannerableText limits to top-level text/plain (RFC 2045 missing-CT default, ASCII-superset charset whitelist so an ASCII banner can't corrupt a non-ASCII encoding); every other shape returns the body untouched, headers-only, never an error.
- CTE handled for the single target part only: 7bit/8bit/binary identity, quoted-printable and base64 decode->banner->re-encode in the same scheme (CTE header stays valid); unknown/undecodable CTE -> headers-only.
- Idempotency is exact: applyBanner strips any existing block first; stripBanner consumes exactly the block + the single blank-line separator bannerBlock writes (not the message's own leading blanks), so re-run is a byte-level no-op — covered by the idempotent and leading-blank-lines tests. This also makes a re-processed already-bannered blob (slice 5) safe, and an attacker's fake top-of-body banner is replaced by the authentic one rather than spoofed.
- Banner is advisory-only and says so; the X-Darbaan-* headers remain authoritative and ship regardless of MIME shape, so a skipped banner loses no trust info. Note is headerSafe'd in the banner body too.

multipart/alternative correctly deferred to slice 4b. CI green (Analyze, CodeQL, check). Good to merge.

PR #207, opened 2026-07-23
feat(listener): serve-path sanitize-then-stamp backstop (ADR 0030 slice 5)

### LINKED ISSUE #200: gate: stamp trust/provenance headers on inbound mail for agent consumers (ADR-first)

forge-rubigd:
ADR 0030 slice 5 (of #200) — the serve-path backstop. Last core slice; closes the legacy gap.

## What it does

The read face now re-runs the **full `provenance.Sanitize`** (the same strip + stamp as the write chokepoint) on a message's raw before serving it, in `rawResolver` — the single memoized point every body FETCH resolves through.

- **Keyed strictly on the session's server-side selected inbox** (`s.selectedInbox`, the authenticated SELECT state), never anything read from the blob — so the serve-path trust anchor is as unspoofable as the write-path one. SELECT serves one authenticated mailbox and FETCH serves from it, so serving-inbox == the blob's stored inbox → each blob gets its true trust.
- Gives **pre-ADR / legacy blobs** their real trust on serve (rather than a stale `unknown`), and keeps served trust **fresh** if an inbox's trust config later changes.
- **Idempotent by construction**: strip removes any write-time stamp and the re-stamp re-adds the same value; the banner strip-then-rewrite already no-ops. So it's a safe belt-and-suspenders on *every* fetch.
- A blob that can't be safely re-stamped serves **empty** rather than un-sanitized (real stored blobs are sanitized at write → near-dead edge). Per-fetch cost is one header-block rewrite (+ banner if toggled) — acceptable per the ADR; **no caching** in v1.

## Wiring

A `ServeStamp` hook on `IMAPServerConfig`, built in `main` from the **same** provenance resolver the store uses (`provResolver`), so write-path and serve-path stamp identically. `nil` disables it (tests serve as stored, and stored blobs are already write-stamped).

## Test

A blob stored with `trust=unknown` (default resolver) serves as the current config's `trusted` via a `ServeStamp` resolver — exactly one trust header (freshness + idempotence), body preserved. End-to-end over the real IMAP FETCH path.

## #200 status after this

ADR ✅ · slice 1 (strip) ✅ · 2 (trust) ✅ · 3 (note) ✅ · 4 (banner) ✅ · **5 (serve-path backstop)** here. That's all core slices — the one remaining optional follow-on is **slice 4b** (multipart/alternative banner). This PR does **not** close #200 (4b pending).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -342,8 +342,8 @@ func (c *CLI) resolveAdminClients() ([]admin.ScopedClient, error) {
 // openInbound opens the inbound (served mailbox) store per config, wiring the
 // per-inbox provenance resolver so the content-write chokepoint stamps
 // X-Darbaan-Trust (and X-Darbaan-Note) by authenticated inbox (ADR 0030).
-func (c *CLI) openInbound(inboxes []inboxcfg.Inbox) (inbound.InboundStore, error) {
-	return inbound.New(c.InboundType, c.InboundDB, inbound.WithProvenanceResolver(provenanceResolver(inboxes)))
+func (c *CLI) openInbound(resolver inbound.ProvenanceResolver) (inbound.InboundStore, error) {
+	return inbound.New(c.InboundType, c.InboundDB, inbound.WithProvenanceResolver(resolver))
 }
 
 // provenanceResolver maps an inbox name to the trust + note to stamp (ADR 0030),
@@ -926,8 +926,12 @@ func (*ServeCmd) Run(cli *CLI) error {
 	if err != nil {
 		return err
 	}
+	// One provenance resolver, shared by the write chokepoint (the store) and the
+	// serve-path backstop (the IMAP read face), so both stamp identically by
+	// authenticated inbox (ADR 0030).
+	provResolver := provenanceResolver(inboxes)
 
-	inbox, err := cli.openInbound(inboxes)
+	inbox, err := cli.openInbound(provResolver)
 	if err != nil {
 		return err
 	}
@@ -1113,6 +1117,9 @@ func (*ServeCmd) Run(cli *CLI) error {
 		Addr:          cli.IMAPAddr,
 		TLSConfig:     tlsConfig,
 		AllowInsecure: cli.ListenerAllowInsecure,
+		ServeStamp: func(inbox string, raw []byte) ([]byte, error) {
+			return provenance.Sanitize(raw, provResolver(inbox))
+		},
 	}, auth, inbox, imapContentFetch(syncers, inbox), imapKeywordWriter(syncers), filters, guard, holdSpoof, pw.mailOwner, syncNow)
 	if err != nil {
 		return err
--- a/internal/listener/imap.go
+++ b/internal/listener/imap.go
@@ -34,7 +34,18 @@ type IMAPServerConfig struct {
 	Addr          string
 	TLSConfig     *tls.Config // enables STARTTLS; required unless AllowInsecure
 	AllowInsecure bool        // permit AUTH over plaintext — local/testing only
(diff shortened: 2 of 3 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Architecture-fit and code-fit both pass — clean close on the last core slice.

Verified:
- Serve-path re-runs the identical provenance.Sanitize as the write chokepoint, wired from ONE shared resolver (provResolver feeds both openInbound and ServeStamp), so write-path and serve-path stamp identically — consistent and idempotent by construction.
- Anchor is unspoofable: keyed strictly on s.selectedInbox (server-side SELECT state, ADR 0023), never anything from the blob. selectedInbox is set at SELECT and FETCH requires a prior SELECT, so it's always the authenticated serving mailbox; an unset/unconfigured inbox resolves to the unknown fail-safe.
- Fail-closed on error: a blob that can't be safely re-stamped serves raw=nil (empty body), never the un-sanitized bytes, so a smuggled X-Darbaan-* can't leak on serve. Near-dead edge since stored blobs are sanitized at write, but correct to guard.
- nil ServeStamp serves as-stored, which is still write-sanitized, so the backstop is pure belt-and-suspenders with no regression.
- Legacy gap closed: pre-ADR blobs now serve their real trust instead of a fallback unknown, and served trust stays fresh on a config change. TestIMAPServeStampReStampsFresh covers stored-unknown -> served-trusted end-to-end over FETCH.

Slice 4b (multipart banner) correctly left open on #200. CI green (Analyze, CodeQL, check). Good to merge.

PR #209, opened 2026-07-23
docs(readme): security-boundary callout — darbaan is not an authentication source

forge-rubigd:
Adds a prominent top-of-README `> [!CAUTION]` alert making the trust-model boundary explicit.

Substance: darbaan is a proxy that reads from an upstream mail server and does **not** itself verify a sender is genuine, so the provenance it stamps (`X-Darbaan-Trust` / `X-Darbaan-Note`, ADR 0030) is only as trustworthy as the upstream server's own SPF/DKIM/DMARC/spam filtering. Pointed at a server that authenticates senders well the signals are meaningful; pointed at one that does not, a forged `From` would be served as genuine.

Placed high — right after the intro, before the Outbound section — so it can't be missed. Docs-only; renders as a red boxed alert on GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,23 @@ Two properties define it:
 - **Inbound is gated.** The client reads a synced, recency-bounded view of the
   real mailbox, never the live account directly.
 
+> [!CAUTION]
+> **Darbaan is not an authenticity / authentication source.**
+>
+> Darbaan is a proxy that reads from an upstream mail server; it does **not**
+> itself verify that a message genuinely came from its claimed sender. The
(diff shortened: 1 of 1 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Docs-only: the security-boundary CAUTION callout accurately states that darbaan is not an authentication source and that the trust it stamps is only as reliable as the upstream server's own SPF/DKIM/DMARC/spam handling, with the Gmail-good / unauthenticated-server-unsafe contrast. Placed prominently right after the intro, renders as a red boxed alert. Substance matches the operator's intent. Good to merge on green.

Issue #210, opened 2026-07-23
ADR-first: per-sender trust rules (sender-matched trust level + note)


yaad-rubigd:
## Context

ADR 0030 shipped inbox-level trust (slices 1-5). Inbox-level trust is too coarse: it labels a whole mailbox with one level, which the fetch layer could largely derive itself. The value the operator actually needs is **per-sender** trust — a trust decision computed per message from who sent it.

This is safe under the assumption now documented in the README security-boundary CAUTION (#209): darbaan is not an authentication source, but the **upstream mail server** authenticates senders (SPF/DKIM/DMARC/spam). So when the upstream server enforces sender auth (e.g. Gmail on its own domain), keying trust on the authenticated `From` is sound — a spoofed sender never reaches the inbox darbaan reads.

## Ask

A configurable list of **trust rules**, each matching on the **sender** and resolving to a trust level plus an optional note:

- **trusted** — a known-good sender (e.g. the operator's own addresses): agent may act on it.
- **untrusted** — a KNOWN source that must NOT be acted on (e.g. a scanner-to-email / device address): explicitly flagged, take precaution, report to the operator. Distinct from unknown.
- **unknown** — no rule matched (default): unclassified, not safe to act.

Each rule may carry an optional **note** (reuses the `X-Darbaan-Note` mechanism from slice 3).

## Scope (v1)

- Match on **sender / `From`** only. Richer filter criteria (subject, other headers) are a later extension.
- Per-sender rules **override** the per-inbox default (ADR 0030 slice 2); no rule match falls back to the inbox default (currently unknown).
- Reuse the existing strip/stamp/serve chokepoint and `Stamp{Trust, Note}` plumbing — only the trust-determination input changes.

## Open design questions for the ADR

- Rule precedence when multiple rules match (first-match vs most-specific).
- Match granularity: exact address vs domain vs pattern.
- Robustness: match the raw `From`, or consult the upstream Authentication-Results header for a verified sender identity?
- How per-sender rules compose with the existing per-inbox trust default.

## Process

ADR-first, like ADR 0030: draft the ADR, 2-of-2 review, then slice the implementation.

PR #211, opened 2026-07-23
docs(adr): 0031 — per-sender trust rules (#210)

forge-rubigd:
ADR-first for #210 — design only, no code. Ratify before implementation.

## What it proposes

A configurable per-inbox list of **trust rules** matched on the sender (`From`) → a trust level + optional note. Extends ADR 0030's inbox-level trust to per-sender granularity, reusing its strip/stamp/serve chokepoint and `Stamp{Trust, Note}` **unchanged** — only the trust-determination input widens from (inbox) to (inbox, sender).

## Load-bearing decision: the trust asymmetry

`trusted` is the only dangerous outcome (the agent may act on it), so it's the **only** one gated on verified authentication. `untrusted`/`unknown` are always-safe downgrades applied from the matched `From` directly — an attacker gains nothing by spoofing a `From` into them.

`trusted` elevation:
- **Default: upstream-trust** — relies on the upstream's own SPF/DKIM/DMARC + spam filtering (the README `[!CAUTION]`, #209). Correct for a strong upstream (Gmail); its safety is the operator's upstream choice.
- **Opt-in hardening: `Authentication-Results` gate** — elevate to `trusted` only on `dmarc=pass` (or aligned `dkim=pass`) for the `From` domain. **The trusted `authserv-id` is operator-configured and required — no default trusts any/all authserv-ids.** `Authentication-Results` is outside darbaan's stripped `X-Darbaan-*` namespace, so a sender *can* inject a forged one; the gate reads **only** the header whose `authserv-id` matches the configured value and ignores every other. No trustworthy auth → `trusted` unavailable → falls back to `unknown` (fail-safe, never open).

## Also decided

- **Precedence:** most-specific-wins (exact address beats domain), order-independent; duplicate matchers rejected at load.
- **Granularity (v1):** exact address / domain; pattern/regex deferred.
- **Composition:** most-specific rule → the per-inbox default level (ADR 0030 slice 2) → global `unknown`. Removing all rules is exactly ADR 0030 behavior.
- **Integration:** the resolver widens to `(inbox, sender)`, reading config only; the chokepoint extracts `From` from the raw and resolves the `Stamp` — write- and serve-path derive from the same stored `From`, so it stays consistent + idempotent + legacy-covered.

Boundaries (README-pairing, From-only, address/domain-only, bounded Auth-Results parsing) and implementation slices (rules+resolution → Auth-Results gate → richer matchers) at the bottom.

## One route flagged for review

Whether the Authentication-Results gate is opt-in (this proposal), mandatory-for-`trusted` in v1, or a deferred slice — recommendation in the ADR is design-now / implement-opt-in.

ADR (hard-gate) — holding for explicit approval before it turns into code. Reviewers requested.

### DIFF
--- a/adr/0031-per-sender-trust-rules.md
+++ b/adr/0031-per-sender-trust-rules.md
@@ -1,0 +1,94 @@
+# ADR 0031: Per-sender trust rules
+
+**Status:** Proposed (2026-07-23)
+
+## Context
+
+ADR 0030 shipped inbox-level trust (slices 1–5): a whole mailbox is stamped with one `X-Darbaan-Trust` level. That is too coarse — it is a property of the *channel*, which a consumer could largely derive itself, not of *who sent a given message*. The value the operator actually needs is **per-sender** trust: act on the operator's own addresses, explicitly flag a known device (a scanner-to-email box) as not-to-act-on, and leave everything else unclassified.
+
+This is sound only under the boundary the README `[!CAUTION]` (#209) now states: darbaan is not an authentication source, but the **upstream** mail server authenticates senders (SPF/DKIM/DMARC/spam). When the upstream enforces sender auth (e.g. Gmail on its own domain), keying trust on the `From` is meaningful because a spoofed sender never reaches the inbox darbaan reads. Point darbaan at an upstream that doesn't, and a `From` is forgeable — so this ADR's central job is to make per-sender trust *no less safe than* that boundary, and to offer a way to harden past it.
+
+The three levels all get real use here:
+- **trusted** — a known-good sender (the operator's own addresses): the agent may act on it.
+- **untrusted** — a KNOWN source that must NOT be acted on (a device / scanner address): flagged, treat with precaution, report to the operator. Distinct from unknown.
+- **unknown** — no rule matched (the default): unclassified, not safe to act.
+
+## Decision
+
+Add a configurable, per-inbox list of **trust rules**, each matching on the sender and resolving to a level plus an optional note, reusing ADR 0030's `Stamp{Trust, Note}` and the strip/stamp/serve chokepoint unchanged — **only the trust-determination input widens**, from (inbox) to (inbox, sender).
+
+### The trust asymmetry (the load-bearing decision)
+
+`trusted` is the only dangerous outcome — it tells the agent it may act. `untrusted` and `unknown` are the cautious side; an attacker gains nothing by spoofing a `From` into them. So the auth requirement is asymmetric:
+
+- **untrusted / unknown** outcomes are applied from the matched `From` directly. No authentication gate — they only ever *withhold* trust.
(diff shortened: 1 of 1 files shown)


adr/0031-per-sender-trust-rules.md
```diff
@@ -0,0 +1,93 @@
...
+
+> **Route to confirm in review:** ship the Authentication-Results gate as opt-in (this proposal), make it mandatory-for-`trusted` in v1, or defer it to a later slice. Recommendation: design it now, implement it as an opt-in slice, so a strong-upstream operator stays simple and a weak-upstream operator can harden. The From-based match is a deliberate, bounded relaxation of ADR 0030's "never from message content" — bounded by this asymmetry, so it never *elevates* trust on unauthenticated content.
+
+### Rule shape
```
sora-rubigd, line 33: Slice 2 precision note: "aligned dkim=pass" needs a concrete definition in the slice 2 ADR/code. DKIM pass alone doesn't guarantee `From` alignment — a DKIM signature can be valid for a domain other than the `From` domain. `dmarc=pass` already encapsulates the alignment check (either SPF or DKIM alignment with the `From` domain, per DMARC policy). Recommend treating `dmarc=pass` as the primary gate in slice 2, with "aligned dkim=pass" defined as a fallback only for upstreams that emit per-method results without a `dmarc=` entry. Not a blocker for this design ADR — just flag it for the slice 2 review.

sora-rubigd (dismissed):

sora-rubigd (approved):

yaad-rubigd (approved):
Approve (design).

- Trust asymmetry is the load-bearing decision: only the 'trusted' verdict (act-on-it) is gated on verified auth; untrusted/unknown are always-safe downgrades applied from the matched From directly, so a spoofed From can never elevate trust.
- The Authentication-Results gate's own input is unspoofable: the trusted authserv-id is required with no any/all default, only that authserv-id's header is honored (Authentication-Results is outside the stripped namespace, so forged ones are ignored), and with no trustworthy result trusted-elevation is unavailable and falls back to unknown (fail-safe, never open).
- Auth check now precise: dmarc=pass primary (already requires From alignment); aligned dkim=pass (d= matches From domain) as fallback where no dmarc= result is emitted; a bare dkim=pass is explicitly insufficient.
- from_domain public-domain caveat documented (exact-address rules are the safe pattern for public domains).
- Most-specific-wins precedence is deterministic, duplicate matchers rejected at load, per-sender rules strictly refine the inbox default (removing all rules == ADR 0030 behavior).
- Reuses the ADR 0030 chokepoint + Stamp{Trust,Note} unchanged; only the trust-determination input widens to (inbox, From). Gate designed now, ships as opt-in slice 2; slice 1 is the upstream-trust model.

CI green. Design is complete and internally consistent.

PR #212, opened 2026-07-23
feat(inboxcfg): per-inbox per-sender trust rules + resolution (ADR 0031 slice 1)

forge-rubigd:
ADR 0031 slice 1 (of #210) — per-sender trust rules on the **upstream-trust model**. No Authentication-Results gate yet; that's the opt-in slice 2.

## What it does

Trust is now resolved **per message from its sender**, not just per inbox.

- **Config** — a `rules` list on the per-inbox `trust` block. Each rule matches the message `From` by exact address (`from`) or domain (`from_domain`) → a `trusted`/`untrusted` level + optional note. Validated at load: exactly one matcher, a trusted/untrusted level (`unknown` is the *absence* of a rule, not an outcome), a sane note (slice-3 rules), and **no duplicate matcher**.
- **Resolution** — `Inbox.SenderStamp(from)` is **most-specific-wins** (an exact-address rule beats a domain rule), then the inbox default level + note, then `unknown`. Case-insensitive; the banner toggle stays inbox-level; an empty/unparseable `From` matches no rule.

```yaml
trust:
  level: unknown            # no-rule-match default
  rules:
    - from: ops@example.com
      level: trusted
      note: "operator address"
    - from_domain: scanner.local
      level: untrusted
      note: "device — do not act; report to the operator"
```

## Integration (reuses the ADR 0030 chokepoint)

The trust-determination input widens from `(inbox)` to `(inbox, sender)`: `provenance.From` extracts the normalized `From` from the raw, and both the store's `putBlob` (write) and the serve-path backstop resolve with it. The strip/stamp chokepoint and `Stamp{Trust, Note, Banner}` are otherwise unchanged — write- and serve-path derive from the **same** raw `From`, so they stay consistent + idempotent + legacy-covered.

## Safety

The `From` is message-derived — a deliberate, bounded relaxation of ADR 0030's inbox-only anchor. The **trust asymmetry** keeps it safe: a `From` only ever *downgrades* to `untrusted`/`unknown` for free; elevation to `trusted` is sound only under the upstream-authentication boundary (the README caution), and slice 2 adds the opt-in `Authentication-Results` gate on top.

## Tests

`SenderStamp` resolution (most-specific / domain / default / empty-From / case-insensitive); rule validation (matcher / level / note / dedup); `provenance.From` extraction; and the store end-to-end resolving trust by sender at the write chokepoint.

## Cross-package touch

`inboxcfg` (rules + resolution — the bulk), `provenance` (one `From` helper), `inbound` (resolver signature widened + `putBlob` passes the From), `cmd/darbaan` (resolver + serve closure). No behavior change for an inbox with no rules — that is exactly ADR 0030.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -351,13 +351,13 @@ func (c *CLI) openInbound(resolver inbound.ProvenanceResolver) (inbound.InboundS
 // defaulting to unknown trust (no note) for an unconfigured inbox. It reads only
 // config — never message content.
 func provenanceResolver(inboxes []inboxcfg.Inbox) inbound.ProvenanceResolver {
-	m := make(map[string]provenance.Stamp, len(inboxes))
+	byName := make(map[string]inboxcfg.Inbox, len(inboxes))
 	for _, in := range inboxes {
-		m[inbound.NormInbox(in.Name)] = provenance.Stamp{Trust: in.TrustHeaderValue(), Note: in.Trust.Note, Banner: in.Trust.BodyBanner}
+		byName[inbound.NormInbox(in.Name)] = in
 	}
-	return func(inbox string) provenance.Stamp {
-		if s, ok := m[inbound.NormInbox(inbox)]; ok {
-			return s
+	return func(inbox, from string) provenance.Stamp {
+		if in, ok := byName[inbound.NormInbox(inbox)]; ok {
+			return in.SenderStamp(from) // per-sender rule → inbox default → unknown (ADR 0031)
 		}
 		return provenance.Stamp{Trust: provenance.TrustUnknown}
 	}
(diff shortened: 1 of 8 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Architecture-fit and code-fit both pass.

Verified:
- provenance.From uses mail.ParseAddress (handles display-name / angle-brackets), lowercases + trims, and returns "" on parse failure / no From / unparseable — empty From matches no rule and falls to the inbox default (fail-safe).
- Resolution is most-specific-wins and deterministic: matchRule returns an exact-address match immediately, a domain match only as fallback; validateRules rejects duplicate from / from_domain at load so at most one match per tier, order-independent.
- SenderStamp layers correctly: inbox default (level + note) → overridden by a matched rule's level/note; a rule level is trusted|untrusted only (unknown = absence, validated); banner stays inbox-level; no rules == exact ADR 0030 behavior.
- Validation: exactly-one-of from/from_domain, level constrained, note reuses the slice-3 length/control-char check, case-insensitive dedup on both matchers.
- Write/serve consistency: putBlob and the serve-path both resolve via provenance.From(raw) on the same stored raw (From is outside the stripped namespace, so it survives), so the stamp is identical and idempotent across the chokepoint; Sanitize/Stamp plumbing unchanged.
- Safety per the approved design: the From input can only downgrade for free (untrusted/unknown); trusted elevation is bounded by the upstream in this slice (the Authentication-Results gate is slice 2).

PR #215, opened 2026-08-01
docs(adr): ADR 0032 — incoming-message injection assessment

forge-rubigd:
## ADR 0032 — Incoming-message injection assessment

**Proposed. ADR only, no implementation. Hard sign-off gate — no code lands until this is approved.**

This replaces the earlier (never-merged) send-side pre-screener framing on this branch. The automated risk effort is **repointed from outbound to inbound**, where untrusted content actually enters — outbound stays the default-deny human sluice (ADR 0003 / 0004), so no automated send-verdict is load-bearing there. The real risk an autonomous mail agent carries is **prompt injection** on the incoming path: attacker-controlled content that hijacks the agent (confused-deputy) into a rash action with its own access.

### What it decides

1. **An incoming-assessment layer** that produces, for untrusted incoming mail, a **sanitized summary + a 0–100 numeric risk score + the list of factors** behind that score. It flags and scores; it never acts. The score is shown against **labeled bands** (0–33 low, 33–66 medium, 66–100 high) whose cutoffs are config — the number carries the resolution, the labels are a presentation layer.
2. **The score is composed by rules, never emitted by the model.** The isolated assessor detects **bounded content factors** (instruction-to-the-reader, secrets request, hidden/quoted-text directives, attachment-smuggled directives) and reports *which* it sees; a config **point-table** turns each flag into `+N`. A compromised assessor can lie about which factors it saw, but can never assert a number or take an action.
3. **Sender baseline + recipient position are added by the system, outside the model.** A configured sender baseline (trusted ≈ 0, unknown = `+N`, keyed on the ADR 0030/0031 trust the gate already stamps) and a recipient-position adjustment (To/CC/BCC) fold into the same total — fed into both the assessor prompt and the routing arithmetic.
4. **Short-circuit at the gate.** Contributions are additive; once the running total crosses the human-surface threshold, tallying stops and the message is held — still explainable by which factors carried it across.
5. **Cheap-to-expensive order.** The no-LLM sender + recipient terms are applied first; the expensive content assessor runs only if those have not already crossed the gate. A known-bad sender is held **without the assessor ever running**.
6. **Isolation** — the assessor is a zero-access component (no tools, no credentials, no send/store/admin power; a runtime client per ADR 0017 behind the ADR 0002 boundary). Only the sanitized summary + factor list + composed score cross to the privileged agent — never the raw payload as an instruction stream. The gate's safety does not depend on the assessor being uncompromised.
7. **Disposition — human-surface threshold is operator config, default HIGH (≈70).** Low + medium bands flow to the agent (with summary + score + factors); high is held for the human. This **replaces the earlier medium-held-by-default**; an operator wanting a stricter posture just lowers the threshold. The agent only ever sees the sanitized verdict, and any consequential action still meets the outbound human send gate — defense in depth.

**Source-trust gradient + tunability:** routing keys on the existing `X-Darbaan-Trust` stamp and per-sender rules (ADR 0030 / 0031), not a spoofable `From`; the point-table, sender baselines, recipient adjustment, band cutoffs, human-surface threshold, and assessor prompt are all config. The fixed contract is the *shape* — isolation, model-flags-not-numbers, system-composed score, verdict-only output, fail-safe routing.

**Fail-safe throughout:** assessor absent/unreachable/errored/timed-out → its content contribution is treated as not-cleared and the message is held, never auto-passed as low; the assessor can only *raise* scrutiny, never lower the human gate. Attachment text is in scope (a named injection vector) via inert, bounded extraction (size/type limits, no active-content execution).

### Content scope

Decoded text from body parts **and** attachments (e.g. PDF/HTML text), not just top-level `text/plain`. Extraction is inert and bounded; exact decoders/limits are an implementation detail.

### Files

- `adr/0032-incoming-message-injection-assessment.md`
- `adr/README.md` — index row for 0032 (also backfills the 0030/0031 rows)

### Gate

Architecture-decision PR — hard sign-off gate. Status stays **Proposed**; implementation (the assessor process + its zero-access seam, the factor detectors, the point-table / baseline / band / threshold config, the trust-gradient routing hook, the held/flagged disposition surface) is deferred to follow-up PRs after approval.

### DIFF
--- a/adr/0032-incoming-message-injection-assessment.md
+++ b/adr/0032-incoming-message-injection-assessment.md
@@ -56,6 +56,17 @@ secrets, or smuggle directives through quoted text or attachments raises the sco
 Coarse buckets are deliberate — they avoid false precision and can be refined as
 real traffic is seen (the ADR 0005 rationale).
 
+**Content in scope.** Because attachments are a named injection vector, the assessor
+is fed the message's **decoded text content — the body's text parts and text
+extracted from attachments** (e.g. the text of a PDF or an HTML part), not just the
+top-level `text/plain`; assessing only the outer body would blind it to a vector the
+threat model calls out. Extraction is **bounded** (per-part size and type limits, and
+no execution of active content — the assessor renders attachments to inert text, it
+never opens them as programs), so a hostile attachment cannot turn extraction itself
+into code execution. The exact decoders, media types, and limits are an
+implementation detail (follow-up); this ADR fixes only that decoded attachment
+content is *in* the assessment's scope and that extraction stays inert and bounded.
+
 ### 2. Isolation: the assessor is a zero-access component
 
 The assessment runs in a component with **no tools and no privileges** — no mail
@@ -95,10 +106,19 @@ rules, not re-derived from a spoofable `From` alone. ADR 0031's trust asymmetry
 carries over: only an authenticated or explicitly-vouched source earns the fast
 path, and an unknown/unauthenticated sender always gets full scrutiny — fail-safe.
 
-### 4. Disposition: low passes, high is held for the human
+### 4. Disposition: low passes, medium and high are held for the human
+
+All three score buckets map to a disposition; the fail-safe default groups the
+uncertain middle with the high end:
 
 - **Low risk → pass:** the summary + score travel with the message into the
   agent's view; normal flow.
+- **Medium risk → held / flagged for the human (v1 default):** medium is genuine
+  uncertainty, so it resolves toward *more* scrutiny, not less — it is held for a
+  person exactly like high. Splitting medium off to a lighter disposition (e.g.
+  pass-with-prominent-flag) is a policy call the operator can make later via the
+  tunable config (§5); v1 holds it. This keeps the coarse three-bucket score from
+  leaving any value undispositioned.
 - **High risk → held / flagged for the human:** the message is surfaced to the
   operator rather than flowing straight to the agent, so a person decides before
   the agent can act on it.
--- a/adr/0032-incoming-message-injection-assessment.md
+++ b/adr/0032-incoming-message-injection-assessment.md
@@ -1,6 +1,6 @@
 # ADR 0032: Incoming-message injection assessment
 
-**Status:** Proposed (2026-08-07)
+**Status:** Proposed (2026-08-08)
 
 Repoints the pre-screener concept from outbound to inbound.
 [ADR 0005](0005-agent-prescreener-plugin-risk-routing.md) introduced an automated
@@ -45,16 +45,26 @@ worth hijacking.
 ### 1. An incoming-assessment layer whose only job is injection defense
 
 Add an assessment layer that, for untrusted incoming mail, produces a **sanitized
-summary + a coarse risk score** (`low` | `medium` | `high`) describing how likely
-the content is an injection/hijack attempt. It **flags and scores; it never
-acts** — it cannot approve, send, mark-as-trusted, delete, label, or take any
-agent action. Its output is advisory signal, not a decision.
+summary + a numeric risk score in the range 0–100** describing how likely the
+content is an injection/hijack attempt, together with the **list of factors** that
+produced that score. It **flags and scores; it never acts** — it cannot approve,
+send, mark-as-trusted, delete, label, or take any agent action. Its output is
+advisory signal, not a decision.
 
-The score reflects *injection risk*, not importance or spam: content that tries to
-issue instructions to the reader, impersonate the operator, request credentials or
-secrets, or smuggle directives through quoted text or attachments raises the score.
-Coarse buckets are deliberate — they avoid false precision and can be refined as
-real traffic is seen (the ADR 0005 rationale).
+The score is a *risk score*, not an importance or spam score. It is presented
+against **labeled bands** so humans and routing rules can speak in words while the
+underlying number stays precise:
+
+- **0–33 → low**
+- **33–66 → medium**
+- **66–100 → high**
+
+The band cutoffs are **configuration**, tunable without re-plumbing anything: an
+operator can move the low/medium or medium/high boundary, and every downstream rule
+that keys on a band follows, because a band is only a labeled interval over the same
+0–100 number. This is the deliberate improvement over fixed coarse buckets — the
+score carries real resolution, and the labels are a presentation layer on top of it,
+not the primitive.
 
 **Content in scope.** Because attachments are a named injection vector, the assessor
 is fed the message's **decoded text content — the body's text parts and text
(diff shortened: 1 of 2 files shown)


yaad-rubigd (dismissed):
Approved (2-of-2 with sora-rubigd; operator ADR sign-off is the hard gate). Reviewed HEAD a7083e3ed: the three load-bearing properties are concrete and correct — (1) downgrade-only, a verdict only selects a chain and the lightest outcome is still a human tap, so no verdict can send/approve/reject/edit; (2) authenticated + message-bound, prescreen-scoped token over the loopback-only API bound to message id + content digest, so a low verdict can't be forged or replayed onto another/edited message; (3) fail-closed, absence/error/timeout/malformed/mis-bind/late-arrival all resolve to strict, and approver timeout still never auto-sends. reasons are non-routing and escaped (no injection via reasons), unknown flags fail to strict, no disposition field, auto-reject deferred, no-AI build preserved by runtime absence. Residual risk (stolen prescreen token → messages movable to the one-tap human path, never sent) is bounded and stated. ADR-only, Status Proposed.

sora-rubigd (dismissed):

adr/0032-incoming-message-injection-assessment.md
```diff
@@ -0,0 +1,166 @@
...
+stamps (ADR 0030 provenance, ADR 0031 per-sender rules), so assessment effort is
+spent where the risk is:
+
+- **Highest trust — operator hand-tagged / labeled mail:** the operator has
+  explicitly vouched for this specific message → fast path; assessment can be
+  skipped or minimal.
+- **Medium — the operator's own authenticated account, auto-flowing** (not
+  per-message tagged): reduced assessment.
+- **Lowest — other accounts, shared/family boxes, and unknown senders:** →
+  **full assessment.**
```
sora-rubigd: **Medium-risk disposition is unspecified.** Section 4 defines two outcomes: low → pass, high → held for the human. The coarse score has three buckets (low/medium/high) but medium has no defined disposition. Does medium pass with an appended warning, or is it held? This needs a call before implementation.

yaad-rubigd (dismissed):
Reviewed the ADR in full. Well-formed (Status / Context / Decision / Fail-safe / Consequences / Boundaries) and faithful to the agreed design: injection-defense purpose (not spam/importance), a zero-access isolated assessor whose only export is a sanitized summary + coarse score so a compromised assessor degrades to a wrong verdict rather than an action, source-trust-gradient routing built on the existing provenance stamp and per-sender trust rules, and a low-pass / high-hold disposition with the human release remaining the backstop. The fail-safe posture (absent or errored assessor treated as not-cleared, never auto-passed; assessor output always rendered as fenced, escaped data, never executed) and the explicit v1 read/flag-only boundaries (no auto-quarantine) are correct and load-bearing. Approving the document. This is an ADR: it stays Proposed and must not merge without the maintainer's explicit decision sign-off, independent of code-review approvals.

adr/0032-incoming-message-injection-assessment.md
```diff
@@ -0,0 +1,166 @@
...
+step the sluice already fully guards. That effort is better spent where the
+untrusted content actually *enters*: the inbound path.
+
+What is missing is a defense that reads incoming content *before* the privileged
+agent treats it as instruction, and tells the human (and the agent) how risky it
+looks — **without itself becoming a new injection foothold**. That is the hard
+constraint: the component that reads attacker-controlled bytes must have nothing
+worth hijacking.
+
+## Decision
```
sora-rubigd, line 43: **Attachment content scope is unnamed.** The threat model explicitly includes "smuggle directives through... attachments" as an injection vector, but the ADR does not say whether the assessor is fed decoded attachment content. PDFs, HTML parts, and inline images are real injection surfaces. Implementation will need to decide: assess raw MIME body only, or decode and assess attachment content too? Worth calling out now so the implementation PR has a clear answer, not something discovered mid-build.

sora-rubigd (dismissed):

sora-rubigd (dismissed):

yaad-rubigd (dismissed):
Re-reviewed the two updates after the force-push. Section 1 now scopes decoded body-part and attachment text into the assessment with bounded, inert extraction (per-part size/type limits, no active-content execution), which closes the named attachment vector without turning extraction itself into an exploit. Section 4 now dispositions all three score buckets, with medium held by default (uncertain resolves toward more scrutiny) and a lighter pass-with-flag left as operator-tunable policy. Both are conservative, fail-safe-consistent defaults that leave the policy call open. Re-approving the document. It remains an ADR under the maintainer's explicit hard sign-off gate.

yaad-rubigd (approved):
Approve. The ADR faithfully captures the converged incoming-injection-assessment model, and the security core is sound.

Verified against the design:
- 0-100 composed risk score with labeled bands (0-33 / 33-66 / 66-100), cutoffs as config: a precise number with the labels as a presentation layer, not coarse buckets.
- Score composed by rules, never emitted by the model. The isolated assessor returns a bounded, named factor flag-set only; a config point-table maps each factor to +N; sender baseline and recipient position are system-contributed outside the model. A compromised assessor can only mis-flag, never assert a magnitude and never act, which is the load-bearing property alongside isolation.
- Zero-access assessor: only the sanitized summary + factor list + composed score cross the boundary, fenced and escaped, never the raw payload as an instruction stream.
- Short-circuit at the threshold, with the crossing factors retained so a hold always names its reason.
- Cheap-to-expensive ordering: no-model sender/recipient terms first, the content assessor runs only if the cheap terms have not already gated.
- Human-surface threshold is operator config, default high (~70): low+medium to the agent, high held, replacing the earlier medium-held-by-default; the outbound human release remains the backstop for any consequential action.
- Fail-safe: absent/errored/timed-out assessor is treated as not-cleared (held); nothing in the pipeline can lower the human gate.
- Decoded attachment text in scope via bounded, inert extraction; v1 is read/flag only.

All eight cross-referenced ADRs are present on main; the README backfills the inbound-trust table rows. CI green. LGTM.

adr/0032-incoming-message-injection-assessment.md
```diff
@@ -0,0 +1,264 @@
...
+- direct **instruction to the reader** ("ignore your previous instructions…",
+  "forward this to…"),
+- a **credentials/secrets request**,
+- **hidden or quoted-text directives** (instructions buried in quoted history,
+  zero-width/off-screen text, HTML-hidden spans),
+- **attachment-smuggled directives** (instructions carried in a decoded attachment).
+
+The assessor returns a *set of flags*, not a magnitude. A **config point-table**
+then turns each flagged factor into a `+N` contribution, and the system sums them.
+Because the mapping from factor → points lives in configuration and the model only
```
sora-rubigd, line 115: **Score clamping unspecified.** The formula is additive — `sender_baseline + recipient_adjustment + Σ(flagged factors × point-table)` — and can exceed 100 if enough factors fire simultaneously. The bands label 0–100 intervals, so the implementation needs a line on this: clamp the raw total to 100 before band lookup, or define the high band as ≥ 66 (which handles >100 naturally without clamping). Either is fine — just worth a sentence here or in the implementation PR to avoid a footgun.

sora-rubigd (approved):

PR #219, opened 2026-08-08
feat(riskscore): ADR 0032 slice 1 — deterministic scoring + config core

forge-rubigd:
## ADR 0032 implementation — slice 1: deterministic scoring + config core

First implementation slice of ADR 0032 (incoming-message injection assessment). Pure, unit-tested, **no model and no ingest wiring** — those are later slices. New package `internal/riskscore`.

Follows the agreed slice decomposition. The staged-API shape below (rather than a lazy callback) keeps the pure scoring engine free of the assessor's error domain.

### What it does

Composes the injection-risk score by a fixed rule and decides disposition. **The model never produces the number** — an isolated assessor (a later slice) only reports which bounded factors it sees; this package turns those flags into points.

```
score = sender_baseline + recipient_adjustment + Σ(flagged_factor × points)   (clamped to 0..100)
```

- **Sender baseline** keys on the trust the gate already stamped (`provenance.Trust*`, ADR 0030/0031) — never a re-read of `From`. `map[trust]→points`, with the `unknown` level required as the fail-safe floor.
- **Recipient adjustment** (to/cc/bcc) and the **factor point-table** are config. The point-table is authoritative: a flagged factor with no entry contributes zero (keeping the detector's factor set aligned with the table is the assessor slice's concern).
- **Bands** are labeled intervals over the number (0-33 low / 33-66 medium / 66-100 high), cutoffs config.
- **Human-surface threshold** default is HIGH (70): low+medium → agent-handled, high → held.

### Staged API (gates the assessor *invocation*, not just the sum)

```go
cheap := s.CheapScore(trust, recipient)   // no-model prefix: baseline + recipient; reports Gated
var factors []riskscore.Factor
if !cheap.Gated {                          // cheap-first: only pay for the assessor when it can change the outcome
    factors = assessor.Flags(...)          // the later, fallible RPC lives in the orchestration layer
}
result := s.Compose(cheap, factors)        // short-circuits on a gated prefix; else composes + bands + dispositions
```

The engine stays pure — the RPC error domain (timeouts/errors) is owned by the orchestration layer (slice 4), which routes a failed assessment through the fail-safe. `Compose` also short-circuits defensively if handed a gated prefix (tally stops at the cheap total, ADR 0032 §4), so a known-bad sender is held without the assessor ever running.

### Clamp resolution (review carry-over)

The additive total can exceed 100, so it is **clamped into [0, 100] before banding** — the score stays a real 0-100 number and bands stay bounded intervals. Covered by a `130 → 100 → high/held` test.

### Fail-safe

`NotCleared(reason)` holds regardless of score — the orchestration routes an absent/unreachable/errored/timed-out assessment here, never auto-passing it.

### Config

`Parse([]byte)` unmarshals YAML, overlays onto `DefaultConfig` **per key** (partial configs are valid), then validates (ordered band cutoffs, threshold in range, non-negative weights, known recipient positions, required `unknown` floor).

### Tests

testify, **97.4%** coverage: band boundaries, over-100 clamp, short-circuit stops the tally, cheap-first gating, unknown-trust fail-safe floor, unknown-factor zero, `NotCleared`, config parse/merge/validation. gofmt / go vet / `go test -race` / golangci-lint v2.11.4 all clean locally.

### Explicitly deferred to later slices

Content extraction (bounded/inert body+attachment text), the isolated zero-access assessor client (factor detection / the model call), and the ingest/serve routing hook + disposition surface.

### Gate

Implementation PR — 2-of-2 review, no operator gate (ADR 0032 is Accepted; the operator gate was the ADR).

### DIFF
--- a/internal/riskscore/config.go
+++ b/internal/riskscore/config.go
@@ -1,0 +1,158 @@
+package riskscore
+
+import (
+	"fmt"
+
+	"github.com/yaad-index/darbaan/internal/provenance"
+
+	"gopkg.in/yaml.v3"
+)
+
+// unknownTrust is the trust level used as the sender-baseline fail-safe floor:
+// an unrecognized or empty trust maps here (ADR 0030's fail-safe default).
+const unknownTrust = provenance.TrustUnknown
+
+// Config is the tunable schema for the deterministic scorer (ADR 0032 §7). Every
+// weight is operator configuration; only the composition shape is fixed.
+type Config struct {
+	// SenderBaselines maps a stamped trust level (provenance.Trust*: "trusted",
+	// "untrusted", "unknown") to the baseline points a message from such a source
+	// starts with. Keyed on the trust the gate already stamped (ADR 0030/0031),
+	// never on a re-read of From. Must define the "unknown" floor.
+	SenderBaselines map[string]int `yaml:"sender_baselines"`
+	// RecipientAdjust maps a recipient position (to/cc/bcc) to its point
+	// adjustment; an unlisted position adds nothing.
+	RecipientAdjust map[Recipient]int `yaml:"recipient_adjust"`
+	// FactorPoints is the point-table: each flagged content factor's contribution.
+	// The table is authoritative — a flagged factor with no entry adds nothing.
+	FactorPoints map[Factor]int `yaml:"factor_points"`
+	// Bands are the labeled-interval cutoffs over the 0-100 score.
+	Bands Bands `yaml:"bands"`
+	// Threshold is the human-surface cutoff: a composed score at or above it is
+	// held for a human. Default is HIGH (see DefaultConfig); lower it for a more
+	// conservative posture.
+	Threshold int `yaml:"threshold"`
+}
+
+// Bands holds the band cutoffs. A score below LowMax is low; below MediumMax is
+// medium; at or above MediumMax is high.
+type Bands struct {
+	LowMax    int `yaml:"low_max"`
+	MediumMax int `yaml:"medium_max"`
+}
+
+// DefaultConfig returns the v1 default weights: a source-trust gradient on the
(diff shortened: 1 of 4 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Reviewed the full implementation; correct and faithful to the ADR.

Verified:
- Staged API gates the assessor invocation, not just the arithmetic: CheapScore returns Gated, so orchestration can hold a known-bad sender without invoking the assessor. Compose defensively re-short-circuits on a gated prefix.
- Security core intact: Compose consumes a factor presence-set and applies the config point-table, so the number is system-composed. A compromised assessor can only misreport which factors it saw, never assert a magnitude.
- Clamp to [0,100] before banding, covered by the 130 to 100 to high test.
- Short-circuit stops the tally and leaves Factors nil (assessor skipped). Correct; a mid-factor short-circuit is unnecessary, since the factors are cheap integer lookups once the flag set exists and the only meaningful skip is the cheap prefix.
- Fail-safe: NotCleared holds regardless of score; unrecognized trust falls back to the higher unknown floor (required by validate); unlisted recipient position and unknown factor contribute zero (table authoritative).
- Config: pointer scalars distinguish an omitted key from an explicit zero; per-key map merge onto defaults; validation covers band ordering, ranges, the required unknown floor, and non-negativity.
- Tests comprehensive; CI green.

Two non-blocking carry-overs for later slices:
1. Unknown-factor-contributes-zero is fine at this layer; the assessor slice should validate detector-to-point-table alignment so a newly-added detector factor cannot silently contribute zero.
2. NotCleared returns an empty Band; the slice-4 surface should check NotCleared/Disposition before rendering band or score, to avoid a blank label.

LGTM.

PR #221, opened 2026-08-08
feat(mailtext): ADR 0032 slice 2 — bounded/inert content extraction

forge-rubigd:
## ADR 0032 implementation — slice 2: bounded/inert content extraction

Second implementation slice of ADR 0032. New package `internal/mailtext`: a pure transform from raw RFC 822 bytes to the decoded, assessable text the injection assessor reads. It decodes transfer-encoding and charset (via go-message) and flattens markup — it **never executes, renders, or fetches** anything.

No scoring, no assessor, no ingest wiring — those are other slices. This produces the text; slice 3's isolated assessor consumes it and emits factor flags.

### Two threat-model-driven choices (flagging for review input)

1. **Extract every text variant, not the one a mail client would display.** A `multipart/alternative` can carry benign `text/plain` and a hidden directive in its `text/html` sibling; a human client shows one, so the assessor must see both. Extraction includes more, never less.
2. **HTML is flattened with all textual content kept**, including text a page would hide (`display:none`, zero-width, off-screen) — concealment is exactly the vector the assessor must catch. Only `<script>`/`<style>` bodies (code, not reader-facing prose) are dropped.

### HTML flattening — x/net/html tokenizer

The flattener uses the **`golang.org/x/net/html` tokenizer** (promoted from a transitive reference to a direct dep, pinned to the version already in the module graph — no transitive bumps). It streams tokens, collects text-token content, drops `<script>`/`<style>` element data **by construction** (their bodies are skipped, not pattern-matched), maps block tags to line breaks, and uses the tokenizer's entity-unescaped text. Strictly inert — it tokenizes only, never renders/executes/fetches. Boring and spec-conformant for the untrusted-parse layer; predictable on malformed HTML and lower-noise for the assessor than tag-soup stripping.

### Attachments (a named vector)

Text-typed attachments are extracted (HTML flattened); binary attachments (PDF, images) carry **metadata only** in v1 — their decoders are deferred (the ADR defers exact decoders). The assessor still sees an opaque attachment's presence (filename/type).

### Bounds (inert + memory-safe)

Per-part byte cap, total-text byte cap, part-count cap, and recursion-depth cap (a part-bomb guard), with a `Truncated` flag when any cap is hit. Defaults: 256 KiB/part, 1 MiB total, 200 parts, depth 20.

### API

```go
func Extract(raw []byte, lim Limits) (Content, error)   // error only when the message can't be parsed at all
type Content struct { Body string; Attachments []Attachment; Truncated bool }
```

Best-effort within the message (an unreadable part is skipped, not fatal); the fail-safe on a wholly-unparseable message (not cleared) lives in the orchestration layer (slice 4).

### Tests

testify, **95%** coverage: transfer/charset decode, all-variants extraction, hidden-text-kept + script/style-dropped, entity unescaping, text vs binary attachments, unnamed-attachment default, and the per-part / total / part-count / depth bounds. gofmt / go vet / `go test -race` / golangci-lint v2.11.4 all clean locally.

### Gate

Implementation PR — 2-of-2 review, no operator gate (ADR 0032 is Accepted).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/go-telegram/bot v1.22.0
 	github.com/stretchr/testify v1.11.1
 	go.etcd.io/bbolt v1.5.0
+	golang.org/x/net v0.21.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
--- a/go.sum
+++ b/go.sum
@@ -44,6 +44,8 @@ golang.org/x/net v0.0.0-20190620200207-3b0461eec859/go.mod h1:z5CRVTTTmAJ677TzLL
 golang.org/x/net v0.0.0-20210226172049-e18ecbb05110/go.mod h1:m0MpNAwzfU5UDzcl9v0D8zg8gWTRqZa9RBIspLL5mdg=
 golang.org/x/net v0.0.0-20220722155237-a158d28d115b/go.mod h1:XRhObCWvk6IyKnWLug+ECip1KBveYUHfp+8e9klMJ9c=
 golang.org/x/net v0.6.0/go.mod h1:2Tu9+aMcznHK/AK1HMvgo6xiTLG5rD5rZLDS+rp2Bjs=
+golang.org/x/net v0.21.0 h1:AQyQV4dYCvJ7vGmJyKki9+PBdyvhkSd8EIx/qb0AYv4=
+golang.org/x/net v0.21.0/go.mod h1:bIjVDfnllIU7BJ2DNgfnXvpSvtn8VRwhlsaeUTyUS44=
 golang.org/x/sync v0.0.0-20190423024810-112230192c58/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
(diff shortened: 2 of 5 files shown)


sora-rubigd (dismissed):

sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Tokenizer revision is clean and correct.

- htmlToText streams x/net/html tokens: text is collected only outside script/style (a skip counter, incremented on a script/style StartTag and decremented on the EndTag under a skip>0 guard); block tags on start/end/self-closing emit a newline; TextToken data is entity-unescaped by the tokenizer; ErrorToken (EOF or fault) returns the text collected so far. Strictly inert — tokenize only, no render/execute/fetch.
- Hidden text (display:none, zero-width, off-screen) is still kept as real text nodes; only script/style bodies drop, now by construction rather than by regex match. An unclosed script degrades to browser-equivalent behavior (content to EOF is script data, i.e. non-reader-facing), so there is no asymmetric false-negative versus a human reader.
- readCapped now surfaces a non-EOF read fault as truncation (io.ReadAll returns nil on EOF, so this flags only real faults) — a partial read is never reported complete.
- x/net promoted transitive to direct at the version already in the module graph (v0.21.0), no transitive bumps.
- Both alternative variants extracted; all four bounds intact. Tests comprehensive; CI green.

One non-blocking observation: CommentToken content is dropped, not collected. This matches reader and agent behavior — HTML comments are non-rendered and are not reachable as agent instructions — so it is correct as-is; noting only that it is the one spot slightly less inclusive than the prior regex, with no reachable-injection consequence.

LGTM.

PR #223, opened 2026-08-08
feat(assessor): ADR 0032 slice 3 — isolated zero-access injection assessor

forge-rubigd:
## ADR 0032 implementation — slice 3: isolated zero-access assessor

Third slice of ADR 0032 — the **security keystone**. New package `internal/assessor`: an isolated component that reads only the extracted text of an untrusted message and reports which bounded risk factors it detects — **never a score, never an action**.

### Zero-access by construction

The `Assessor`'s only fields are a `Detector` and an optional timeout — no mail credentials, no send path, no store access, and **no network calls**. An injected payload has nothing to seize; a fully-injected assessor can only misreport which factors it saw. It can never assert a magnitude (the scorer composes that from the factors) or act. This is the load-bearing property of ADR 0032: the gate's safety doesn't depend on the assessor being uncompromised.

### Contract + fail-safe

```go
Detect(ctx, mailtext.Content) ([]riskscore.Factor, error)   // error on ANY failure; empty+nil = "assessed, clean"
Factors() []riskscore.Factor                                 // declared emittable set
```

The **error vs empty+nil** distinction is what makes the fail-safe reliable: a failure (detector error, cancelled/timed-out context, absent detector) is never reported as a clean assessment. The `error → riskscore.NotCleared` mapping stays in the slice-4 orchestration (disposition belongs in the scoring layer, not the assessor).

### Sanitized summary — zero attacker bytes

Hard invariant, enforced **structurally** (not by convention):
- `Assess` filters `Detect`'s output to the detector's **declared vocabulary** — snapshotted from `det.Factors()` at construction — before anything reads it. A compromised detector that returns `Factor("<attacker bytes>")` has it dropped; only system-defined factor names can ever surface. This is what makes the ADR's "an injected assessor can only misreport *which* factor it saw" true structurally.
- `summarize`'s signature is narrowed to `(factors, truncated bool)` — it has no access to the message content, so it *cannot* quote message bytes even by a future edit.

The summary is thus composed only of system-defined strings that don't vary with message content (factor names + a fixed truncation note). `Fence()` wraps any raw content that must accompany a *held* message for a human — neutralizing spoofed fence markers so the payload can't break out (the ADR 0006 principle). The machine-consumed path carries no attacker text.

### Alignment (slice-1 invariant)

`ValidateAlignment(det, cfg)` checks every factor the detector can emit has a point-table entry, so a detector factor can never silently contribute zero. Call it at wiring time to fail closed on a detector/config mismatch.

### v1 detector — heuristic, best-effort

`HeuristicDetector` is a deterministic pattern ruleset (ADR 0032 §5). **Best-effort / defense-in-depth, NOT the gate** — stated in the package doc, not only prose — it will miss novel phrasing; the human send-gate and the sender-baseline term of the score remain the real backstops. A model-backed detector is a fast-follow behind the same `Detector` contract, behind the isolation this slice proves out. It emits instruction / secrets / attachment-directive factors.

**hidden_directives is deferred** (tracked follow-up): distinguishing hidden-vs-visible needs hidden-segment tagging in extraction (slice 2 flattens hidden text into Body). A directive in a `display:none` span is still caught — as `instruction_to_reader` (40pts ≥ the 30 a hidden label would add, so no under-scoring on the single-label case); the only v1 loss is the extra "hidden is doubly-suspicious" stacking. `emittable ⊆ table` holds with the latent table entry.

### Tests

testify, **96%** coverage: nil-detector, dedupe/sort, detector-error and timeout fail-safe, zero-attacker-bytes summary, fence marker neutralization, alignment validation, and the heuristic detector (instruction/secrets/attachment factors, clean content, hidden-in-body, context cancellation, end-to-end). gofmt / vet / `go test -race` / golangci-lint v2.11.4 all clean.

### Gate

Implementation PR — 2-of-2 review, no operator gate (ADR 0032 is Accepted).

### DIFF
--- a/internal/assessor/assessor.go
+++ b/internal/assessor/assessor.go
@@ -53,11 +53,19 @@ type Detector interface {
 }
 
 // Assessor wraps a Detector with the isolation and fail-safe contract. It holds
-// no capability beyond the detector and an optional timeout — the zero-access
-// seam is that these are its only fields.
+// no capability beyond the detector, an optional timeout, and the detector's
+// declared factor vocabulary — the zero-access seam is that these are its only
+// fields.
 type Assessor struct {
 	det     Detector
 	timeout time.Duration
+	// declared is a snapshot of det.Factors() taken at construction. Detect's
+	// output is filtered against it (see Assess), so a compromised detector can
+	// only ever surface a declared, system-defined factor name — never an
+	// arbitrary string. This makes the "no attacker bytes cross the boundary"
+	// invariant structural rather than by-convention: even an injected assessor
+	// can only misreport WHICH known factor it saw, matching the ADR's claim.
+	declared map[riskscore.Factor]struct{}
 }
 
 // Option configures an Assessor.
@@ -78,6 +86,11 @@ func New(det Detector, opts ...Option) (*Assessor, error) {
 	a := &Assessor{det: det}
 	for _, o := range opts {
 		o(a)
+	}
+	declared := det.Factors()
+	a.declared = make(map[riskscore.Factor]struct{}, len(declared))
+	for _, f := range declared {
+		a.declared[f] = struct{}{}
 	}
 	return a, nil
 }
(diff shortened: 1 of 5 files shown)


internal/assessor/assessor.go
```diff
@@ -0,0 +1,161 @@
...
+		seen[f] = struct{}{}
+		out = append(out, f)
+	}
+	sort.Slice(out, func(i, j int) bool { return out[i] < out[j] })
+	return out
+}
+
+// summarize renders a sanitized, factual summary. The invariant it must hold: it
+// is composed only of system-defined strings that do not vary with message
+// content — factor names and a fixed truncation note. It never quotes message
```
sora-rubigd: summarize accepts the full Content but reads only c.Truncated. Consider narrowing to summarize(factors []riskscore.Factor, truncated bool) — it makes "no message bytes in the signature" explicit in types rather than by convention. TestSummaryHasZeroAttackerBytes is the current enforcement, but the tighter signature removes the affordance for a future edit to include c.Body for "helpful context." Non-blocking.

sora-rubigd (dismissed):

yaad-rubigd (approved):
Approve. Both structural changes verified; the summary invariant is now enforced by construction, not by convention.

- declared is snapshotted from det.Factors() at New(), and Assess filters Detect's output against it (filterDeclared, into a fresh allocation, before dedupeSort/summarize). An undeclared factor carrying arbitrary bytes is dropped, so only system-defined factor names ever reach the summary or the scorer. A compromised detector can only misreport which known factor it saw, never emit an arbitrary string.
- summarize's signature is narrowed to (factors, truncated bool): it structurally cannot reach message content, and a future edit cannot reintroduce a body quote because the bytes are out of scope.
- Tests cover both paths: a message-body sentinel never appears in the summary, and a mock detector returning an undeclared factor with attacker bytes has it dropped from both the factor list and the summary while the declared factor survives.
- Struct and package docs name the declared field and confirm it derives from system-defined constants, so the zero-access guarantee is unchanged.

Zero-access construction, Fence, ValidateAlignment, the heuristic ruleset, and the fail-safe contract all stand. CI green. LGTM.

sora-rubigd (approved):

PR #224, opened 2026-08-08
feat(screener): ADR 0032 slice 4a — assessment orchestrator (pure)

forge-rubigd:
## ADR 0032 implementation — slice 4a: assessment orchestrator

The routing-hook logic, split out from the live wiring so the security decision has a clean review surface. New package `internal/screener`: a **pure** component that composes the scorer, the isolated assessor, and content extraction into one disposition for an incoming message — pass to the agent, or hold for a human. **No ingest or serve plumbing** — that's slice 4b, which wires against this stable interface.

### `Screen(ctx, raw, trust, recipient) → Outcome`

**Cheap-first** (ADR 0032 §5): the no-model sender-baseline + recipient terms score first. If they alone reach the human-surface threshold, the message is held immediately and **neither content extraction nor the assessor runs** — a known-bad sender is held without ever invoking the assessor. Otherwise the content is extracted and assessed, and the factors compose the final score.

**Fail-safe** (ADR 0032): content that won't extract, or an assessor that errors or times out, resolves to `riskscore.NotCleared` (held) — never an auto-pass. `New` requires a non-nil scorer **and** assessor: once assessment is running, an absent component is a configuration error, not a silent pass.

`trust` is the level the gate already stamped (`provenance.Trust*`, ADR 0030/0031) — **never a re-read of From**. The `Outcome` carries the `riskscore.Result` (score/band/disposition/factors/not-cleared) plus the assessor's sanitized `Summary` (set only when the assessor actually ran; never message bytes).

### Recipient resolution

`ResolveRecipient(self, to, cc)` maps the mailbox against the envelope addresses: To beats Cc, and a mailbox in neither (Bcc, alias, list) is treated as the **cautious Bcc** position — an unresolved position never earns the safest score.

### Carry-over placement

- (a) cheap-first + short-circuit → `Screen`
- (b) error / absent → `NotCleared` → `Screen` + `New`
- (c) `ValidateAlignment(det, cfg)` stays a **wiring-time** check the startup path runs (slice 4b) — it needs the detector, which lives at construction in `main`
- (d) trust from the stamp → `Screen`'s `trust` param (4b feeds it from the `ProvenanceResolver`)
- (e) not-cleared has no band/score → `Outcome` documents keying on `Disposition`/`NotCleared`; the 4b disposition surface renders accordingly

### Tests

testify, **100%** coverage: cheap-first short-circuit (assessor **and** extractor both skipped, proven with a spy), agent-handled and held paths, extraction-error and assessor-error fail-safes, the real extractor on unparseable input, recipient resolution, and limit plumbing. gofmt / vet / `go test -race` / golangci-lint v2.11.4 all clean.

### Next

Slice 4b wires `Screen` into `imapsync.FetchContent`, routes "held" through the existing hold-for-human queue, and adds config + `main` construction with `ValidateAlignment` fail-closed at startup — **behind a default-OFF enable flag** (merge must not change live mail behavior until the operator activates it; startup-abort on misconfig regardless of flag state).

### Gate

Implementation PR — 2-of-2 review, no operator gate (ADR 0032 is Accepted; activation of the feature on a live inbox is the operator's separate call, handled in 4b).

### DIFF
--- a/internal/screener/screener.go
+++ b/internal/screener/screener.go
@@ -1,0 +1,137 @@
+// Package screener is the injection-assessment orchestrator (ADR 0032 slice 4a):
+// the routing-hook logic that ties the deterministic scorer, the isolated
+// assessor, and content extraction into a single disposition for one incoming
+// message. It is pure — no ingest or serve plumbing — so the security decision
+// (pass to the agent vs hold for a human) is reviewable and testable in one place.
+//
+// The order is cheap-first (ADR 0032 §5): the no-model sender-baseline + recipient
+// terms run first, and the expensive content assessor is invoked only when those
+// have not already crossed the human-surface threshold. A known-bad sender is
+// held without the assessor — or content extraction — ever running.
+//
+// Fail-safe (ADR 0032): any failure on the assessment path — content that will
+// not extract, or an assessor that errors or times out — resolves to a
+// not-cleared hold, never an auto-pass.
+package screener
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/yaad-index/darbaan/internal/assessor"
+	"github.com/yaad-index/darbaan/internal/mailtext"
+	"github.com/yaad-index/darbaan/internal/riskscore"
+)
+
+// ExtractFunc extracts assessable text from raw message bytes. It matches
+// mailtext.Extract and is injectable for testing.
+type ExtractFunc func(raw []byte, lim mailtext.Limits) (mailtext.Content, error)
+
+// Outcome is the result of screening one message.
+type Outcome struct {
+	// Result is the composed score, band, disposition, contributing factors, and
+	// (on a fail-safe hold) the NotCleared marker. Callers key routing on
+	// Result.Disposition and Result.NotCleared — never on the band/score of a
+	// not-cleared hold, which has no composed score.
+	Result riskscore.Result
+	// Summary is the assessor's sanitized, system-defined summary. It is set only
+	// when the content assessor actually ran (not on a short-circuit hold or a
+	// fail-safe not-cleared), and never contains message bytes.
+	Summary string
+}
+
+// Screener orchestrates one message's assessment.
+type Screener struct {
+	scorer   *riskscore.Scorer
+	assessor *assessor.Assessor
+	extract  ExtractFunc
+	limits   mailtext.Limits
+}
+
+// Option configures a Screener.
+type Option func(*Screener)
+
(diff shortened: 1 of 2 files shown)


internal/screener/screener.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return s, nil
+}
+
+// Screen assesses one incoming message and returns its disposition. trust is the
+// level the gate already stamped (provenance.Trust*, ADR 0030/0031) — never a
+// re-read of From — and recipient is the position the assessed mailbox held.
+//
+// Cheap-first: the sender-baseline + recipient terms are scored first; if they
+// alone reach the threshold the message is held immediately, and neither content
```
sora-rubigd, line 84: 4b note: riskscore.NotCleared receives the raw error string here, which could carry parser-derived bytes if the extractor error includes a content snippet. Result.Reason is an internal field and should not be surfaced to humans in the hold queue render — use Outcome.Summary (zero-attacker-bytes by construction) for any human-facing display. Worth an explicit check in 4b's review.

sora-rubigd (approved):

yaad-rubigd (approved):
Approve. The orchestrator implements cheap-first and the fail-safe correctly.

- Cheap-first: CheapScore runs first; a gated cheap prefix returns Compose(cheap, nil) as a short-circuit hold with neither extraction nor the assessor invoked. Otherwise extract, then Assess, then Compose(cheap, factors).
- Fail-safe: an extraction error, or an assessor error/timeout, both resolve to riskscore.NotCleared (held), never an auto-pass; New requires a non-nil scorer and assessor.
- trust is passed through to CheapScore as the stamped level, not a re-read From.
- Outcome.Summary is set only when the assessor actually ran (empty on a short-circuit or a not-cleared hold) and carries the assessor's system-defined summary, never message bytes.
- ResolveRecipient: To wins over Cc, and a mailbox found in neither resolves to the most cautious position (Bcc), so an unresolved recipient never earns the safest score.
- The only agent-handled path is a non-gated message that extracts and assesses cleanly below threshold; every gate-or-error path holds.

One note for 4b wiring (non-blocking here): ResolveRecipient matches on lowercased/trimmed strings, so it must be fed Envelope (bare) recipient addresses. If header To/Cc are ever passed instead, parse them first with net/mail, since a display-name header would otherwise mis-resolve a direct recipient to Bcc. That errs toward more scrutiny, but it would cause false-positive hold nudges, so pin the Envelope-address contract at the call site.

LGTM.

PR #225, opened 2026-08-08
feat: ADR 0032 slice 4b — live ingest integration (default-off)

forge-rubigd:
## ADR 0032 implementation — slice 4b: live ingest integration (default-OFF)

The final slice — wires the assessment pipeline (scorer + isolated assessor + extraction, composed by the slice-4a screener) onto the live inbound path, **behind a default-off enable flag**. Merging this changes **no mail behavior** until an operator opts in on their running instance (a separate post-merge decision).

### Ingest hook

`imapsync` gains an optional `AssessHook`, run **once per message at `FetchContent`**, off the blocking agent-read path. It resolves trust from the provenance stamp (ADR 0030/0031, **never a re-read `From`**) and the recipient position from the envelope To/Cc, screens the fetched content, and persists the disposition **atomically with the body** via a new `SetContentAssessed`. With no hook installed (assessment off), `FetchContent` is **byte-identical to before** — no `Screen` call, no assessment written.

### Disposition surface — reuse the existing hold queue

`inbound.Message` carries a flat, plain `Assessment` {Disposition, NotCleared, Score, Band, Factors, Summary}. **`Result.Reason` is deliberately excluded** — no raw error bytes are ever stored or rendered; only the system-defined Summary can reach a human.

An assessment-held message joins the guard/filter holds in **both** the read face (hidden until `HoldDecision==approved`) and the admin held list. Confirmed the existing semantics: a **single `HoldDecision` releases past every source** (guard `imap.go:284-286` + filter both key on it), so approving reveals past all — assessment slots into that exact model, no new inconsistency. `darbaan holds ls` surfaces the assessment reason per flagged message, so the operator sees **why** before the one `expose` that releases it (the only release path is human expose-by-id off the queue). The render keys on `NotCleared`/`Disposition` (no band/score on a not-cleared hold) and uses the Summary only.

### Startup wiring

Detector/scorer/assessor/screener are constructed and **`ValidateAlignment`-checked regardless of the flag** — a detector/config misalignment fails startup rather than lurking behind the flag and surprising the operator on flip. The hook is installed on the syncers only when enabled. v1 uses the default weights + the heuristic detector.

### Config

`assessment-enabled` (default `false`) and `assessment-timeout` (default `5s`), documented in `config.example.yaml`.

### Tests (the full checklist)

- flag-off → `FetchContent` writes no assessment (byte-identical)
- nil `Assessment` → never held (old + flag-off records untouched)
- `ValidateAlignment` aborts startup on misalignment **even when disabled**
- enabled hook holds an injection message
- read-face hides an assessment-held message until approved
- **held-by-both** (assessment + filter Hold) → reveals on one approval
- held list surfaces the reason; expose removes it
- not-cleared renders via Summary, **no band/score/Reason**
- `Outcome→Assessment` mapping **round-trips every disposition + band**; disposition constants match the scorer's

All packages `-race` green; golangci-lint v2.11.4 clean.

### Follow-ups (noted, not in scope)

Agent-facing surfacing of the score/summary on a passed message (the ADR's "travels with the message into the agent's view"); operator-tunable weights via config (v1 uses defaults); the hidden_directives factor (needs slice-2 hidden-segment tagging).

### Gate

Implementation PR — 2-of-2 review, no operator gate **on the merge**. Activating the feature on a live inbox (flipping `assessment-enabled`) is the operator's separate call after this lands dormant.

### DIFF
--- a/cmd/darbaan/assessment_test.go
+++ b/cmd/darbaan/assessment_test.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"testing"
+	"time"
+
+	"github.com/yaad-index/darbaan/internal/inbound"
+	"github.com/yaad-index/darbaan/internal/provenance"
+	"github.com/yaad-index/darbaan/internal/riskscore"
+	"github.com/yaad-index/darbaan/internal/screener"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func nilResolver(string, string) provenance.Stamp {
+	return provenance.Stamp{Trust: provenance.TrustUnknown}
+}
+
+// The store's disposition constants must equal the scorer's, since the mapping
+// copies the string across the stringly-typed boundary (a typo would be silent).
+func TestAssessmentDispositionConstantsMatch(t *testing.T) {
+	assert.Equal(t, string(riskscore.DispositionAgentHandled), inbound.AssessmentAgentHandled)
+	assert.Equal(t, string(riskscore.DispositionHeld), inbound.AssessmentHeld)
+}
+
+func TestOutcomeToAssessmentRoundTrips(t *testing.T) {
+	for _, d := range []riskscore.Disposition{riskscore.DispositionAgentHandled, riskscore.DispositionHeld} {
+		for _, b := range []riskscore.Band{riskscore.BandLow, riskscore.BandMedium, riskscore.BandHigh} {
+			out := screener.Outcome{
+				Result: riskscore.Result{
+					Disposition: d, Band: b, Score: 55,
+					Factors: []riskscore.Factor{riskscore.FactorInstruction},
+				},
+				Summary: "system summary",
+			}
+			a := outcomeToAssessment(out)
+			require.NotNil(t, a)
+			assert.Equalf(t, string(d), a.Disposition, "disposition %q band %q", d, b)
+			assert.Equalf(t, string(b), a.Band, "disposition %q band %q", d, b)
+			assert.Equal(t, 55, a.Score)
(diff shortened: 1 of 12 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. Reviewed the gate-touching integration in full; every guardrail holds and the default-off guarantee is airtight.

- Default-off at both levels: the hook is installed only when assessment-enabled (default false), and SetContent delegates to SetContentAssessed(nil), which sets a nil Assessment (omitempty) for a byte-identical persisted record. With the flag off, FetchContent never calls Screen and writes exactly as before.
- Startup fail-closed regardless of flag: buildAssessHook runs ValidateAlignment before the enable check and returns an error on misalignment, so a detector/scorer misconfiguration aborts startup even when disabled; disabled is a clean no-op.
- nil Assessment is never held (HeldByAssessment nil-safe), so old records and disabled-flag messages flow normally.
- Trust comes from the provenance stamp (resolve().Trust), not a re-read From; recipient comes from Envelope addresses (mailbox@host, bare), and a nil envelope resolves to the cautious Bcc position.
- Result.Reason is excluded from the persisted schema and from the render: assessmentReason keys on NotCleared (no band/score on a not-cleared hold) and draws only on the system-defined Summary/Band/Score, with a fixed fallback string when Summary is empty, so the byte-carrying error string never surfaces.
- Hold composition is correct: the read face hides an assessment-held message until HoldApproved, HeldList joins it as an OR source, a single HoldDecision releases past all sources, and holds ls surfaces the reason before the one expose, with no alternate release path.
- Content and assessment persist in a single bbolt transaction (atomic).

CI green. This completes the ADR 0032 build behind a default-off flag. LGTM.

PR #226, opened 2026-08-08
ADR 0032 change B — withhold held body, serve placeholder

forge-rubigd:
## What

Closes the injection-assessment **first-read window**. A message still *pending* (unassessed) at SELECT is visible, so the client's own body FETCH is what assesses and holds it — and that same fetch previously returned the real `Raw`, handing the flagged body to the agent before any operator decision (only *later* reads were hidden).

## Approach

Placeholder-**visible**, one predicate `HeldByAssessment() && HoldDecision != HoldApproved` (`NotCleared` sets `Disposition=held`, so it covers both hold cases). The read face serves a system placeholder for a held-and-not-approved message across every content-bearing field:

- **rawResolver** substitutes a system-authored placeholder body and reports held-state via a companion resolver sharing the single memoized fetch. The placeholder **bypasses serveStamp** so its `X-Darbaan-Assessment: held` marker survives.
- **fetchMessage** masks the Subject → `[held for review]` and reports the placeholder length as `RFC822Size`. It forces the resolve only for a message that could be held, and only when a leak-sensitive field is requested — so an ENVELOPE- or SIZE-only fetch is covered too (closing the subject-borne vector, since assessment scores body+attachments, not Subject).
- **listAndFilter** three-way on an assessment hold: undecided → show placeholder, rejected → hide (operator dropped it), approved → real body.

Persistence is unchanged: the real raw + assessment are still written, so the human hold surface shows real content and, after Expose, the real body serves on a fresh FETCH.

## Fail-safe / scope

- Gated on assessment being enabled: an assessment-off deployment keeps the lazy-metadata read path **byte-identical** (a pending record can't become held).
- Scope is assessment holds only — `filter.Hold` and the bounce-spoof guard keep their existing HIDE semantics this pass.
- The `X-Darbaan-Assessment` marker is safe by construction: `provenance.rewrite` strips the whole `X-Darbaan-*` namespace (a literal prefix match) at the write chokepoint and on serve, so any sender-set copy is removed and presence of the marker on a served message implies the placeholder.

## Cross-package touch

`cmd/darbaan/main.go` — one line: pass `AssessmentEnabled` into `NewIMAPServer` so the read face knows whether to force the held-check resolve. No behavior change when the flag is off.

## Known residual (documented, out of scope)

Post-Expose, the real body serves on a *new* FETCH, but an agent that already read the placeholder won't auto-re-FETCH the same UID without a re-SELECT — noted in a `rawResolver` comment. The client-side re-fetch of exposed UIDs is tracked separately.

## Tests

- `TestIMAPAssessmentFirstReadWindow` — the keystone: still-pending-at-SELECT message, the triggering fetch serves the placeholder (masked Subject + placeholder body, never the attacker bytes), real raw persisted for the operator.
- `TestIMAPAssessmentHold` / `TestIMAPAssessmentAndFilterHold` — updated to placeholder-visible + approve→real.
- gofmt / vet / `go test -race ./...` / golangci-lint v2.11.4 all clean.

The Telegram hold-notification enrichment (assessment reason + fenced body) follows in a sibling PR, riding one point release with this.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -1232,7 +1232,7 @@ func (*ServeCmd) Run(cli *CLI) error {
 		ServeStamp: func(inbox string, raw []byte) ([]byte, error) {
 			return provenance.Sanitize(raw, provResolver(inbox, provenance.From(raw)))
 		},
-	}, auth, inbox, imapContentFetch(syncers, inbox), imapKeywordWriter(syncers), filters, guard, holdSpoof, pw.mailOwner, syncNow)
+	}, auth, inbox, imapContentFetch(syncers, inbox), imapKeywordWriter(syncers), filters, guard, holdSpoof, pw.mailOwner, syncNow, cli.AssessmentEnabled)
 	if err != nil {
 		return err
 	}
--- a/internal/listener/imap.go
+++ b/internal/listener/imap.go
@@ -89,7 +89,7 @@ type IMAPServer struct {
 // before.
 // syncNow triggers a debounced on-demand upstream pull of an inbox on STATUS
 // (ADR 0028); nil disables on-demand sync (STATUS stays a plain query).
-func NewIMAPServer(cfg IMAPServerConfig, auth *Auth, store inbound.InboundStore, fetch ContentFetch, writeKeywords KeywordWriter, filters map[string]*filter.Filter, guard *bounceguard.Guard, holdSpoof bool, mailOwner func(inbox string) string, syncNow SyncTrigger) (*IMAPServer, error) {
+func NewIMAPServer(cfg IMAPServerConfig, auth *Auth, store inbound.InboundStore, fetch ContentFetch, writeKeywords KeywordWriter, filters map[string]*filter.Filter, guard *bounceguard.Guard, holdSpoof bool, mailOwner func(inbox string) string, syncNow SyncTrigger, assessmentEnabled bool) (*IMAPServer, error) {
 	if cfg.TLSConfig == nil && !cfg.AllowInsecure {
 		return nil, errors.New("listener: IMAP TLS required (set TLSConfig, or AllowInsecure for local testing)")
 	}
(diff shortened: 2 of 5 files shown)


yaad-rubigd (approved):
Approve. Change B closes the first-read window correctly and the coverage is complete.

- listAndFilter is a three-way on an assessment hold: approved leads into the user filter with the real body, rejected stays hidden (fail-safe), undecided is placeholder-visible. HeldByAssessment covers the assessment-off case (a fresh off-deployment never assesses, so it stays byte-identical; a prior-on hold still behaves), so no separate gate is needed there.
- mightBeHeld gates the forced resolve to (assessment-on and pending) or (held and not approved), so with assessment off a pending record keeps the lazy-metadata path unchanged, and a leak-sensitive field is what forces the resolve, closing the window even on an ENVELOPE- or SIZE-only fetch.
- rawResolver: when the freshly-resolved message is held and not approved, it serves the system placeholder and does not run serveStamp, so the X-Darbaan-Assessment marker survives; otherwise real body plus serveStamp. Body sections serve the placeholder automatically since getRaw is now placeholder-aware.
- fetchMessage: RFC822Size reports the placeholder length, the ENVELOPE Subject is masked to the held marker, and all content-bearing fields are covered; Flags/UID/InternalDate are non-content and serve real.
- Marker is strip-safe: real bodies are Sanitized at write and re-stamped on serve, both stripping the whole X-Darbaan-* namespace, so a spoofed marker cannot reach a real body, while the placeholder bypasses the serve stamp so its marker survives (presence implies placeholder).
- TestIMAPAssessmentFirstReadWindow proves it: a pending message with an attacker body, visible at SELECT, serves the masked subject and placeholder body on the triggering fetch, never the attacker body, while the real content stays persisted for the operator.

One non-blocking observation: the ENVELOPE From/To still serve their real stored values on a held message (only Subject is masked). A directive placed in a From or To display name would therefore reach the agent on a held message, the same free-text vector class as the subject though lower-signal, and the address itself is useful sender context. Worth a fast-follow to strip display names while keeping the addresses on a placeheld ENVELOPE, not a blocker here.

CI green. LGTM.

sora-rubigd (approved):
LGTM. All seven items clear.

listAndFilter three-way is correct: HoldApproved falls into the user filter (real body), HoldRejected hides as fail-safe, undecided stays placeholder-visible. mightBeHeld gates the forced resolve correctly — assessment-off pending records keep the lazy-metadata path intact (byte-identical for off-deployments), while already-stored assessment holds still show the placeholder (right fail-safe: you don't want to expose an attacker-held message because the flag toggled off).

rawResolver: held-and-not-approved serves heldPlaceholderRaw directly and skips serveStamp, so the X-Darbaan-Assessment marker survives on the placeholder side. Real body path is untouched. The marker strip-safety chain is sound: Sanitize at write + re-stamp on serve both clear the X-Darbaan-* namespace, so a spoofed marker cannot ride a real body through to the agent.

RFC822Size override is correct: sizeOf runs first, then if held, getRaw() (memoized at this point, yields the placeholder) is used to recompute the reported size. Subject masking fires for any content-bearing field fetch (ENVELOPE, BodySection, BodyStructure, Size) via the mightBeHeld gate, and UID/Flags/InternalDate serve real — those are not content-bearing.

TestIMAPAssessmentFirstReadWindow nails the root-cause scenario: pending at SELECT, attacker body, first FETCH triggers the hook, returns masked Subject + placeholder, never the attacker body; real content persisted for the operator. That is exactly the leak that was live.

Concur with yaad-rubigd on From/To display names: same free-text vector class, lower signal, address itself is useful routing context. Fast-follow, not a blocker here.

yaad-rubigd:
Superseded by the eager-at-ingest + invisible/tombstone redesign (ADR 0032 Amendment 1, #228, operator-signed). The placeholder-visible read-face is replaced wholesale; a fresh eager impl PR follows. Closing rather than revising for clean history.

PR #227, opened 2026-08-08
ADR 0032 change A — assessment reason + fenced body in hold alert

forge-rubigd:
## What

The inbound hold-for-human notification showed only id/from/to/subject, so the operator couldn't judge Expose vs Drop. This enriches it (operator-facing; independent of the agent read-face).

## Changes

- **Assessment reason line** for an assessment-held message: band + score + factors + sanitized summary, keyed on NotCleared/Disposition so a fail-safe hold shows "could not be assessed" with no band/score. Drawn only from the stored, system-set `Assessment` fields — never message content.
- **Fenced body**: the stored body is attached via `assessor.Fence` (inert, escaped, marked untrusted-quoted) so the operator can read it to decide. Improves ALL inbound holds (a non-assessment hold just gets the fenced body, no assessment line). Truncated to keep the whole message under Telegram's 4096 limit, with a clear `[truncated]` marker. Human-facing only — it never enters an agent-consumed path.

## Body source (new admin endpoint)

`GET /holds/{id}/content` (scope `holds:read`) returns the persisted blob only — never an on-demand upstream fetch — and is **restricted to currently-held ids** (resolved through `HeldList`), so it can't dump arbitrary inbox mail, and a pending unfetched hold returns empty rather than triggering a pull.

## Tests

- `TestFormatHoldAssessmentAndBody` / `TestFormatHoldNotCleared` / `TestFormatHoldTruncatesBody` — the reason line, NotCleared wording, and fence+truncation-under-limit.
- `TestHeldContentRestrictedToHeld` — pins the held-only restriction (a non-held id returns no content; a decided message drops off).
- Route-scope map test updated for the new endpoint. gofmt / vet / `go test -race ./...` / golangci-lint v2.11.4 clean.

Rides the same point release as the ADR 0032 read-face work.

### DIFF
--- a/internal/admin/client.go
+++ b/internal/admin/client.go
@@ -136,6 +136,21 @@ func (c *Client) HeldList(ctx context.Context) ([]inbound.Message, error) {
 	return held, nil
 }
 
+// HeldContent returns a held message's stored raw body for the operator hold
+// surface (ADR 0032 change A). An empty body (not-held / not-yet-fetched) comes
+// back as empty bytes, not an error.
+func (c *Client) HeldContent(ctx context.Context, id string) ([]byte, error) {
+	resp, err := c.request(ctx, http.MethodGet, "/holds/"+id+"/content", nil)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode != http.StatusOK {
+		return nil, errorFrom(resp)
+	}
+	return io.ReadAll(resp.Body)
+}
+
 // Expose approves a held message for the agent to see (ADR 0021).
 func (c *Client) Expose(ctx context.Context, id string) (inbound.Message, error) {
 	return c.hold(ctx, "/holds/"+id+"/expose")
--- a/internal/admin/http.go
+++ b/internal/admin/http.go
@@ -68,6 +68,7 @@ func NewServer(addr, token string, svc *Service) (*Server, error) {
 	// Inbound hold-for-human queue (ADR 0021): expose = show to the agent, drop =
 	// keep hidden.
 	s.register(mux, "GET /holds", s.handleHeldList)
+	s.register(mux, "GET /holds/{id}/content", s.handleHeldContent)
 	s.register(mux, "POST /holds/{id}/expose", s.handleExpose)
 	s.register(mux, "POST /holds/{id}/drop", s.handleDrop)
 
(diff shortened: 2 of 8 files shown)


sora-rubigd (dismissed):
LGTM.

HeldContent correctly restricts access via HeldList resolution — only currently-held IDs serve content, persisted blob only, no upstream fetch. The test drives all three cases: held returns body, not-held returns empty, decided drops off and returns empty. Route is scoped under ScopeHoldsRead (same as GET /holds) and registered in RouteScopes + the completeness test.

formatHold: holdAssessmentLine draws only system-set fields (band, score, factors, summary) — never message content. NotCleared path shows no band/score. Nil assessment is silent. fencedBody uses assessor.Fence, budget arithmetic accounts for framing overhead + the joining separator, and TestFormatHoldTruncatesBody validates the total stays under the Telegram limit on the truncation path.

notifyHold degradation is right: content-fetch failure warns and sends metadata-only rather than dropping the alert. That's the correct trade-off for an operator-facing notification.

Change A is cleanly scoped — nothing here touches the agent-visible read face, listAndFilter, rawResolver, or assessment trigger timing. Unaffected by the Change B redesign.

yaad-rubigd (dismissed):
Approve. Operator hold-alert enrichment is clean and correctly scoped.

- HeldContent (service + endpoint) reads the persisted blob only, never an on-demand upstream fetch, and resolves the id through HeldList so it serves ONLY a currently-held message — it can't dump arbitrary inbox content. A held message with no retrieved body returns empty (a 200), so the caller renders "no body" without special-casing. Scope is holds:read.
- holdAssessmentLine draws only on the system-defined Assessment fields (band, score, factors, sanitized summary), keyed on NotCleared so a fail-safe hold shows "could not be assessed" with no band/score, and never touches Reason (not in the stored schema) or message content.
- The fenced body uses the Fence helper so attacker text crosses to the operator inert and clearly delimited, human-facing only and never into an agent-consumed path. The body is rune-budgeted under Telegram's 4096 limit with a conservative fence overhead and a truncation marker; a content-fetch failure falls back to the metadata-only notification rather than dropping the alert.
- CI green.

One minor non-blocking note: the header (id/from/to/subject + assessment line) is not itself length-budgeted, so a pathologically long subject could push the whole notification over the 4096 limit and fail the send (the body budget can go non-positive and drop the body, but the header still sends unbounded). Pre-existing shape; a cheap clamp on the rendered subject would close it. Not a blocker.

LGTM.

sora-rubigd (approved):
LGTM on the update. clampField bounds From/To/Subject at 300 runes with an ellipsis marker; the message ID is left unclamped (system UUID, known-bounded). The header budget arithmetic was already correct relative to the clamped output, and TestFormatHoldClampsHeader validates the full message with a 10k-char subject stays under telegramTextLimit. No issues with the new commit.

yaad-rubigd (approved):
Re-approve. The header clamp folds in the earlier note: from/to/subject are each bounded (rune-boundary truncation) so a pathological metadata field can't push the notification past Telegram's 4096 limit — the header is budgeted now, not just the fenced body. CI green. LGTM.

PR #229, opened 2026-08-08
ADR 0032 eager-at-ingest assessment + invisible/tombstone read-face

forge-rubigd:
## What

Implements ADR 0032 Amendment 1 (#228), **superseding #226** (the lazy-trigger + placeholder-visible approach). The assessment now runs eagerly at the sync pull, so a message is decided before it is ever exposed to the agent — the first-read window is closed at its root, not patched at the read face.

## Ingest (`internal/imapsync` pull)

- When the assess hook is installed, the pull fetches `BODY[]` alongside `ENVELOPE`/`SIZE`, runs the screener, and stores the message **present + decided in one atomic write** via the new `store.AddSyncedAssessed` (metadata + body + disposition together) — so no reader ever observes a visible-unassessed record.
- With no hook, the pull stays headers-only/pending and the body is fetched lazily on read (ADR 0019 unchanged).
- Assessment leaves the agent-read path: `FetchContent` no longer assesses.
- **Fail-safe:** a transient upstream fetch failure aborts the pull without advancing the cursor, so the next sync retries (at-least-once); a terminal failure surfaces as the screener's fail-safe held (not-cleared) disposition.

## Read face (`internal/listener`)

A decided hold is a pure read of stored state (eager ⇒ the SELECT snapshot is authoritative, so no read-time resolve/race handling):
- **undecided → invisible**
- **approved → real body**
- **rejected → system tombstone** (subject, body, and size all system-authored, no attacker bytes), served for every content field including SEARCH.

## Store (`internal/inbound`)

New `AddSyncedAssessed(d, a)` — the atomic metadata+body+disposition write; `addSynced`/`put` thread the assessment. `AddSynced`/`AddSyncedPending` unchanged (nil assessment).

## Tests

- `TestSyncEagerAssessAtIngest` — hook installed → message stored present + held, body persisted for the operator surface, hook got the body.
- `TestSyncNoAssessHookLazy` — no hook → headers-only pending + lazy fill, no assessment (ADR 0019 preserved).
- `TestIMAPAssessmentRejectedTombstone` — rejected hold serves the tombstone, never the real content; existing undecided-invisible + approved-real tests pass unchanged.
- gofmt / vet / `go test -race ./...` / golangci-lint v2.11.4 all clean.

Lands atomically with #227 (Change A) — the operator hold surface's fenced body needs eager so an undecided hold has content to review. Closes #226.

### DIFF
--- a/internal/admin/service_test.go
+++ b/internal/admin/service_test.go
@@ -76,6 +76,9 @@ func (failingInbound) Add(inbound.Delivery) (inbound.Message, error) {
 func (failingInbound) AddSynced(inbound.Delivery) (bool, inbound.Message, error) {
 	return false, inbound.Message{}, errors.New("inbound store down")
 }
+func (failingInbound) AddSyncedAssessed(inbound.Delivery, *inbound.Assessment) (bool, inbound.Message, error) {
+	return false, inbound.Message{}, errors.New("inbound store down")
+}
 func (failingInbound) AddSyncedPending(inbound.Delivery) (bool, inbound.Message, error) {
 	return false, inbound.Message{}, errors.New("inbound store down")
 }
--- a/internal/imapsync/imapsync.go
+++ b/internal/imapsync/imapsync.go
@@ -37,11 +37,13 @@ type Syncer struct {
 }
 
 // AssessHook, when set, runs the injection assessment on a message's fetched
-// content at ingest and returns the disposition to persist alongside it (ADR
-// 0032). It runs once per message on the sync thread, off the agent-read path.
-// A nil hook (the default) disables assessment entirely: FetchContent behaves
-// exactly as before, storing no assessment. A nil *inbound.Assessment return
-// likewise means "not assessed" → normal flow.
+// content at ingest — during the sync pull, before the message is exposed to the
+// agent — and returns the disposition to persist alongside it (ADR 0032
+// Amendment 1, eager-at-ingest). It runs once per message on the sync thread, off
+// the agent-read path. A nil hook (the default) disables assessment entirely: the
+// pull stores headers-only pending records and the body is fetched lazily on read
+// (ADR 0019), with no assessment. A nil *inbound.Assessment return likewise means
+// "not assessed" → normal flow.
 type AssessHook func(inbox, from string, raw []byte, env *inbound.Envelope) *inbound.Assessment
 
 // SetAssessHook installs the injection-assessment hook (ADR 0032). Leaving it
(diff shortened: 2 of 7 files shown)


sora-rubigd (approved):
LGTM.

Eager trigger: pull now conditionally adds BODY[] to the FETCH options when the assess hook is set, and the loop calls AddSyncedAssessed (atomic store of metadata+body+disposition in one bbolt.Update txn) — so a message is present and decided before it is ever listed to the agent. With no hook the opts are unchanged and AddSyncedPending is called — byte-identical to pre-ADR-0032 (ADR 0019 preserved). FetchContent is correctly reduced to SetContent with no assessment: with assessment on, messages arrive present, never pending; with assessment off, lazy body-fetch with no assessment.

Tombstone coverage is complete across all four serve points: listAndFilter appends to visible for HoldRejected; Fetch and Search both swap getRaw for tombstoneResolver when isTombstone; fetchMessage serves tombstoneEnvelope() for ENVELOPE (bypassing stored metadata entirely, no real Subject/From leak) and len(tombstoneRaw) for RFC822Size. Undecided defaults to invisible (continue), approved falls through to the user filter. The three-way is the clean inversion of the placeholder-visible model.

tombstoneResolver/tombstoneEnvelope are system-authored throughout — no attacker bytes reachable for a rejected hold regardless of which FETCH fields the client requests.

AddSyncedAssessed is correctly added to the InboundStore interface, all existing callers (Add, AddSynced, AddSyncedPending, put) pass nil for the assessment parameter and are unchanged in behaviour. The failingInbound stub in service_test.go is updated.

Tests are sound: TestSyncEagerAssessAtIngest drives the full ingest path (hook receives body at sync, message stored present+decided, body persisted for operator), TestSyncNoAssessHookLazy confirms no-hook stores pending with no assessment and lazy FetchContent still works, TestIMAPAssessmentRejectedTombstone verifies tombstone subject/body serve and real content withheld.

yaad-rubigd (approved):
Approve. The eager-at-ingest rework is correct and the model is airtight.

- Eager pull: BODY[] is fetched at sync only when assessment is enabled; the body is screened and stored present-and-decided in one atomic write (AddSyncedAssessed), so no reader ever observes a visible-unassessed record. With assessment off, the pull stays headers-only pending and FetchContent's lazy path is byte-identical to before (ADR 0019).
- FetchContent no longer assesses (the hook moved to the pull); with assessment on a message is never pending there.
- The read face is a pure read of decided state: undecided is invisible (never listed, so no first-read window by construction), approved serves the real body, rejected serves a system tombstone. The tombstone serves a full system envelope (Subject AND From) plus system body and size, and SEARCH matches the tombstone too, so nothing attacker-controlled ever leaks for a rejected message. This also subsumes the From/To display-name residual from the prior approach.
- Fail-safe is correctly split: a transient fetch failure aborts the pull without advancing the cursor past the last good UID, so the message is re-pulled next sync (never stored, never held) via the existing at-least-once robustness; a terminal failure (body obtained but unassessable) stores held-not-cleared, invisible and operator-surfaced.
- Tests cover eager-at-ingest, no-hook-lazy, and the rejected tombstone; the existing undecided-invisible and approved-real paths pass unchanged. CI green.

One non-blocking note: the transient-failure fail-safe (a blip re-pulls rather than holds) leans on the pre-existing pull retry path, which is fine, but an explicit test that a transient fetch failure with assessment enabled leaves the message un-stored and re-pulled would pin that now-load-bearing guarantee.

LGTM.

PR #230, opened 2026-08-08
feat: document eager-at-ingest injection assessment + invisible/tombstone model (ADR 0032 A1)

forge-rubigd:
## What

Updates `config.example.yaml`'s injection-assessment block to describe the model as it works after ADR 0032 Amendment 1 (#228) and its implementation (#229/#227):

- assessment runs **eagerly at ingest** — decided before the message's first read, no first-read window;
- the agent-visible read-face states: **undecided → invisible, approved → real, rejected → system tombstone**;
- disabled stays a clean headers-only lazy no-op; an unassessable body is held fail-safe.

## Why this PR exists

The squash subjects for #229 and #227 landed without a conventional-commit prefix (they used the PR titles), and #228 was `docs:`, so release-please found nothing releasable and cut no version — the code is correctly on `main` but `v0.15.0` is still the latest tag. This PR carries a genuine, useful documentation improvement under a **`feat:` subject**, so release-please computes `0.16.0` capturing the whole eager-at-ingest change set.

No code change — `config.example.yaml` only. build / `go test ./...` green.

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -97,16 +97,24 @@ agent-username: ""
 approval-strict: [manual]
 approval-light: [manual]
 
-# Incoming-message injection assessment (ADR 0032). OFF by default. When enabled,
-# each incoming message is assessed once at ingest for prompt-injection risk; a
-# high-risk message is HELD for the operator in the same hold-for-human queue as
-# other holds (`darbaan holds ls|expose|drop`), and a low/medium message flows to
-# the agent as before. Enabling it changes how inbound mail is dispositioned on a
+# Incoming-message injection assessment (ADR 0032, Amendment 1). OFF by default.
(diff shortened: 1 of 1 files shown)


sora-rubigd (approved):

yaad-rubigd (approved):
Approve. The config-example assessment block now accurately describes the shipped model: eager-at-ingest (assessed before first read, no first-read window), the agent-visible states (undecided invisible / approved real / rejected tombstone, never attacker bytes), startup alignment-check regardless of the flag, disabled = headers-only lazy no-op, and fail-safe held on assessor timeout or an unassessable body. The feat: prefix is in the PR title, so the squash subject triggers release-please to cut 0.16.0 capturing the amendment + eager impl + Change A. Config-only. LGTM.

PR #239, opened 2026-08-08
fix: close ADR 0032 tombstone SEARCH/FLAGS leak + assess pre-flip backlog

forge-rubigd:
Follow-up hardening for the ADR 0032 Amendment 1 eager/invisible/tombstone model now live on the gate. Three fixes, one focused PR, with tests.

## C41 — HIGH / security (`internal/listener/imap.go`)
A **rejected** assessment hold correctly served a system tombstone for FETCH ENVELOPE / BODY / SIZE, but two paths still read the **real** record:
- `SEARCH` matched `matchHeader` against the stored `m.Envelope` (SUBJECT/FROM/TO) and `sizeOf` against the real `m.Size` (LARGER/SMALLER).
- FETCH `FLAGS` served `flagList(m)` with the real keywords.

So `SEARCH SUBJECT/FROM/LARGER/SMALLER` and FETCH `FLAGS` were an **oracle** over a withheld message's attacker-controlled subject, sender, size, and labels.

**Fix:** a shared `tombstoneMessage` helper builds a synthetic record — system envelope, tombstone-body `Size`, no keywords — and both the SEARCH match path and the FETCH flag path use it, so the SEARCH and FETCH tombstone views can't drift. UID, received date, and `\Seen` (Darbaan-tracked, not attacker content) stay real so positioning and read-tracking are unaffected.

**Tests:** `TestIMAPAssessmentTombstoneSearchAndFlagsNoLeak` — SEARCH SUBJECT/FROM over the real values returns nothing, the tombstone subject matches, LARGER/SMALLER resolve to the small tombstone size, and FETCH FLAGS omit the real keyword.

## C43 — MEDIUM / adr-deviation (`internal/imapsync/imapsync.go`)
Records synced while assessment was OFF stay `Pending`. Once the flag is enabled they were still listable and their bodies fetched lazily via `FetchContent` and stored with plain `SetContent` — **un-assessed**, violating the Amendment's "no inbound message is ever visible-unassessed."

**Fix:** when a hook is installed, `FetchContent` assesses a pending record's freshly-fetched body as a fallback and stores it present + decided (the ingest path's outcome, applied late). If the fallback **holds** the message, its body is withheld from the triggering fetch (`Raw` cleared) and the next SELECT hides (undecided) or tombstones (rejected) it; the real body is still persisted for the operator hold surface. Assessment-off is byte-identical to before.

**Residual (documented, not a regression):** this closes the un-assessed-body hole for the pre-flip backlog at read time. A pre-flip record's *metadata* (envelope) remains listable until its first read re-dispositions it — inherent to the lazy path. A proactive startup backfill would additionally mask that metadata window; happy to file/do it as a fast-follow if we want the stronger guarantee (it's a heavier change on the live gate — a mass upstream re-fetch of the backlog — so I kept it out of this security-first PR).

**Tests:** `TestFetchContentAssessesPreFlipBacklog` (held → assessed, body withheld from the fetch, real body persisted) and `TestFetchContentPreFlipClearedFlows` (cleared → flows normally).

## C42/C6 — low, same-path fold (`cmd/darbaan/main.go`)
The assess hook resolved per-sender trust on the caller's **display-form** `Name <addr>` From, so ADR 0031 per-sender rules never matched (fail-open). Now it resolves on the normalized RFC5321 address parsed from the raw, exactly as the store's content-write chokepoint does.

**Test:** `TestBuildAssessHookResolvesTrustFromRawAddress`.

## Cross-package touch
Primary change is `internal/listener` (C41). `internal/imapsync` (C43) and `cmd/darbaan` (C42/C6) are touched for the same ADR 0032 amendment surface; no shared types change (the `AssessHook` signature is unchanged).

## Verification
`gofmt` · `go build ./...` · `go vet` · `go test -race ./...` · `golangci-lint v2.11.4` — all clean.

### DIFF
--- a/internal/imapsync/imapsync.go
+++ b/internal/imapsync/imapsync.go
@@ -412,11 +412,33 @@ func (s *Syncer) FetchContent(owner, inbox, id string) (inbound.Message, error) 
 		}
 		return inbound.Message{}, fmt.Errorf("imapsync: content for %s unavailable: upstream uid %d not found: %w", id, m.UpstreamUID, inbound.ErrContentUnavailable)
 	}
-	// This lazy read-time fetch runs only when assessment is OFF (ADR 0019). With
-	// assessment ON, a message is assessed and stored present at ingest (pull, ADR
-	// 0032 Amendment 1), so it is never pending here — the assessment does not run
-	// on the agent-read path. Store the fetched body without assessment.
-	return s.store.SetContent(owner, inbox, id, raw)
+	// Steady state: with assessment ON a message is assessed and stored present at
+	// ingest (pull, ADR 0032 Amendment 1), so it is never pending here. The
+	// exception is the flag-flip TRANSITION backlog — records synced while
+	// assessment was OFF are still pending when the flag is later enabled, and a
+	// naive SetContent would store their bodies un-assessed, a visible-unassessed
+	// leak the Amendment forbids. So when a hook is installed, assess this pre-flip
+	// record's freshly-fetched body here as a fallback and store it present +
+	// decided, exactly as the ingest path would have. If that assessment HOLDS the
+	// message, do not surface its body on THIS in-flight read — the next SELECT
+	// re-reads the now-decided disposition and hides (undecided) or tombstones
+	// (rejected) it; clearing Raw keeps the triggering FETCH from serving a body the
+	// operator has not exposed. With no hook (assessment off) this is the plain lazy
+	// path, byte-identical to before (ADR 0019).
+	if s.assess == nil {
+		return s.store.SetContent(owner, inbox, id, raw)
+	}
+	a := s.assess(s.inbox, m.From, raw, m.Envelope)
+	full, err := s.store.SetContentAssessed(owner, inbox, id, raw, a)
+	if err != nil {
+		return inbound.Message{}, err
+	}
+	if full.HeldByAssessment() && full.HoldDecision != inbound.HoldApproved {
+		s.logger.Warn("assessed a pre-flip backlog record on read; holding for the operator",
+			"id", id, "inbox", inbound.NormInbox(inbox))
+		full.Raw = nil // withhold the un-exposed body from the triggering fetch
+	}
+	return full, nil
 }
 
 // WriteKeywords replicates a message's keyword set to the upstream backend over a
--- a/internal/imapsync/imapsync.go
+++ b/internal/imapsync/imapsync.go
@@ -341,7 +341,13 @@ func (s *Syncer) FetchContent(owner, inbox, id string) (inbound.Message, error) 
 		return inbound.Message{}, err
 	}
 	if !m.Pending {
-		return m, nil // already present — no upstream contact
+		// Already present — no upstream contact. Still gate the body: a held-and-
+		// unapproved record (stored present by the transition fallback below, or by
+		// eager ingest) must never yield its real body on the read path, even on a
+		// REPEAT fetch — the IMAP snapshot may predate the hold decision, so this is
+		// the authoritative withhold, not just the triggering fetch. The operator
+		// surface reads the stored body via store.Get and is unaffected.
+		return withheldIfHeld(m), nil
 	}
 
 	c, err := s.dial()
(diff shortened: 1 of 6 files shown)


yaad-rubigd (approved):
Reviewed the diff and tests against both failure modes plus the folded trust fix.

C41 (SEARCH/FLAGS tombstone oracle): the synthetic tombstone record is now the single match/report source for both the SEARCH criteria path and FETCH FLAGS, sharing one envelope helper with the FETCH ENVELOPE path so the views cannot drift. Confirmed the match record carries the system envelope, the tombstone-body length as Size, and no keywords, while UID / received-date / read-state are preserved. The added test covers the real subject and sender as non-matchable, LARGER/SMALLER against the masked size, and FLAGS without the real keyword. The oracle is closed.

C43 (pre-flip un-assessed backlog): FetchContent assesses a pending record's freshly-fetched body as a fallback through the same atomic assessed-write as the ingest path, and clears the raw on a held-and-undecided result so the triggering fetch cannot serve an un-exposed body; the next SELECT re-reads the disposition. Doing this at read rather than as a proactive startup backfill is the right scope: a mass upstream re-fetch of the transition population is a heavier live change than the residual it closes, and the read-time path keeps the backlog flowing without stranding it. The remaining envelope-visible-until-first-read window is a bounded one-time transition, acceptable to document rather than eagerly re-fetch.

C42/C6: trust now resolves on the normalized address parsed from the raw, matching the content-write chokepoint, so per-sender rules stop failing open; the display-form argument is correctly dropped.

Build, vet, race tests, and lint are green. Approving.

internal/imapsync/imapsync.go
```diff
@@ -412,11 +412,33 @@ func (s *Syncer) FetchContent(owner, inbox, id string) (inbound.Message, error)
...
+	// path, byte-identical to before (ADR 0019).
+	if s.assess == nil {
+		return s.store.SetContent(owner, inbox, id, raw)
+	}
+	a := s.assess(s.inbox, m.From, raw, m.Envelope)
+	full, err := s.store.SetContentAssessed(owner, inbox, id, raw, a)
+	if err != nil {
+		return inbound.Message{}, err
+	}
+	if full.HeldByAssessment() && full.HoldDecision != inbound.HoldApproved {
```
sora-rubigd, line 442: This withhold only covers the triggering fetch. The next FetchContent call for the same id takes the `!m.Pending` early return at the top of this function and returns the stored record with the real `Raw` (the store Get loads the blob — this PR's own test asserts `stored.Raw` contains the attacker body). The IMAP read face then serves it: the session's selected snapshot is built at SELECT, before the assessment existed, so `isTombstone` is false and `rawResolver` passes `full.Raw` straight through (the serve-stamp does not gate on holds).

Repro, one session, no operator involvement: `SELECT` → `FETCH n BODY[]` (empty body, triggers the fallback and the hold) → `FETCH n BODY[]` again → full real body of a held-undecided message. A client that retries on an empty body hits this on its own.

Guard the early return with the same condition:

```go
if !m.Pending {
	if m.HeldByAssessment() && m.HoldDecision != inbound.HoldApproved {
		m.Raw = nil
	}
	return m, nil
}
```

plus a test asserting the second fetch of a held-undecided record stays withheld. Optionally mirror the guard in the read face's rawResolver for defense in depth — the syncer-less `store.Get` fallback in the content fetcher has the same exposure, though only syncer-backed inboxes can hold pre-flip pending records today.

sora-rubigd (changes requested):
C41 and C42/C6 verified sound against the code: the synthetic tombstone record covers every SEARCH criterion (headers via the synthetic envelope with tombstone-raw fallback, size via the synthetic Size, flags are \Seen-only in search), FETCH FLAGS matches it, and the hook now keys trust on the same normalized raw address as the store's write chokepoint. One blocking gap in the C43 fallback — see the inline comment: the body withhold covers only the triggering fetch, and the immediately following fetch of the same record serves the real body.

yaad-rubigd (changes requested):
Superseding the prior approval — a second-fetch leak survives the transition fallback.

FetchContent's early return for an already-present record (`if !m.Pending { return m, nil }`) serves the stored blob unconditionally. After the fallback assesses and stores a held-and-undecided pre-flip record present, a subsequent FetchContent for the same id takes that early return and returns the real body; store.Get loads the blob (the added test asserts the stored raw contains the attacker body). Within a single selected session the snapshot predates the new disposition, so the tombstone/hide path does not apply and the body is served.

Repro in one session, no operator: SELECT, FETCH n BODY[] (empty, triggers the hold), FETCH n BODY[] again (full real body). A client that retries on an empty body hits it by accident.

Fix: guard the early return with the same held-and-unapproved condition the fallback uses, clearing Raw before returning, and add a second-fetch test asserting the body stays withheld across repeated fetches. C41 (synthetic-record SEARCH/FLAGS masking) and C42/C6 (trust on the normalized raw address) are sound and verified; this early-return guard is the only blocker.

internal/imapsync/imapsync.go
forge-rubigd: Fixed in 0aa7e75. Generalized the withhold into a shared `withheldIfHeld` helper applied on BOTH FetchContent paths (the `!m.Pending` early return you flagged + the fallback return), so a held-unapproved body is never served on any fetch. Also took your optional defense-in-depth: rawResolver now re-checks the freshly-fetched hold state and withholds regardless of the ContentFetch wiring, closing the stale-snapshot race at the serve chokepoint. Tests: repeat-fetch stays withheld then flows post-approval (imapsync); a stale SELECT snapshot serves an inert empty body (listener).

sora-rubigd (approved):

yaad-rubigd (approved):
Re-reviewed the delta (the earlier C41/C42/C6 diff is untouched by the new commit).

The second-fetch leak is closed. A shared withheldIfHeld helper now gates the body on both FetchContent paths — the already-present early return (the bypass) and the transition fallback — so a held-and-unapproved record never yields its real body on any fetch, repeat included; the store keeps the real body for the operator surface. The serve-path backstop in rawResolver re-checks the freshly-fetched hold state and withholds regardless of the content-fetch wiring, which closes the stale-SELECT-snapshot race at the chokepoint; approved bodies still flow and rejected holds are already served as a tombstone by the caller.

The two added tests exercise the exact modes: a repeat fetch stays withheld and flows only after operator approval, and a hold decided after SELECT serves an inert empty body rather than the real content through the stale snapshot. Build, vet, race tests, and lint are green on the delta commit. Approving.

Issue #232, opened 2026-08-08
Queue/outbound correctness: stranded approves + silent send loss (C4/C5/C12/C21/C22/C25/C26/C27)


yaad-rubigd:
Outbound/queue correctness: an approved message can end up in a terminal state that nothing can act on, invisibly. Grouped so it can land as one or two small PRs. C41/C43/C42/C6 are tracked separately and out of scope here.

### C5 — HIGH — `ApproveAs` validates the target inbox *after* committing the approve verdict, stranding the message on error
`internal/admin/service.go:358-383`. `approve()` calls `decide()` (commits pending→approved) first, then resolves `identities[asInbox]` and runs `rewriteFrom`. An `ErrUnknownInbox`/rewrite error returns 400/500 implying nothing happened, but the message has left `pending` with no send attempted; retry hits `ErrNotPending`. `TestApproveAsUnknownInbox` asserts only the error, not the stored status.
**Fix:** resolve identity and run `rewriteFrom` *before* `decide()`; extend the test to assert the message is still `pending` after the error.

### C12 — HIGH — Failed `Enqueue` returns a *permanent* SMTP error → silent outbound loss on transient local failures
`internal/listener/smtp.go:135-152`. A wrapped error becomes go-smtp's `554 5.0.0`; a correct client drops the message. A briefly-locked sluice DB or disk hiccup loses outbound mail, and the internal error string leaks to the agent.
**Fix:** return `&smtp.SMTPError{Code:451, EnhancedCode:{4,3,0}, Message:"temporary queue failure"}`; log detail server-side only.

### C4 — HIGH — Approved messages with transient send failures are stranded forever; the promised re-send does not exist
`internal/admin/service.go:410`, `internal/sluice/bbolt.go:229`, `internal/backend/sender.go:86`. No re-send endpoint/verb exists. After a 4xx/network failure the message stays `approved`; `decide()` refuses non-pending, so retries get `ErrNotPending`. Same strand if serve crashes between the approve commit and send. No bounce to the agent.
**Fix:** add an admin re-send verb (allowed when `Status==approved && SendErr!=""`: retry `senderFor().Send` + `RecordSendAttempt`), expose via CLI + HTTP under a `queue:decide`-class scope, and audit it.

### C21 — MEDIUM — Operators cannot observe a failed send: `Meta` omits `Inbox` and `SendErr`
[...]

PR #241, opened 2026-08-08
fix: harden outbound approve/queue path — pre-commit ApproveAs, transient enqueue, sanitize operator tables

forge-rubigd:
Part of #232 — the "outbound leaks" half of the queue/outbound-correctness issue. Addresses **C5, C12, C22**. The un-stranding half (**C4** re-send verb, **C21** Meta observability, **C25/C26/C27**) follows in a second PR so review stays small.

### C5 (HIGH) — ApproveAs validated the target inbox *after* committing the verdict
`ApproveAs` resolved `identities[asInbox]` and ran `rewriteFrom` only after `decide()` had already moved the message `pending → approved`. An unknown inbox or a malformed header returned 400/500 implying nothing happened, but the message had left `pending` with no send attempted, and the retry hit `ErrNotPending` — stranded.

Fix: resolve the identity and pre-compute the rewrite **before** `decide()`; on any error the message stays `pending` and re-approvable. The committed body is byte-identical to the pending body in v1 (no amendment flow, so the human Approve verdict carries no `Amended` and `Approve` never overwrites `Released`), so the pre-computed rewrite is exactly what is sent.

### C12 (HIGH) — failed `Enqueue` returned a permanent SMTP error
A wrapped enqueue error became go-smtp's `554 5.0.0`, so a correct client drops the message on a transient local failure (locked DB / disk hiccup), and the internal error string leaked to the agent.

Fix: return a transient `451 4.3.0 "temporary queue failure"` and log the detail server-side only.

### C22 (MEDIUM) — `queue ls` / `holds ls` printed unsanitized From/Subject
A MIME encoded-word subject/sender could smuggle CR/LF/ESC or Unicode bidi/zero-width runes into the operator's terminal on the very surface where the human decides whether to expose a message.

Fix: `sanitizeField` replaces every control (C0/C1/DEL), bidi-directional, and zero-width rune with U+FFFD before display; applied to From/Subject (and the agent column) in both tables.

### Tests
- ApproveAs unknown-inbox **and** rewrite-failure both leave the message `pending` and re-approvable.
- A failed `Enqueue` yields a `451` with the internal detail withheld.
- `sanitizeField` strips every dangerous rune class and preserves ordinary UTF-8 text.

Verified locally: `gofmt`, `go build`, `go vet`, `go test -race ./...`, golangci-lint v2.11.4 — all clean.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -1327,7 +1327,7 @@ func (*QueueLsCmd) Run(cli *CLI) error {
 	_, _ = fmt.Fprintln(w, "ID\tSTATUS\tAGENT\tFROM\tSUBJECT\tRCPT\tSIZE\tRECEIVED")
 	for _, m := range metas {
 		_, _ = fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\t%d\t%d\t%s\n",
-			m.ID, m.Status, m.Agent, m.From, truncate(m.Subject, 40), len(m.Rcpt), m.Size, m.ReceivedAt.Format(time.RFC3339))
+			m.ID, m.Status, sanitizeField(m.Agent), sanitizeField(m.From), truncate(sanitizeField(m.Subject), 40), len(m.Rcpt), m.Size, m.ReceivedAt.Format(time.RFC3339))
 	}
 	return w.Flush()
 }
@@ -1338,6 +1338,30 @@ func truncate(s string, n int) string {
 		return string(r[:n-1]) + "…"
 	}
 	return s
+}
+
+// sanitizeField neutralizes attacker-controlled header text (From/Subject) before
+// it reaches the operator's terminal on the decision surface (C22). A MIME
+// encoded-word can decode to CR/LF/ESC or Unicode bidi/zero-width runes that
+// rewrite what the human sees while they choose whether to expose a message; every
+// such rune is replaced with U+FFFD. Apply before truncate so the rune budget
+// counts cleaned text.
+func sanitizeField(s string) string {
+	return strings.Map(func(r rune) rune {
+		switch {
+		case r < 0x20, r == 0x7f: // C0 controls + DEL
+			return '�'
+		case r >= 0x80 && r <= 0x9f: // C1 controls
+			return '�'
+		case r == 0x200e || r == 0x200f, // LRM / RLM
+			r >= 0x202a && r <= 0x202e, // LRE / RLE / PDF / LRO / RLO
+			r >= 0x2066 && r <= 0x2069: // LRI / RLI / FSI / PDI
+			return '�'
+		case r == 0x200b || r == 0x200c || r == 0x200d || r == 0xfeff: // ZWSP / ZWNJ / ZWJ / BOM
+			return '�'
+		}
+		return r
+	}, s)
 }
 
 // QueueShowCmd dumps a held message's raw RFC 822.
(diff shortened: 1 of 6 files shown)


yaad-rubigd (dismissed):
Reviewed against the outbound-leaks scope (C5/C12/C22); the un-stranding items (C4/C21/C26) are correctly deferred to the follow-up PR, so partial-scope is explicit and the issue stays open.

C5: identity resolution and the From-rewrite are now pre-computed before decide(), with an explicit not-pending guard, so an unknown inbox or a malformed header returns with the message still pending and re-approvable — the strand is closed. The pre-decide-body equivalence is sound for v1 (no amendment path overwrites the body) and the future-amendment migration point is documented. Both failure paths assert the pending post-condition.

C12: a failed enqueue now returns a transient 451 4.3.0 and logs the detail server-side only, so a correct client retries instead of dropping mail and the internal error never reaches the agent. Tested for both the code and the opacity.

C22: control, bidi, and zero-width runes are neutralized to U+FFFD before truncate on both operator tables. Non-blocking note: U+200C (ZWNJ) is neutralized too, but it is a legitimate Persian typographic joiner that appears in real subjects — on a Persian operator's holds/queue listing that renders as a replacement char. Worth considering whether ZWNJ specifically should pass through (or map to a benign visible marker) rather than U+FFFD, while still neutralizing the genuine spoofing vectors (bidi overrides, ZWSP, BOM). Not a blocker; the full subject stays available via the show command.

CI green. Approving.

cmd/darbaan/main.go
```diff
@@ -1340,6 +1340,30 @@ func truncate(s string, n int) string {
...
+// such rune is replaced with U+FFFD. Apply before truncate so the rune budget
+// counts cleaned text.
+func sanitizeField(s string) string {
+	return strings.Map(func(r rune) rune {
+		switch {
+		case r < 0x20, r == 0x7f: // C0 controls + DEL
+			return '�'
+		case r >= 0x80 && r <= 0x9f: // C1 controls
+			return '�'
+		case r == 0x200e || r == 0x200f, // LRM / RLM
```
sora-rubigd: Non-blocking: the bidi class misses U+061C (ARABIC LETTER MARK) — it is in Unicode's Bidi_Control set alongside LRM/RLM and can reorder weak characters in RTL contexts. Worth adding to this case. Alternative that covers it and stays future-proof: treat any rune in unicode.Cf (format category) as dangerous — that includes all bidi controls, zero-width runes, and soft hyphen — though note the ZWNJ-legibility trade-off already raised in this review would then need an explicit carve-out.

sora-rubigd (dismissed):
All three fixes verified against the full functions, not just the hunks. C5: the pre-computed rewrite is sound — the injected human verdict carries no Amended, so the chain's Released is always the raw body and the pre-decide rewrite is byte-identical to what the old code computed post-decide; there are now no error returns between the decide() commit and the send attempt, and the concurrent-decide race stays safe because decide() re-checks pending. C12: the 451 4.3.0 is opaque, the detail goes to slog only, and the remaining passthrough error is go-smtp's own reader error (e.g. the 552 size limit), which is correct to propagate. C22: the rune classes cover the real spoof vectors, truncate is rune-safe and runs after sanitization.

One note that could not anchor inline (outside the diff): the sync-status table's LAST ERROR column (cmd/darbaan/main.go:874) is now the one remaining unsanitized operator-table field — a sync error string can embed upstream IMAP server response text (a NO/BAD line), so a hostile or MITM'd upstream could put control runes on the operator terminal. Non-blocking; wrapping it in sanitizeField (plus truncation — errors can be long) fits the cleanup PR. One further non-blocking note inline.

sora-rubigd (approved):
Delta re-verified: the ZWNJ passthrough and the ALM addition are applied consistently in both the sanitizer and the test mirror, with tests covering each direction (ALM neutralized; a ZWNJ-bearing subject preserved byte-for-byte). One residual property on the record, accepted by the operator decision: ZWNJ remains invisible in terminal output, so two distinct strings can render identically on the decision surface — it carries no reordering or escape capability, and address-keyed trust paths already normalize on the raw address, so display-only. Approval stands at the new head.

yaad-rubigd (approved):
Re-reviewed the delta (da5504e). ZWNJ (U+200C) now passes through sanitizeField with a documented rationale — it is a semantic Persian joiner with no reordering or escape capability, so it is not a terminal-hijack vector, while ZWSP/ZWJ/BOM and all controls stay neutralized. U+061C (ALM) is added to the bidi kill-set, closing the one Bidi_Control gap. The strip test now covers ALM and the passthrough test asserts a ZWNJ-bearing subject is preserved verbatim. C5 and C12 are unchanged by this commit. CI green. Approving.

PR #243, opened 2026-08-08
fix: add outbound re-send verb, surface send failures, guard send-attempt state

### LINKED ISSUE #232: Queue/outbound correctness: stranded approves + silent send loss (C4/C5/C12/C21/C22/C25/C26/C27)

forge-rubigd:
Part of #232 — the "un-stranding" half. Addresses **C4, C21, C26**. The two remaining guards (**C25** sender-refuse, **C27** header-From vs envelope-From) land in a final PR that `Closes #232`.

The root problem: an approved message whose upstream send failed transiently had **no recovery**. `decide()` refuses a non-pending message, so a re-approve returns `ErrNotPending`, and the message sat in `approved` forever — invisibly, since the listing never showed the send error.

### C4 (HIGH) — admin re-send verb
The promised re-send did not exist. This adds:
- `POST /queue/{id}/resend` (scope `queue:decide`), `Client.ReSend`, and a `queue resend <id>` CLI command.
- Allowed **only** for an approved message carrying a send error; it retries `senderFor().Send` and records the attempt (audited via `RecordSendAttempt`).
- The message is delivered **as stored** — the ADR 0023 slice 5 `ApproveAs` identity rewrite is send-time only and was never persisted, so a re-send of an approved-as message goes through its stamped inbox with its original From. Documented on the method.

### C21 (MEDIUM) — observability
`Meta` omitted `Inbox` and `SendErr`, so a failed send (and which inbox it routed through) was invisible in `queue ls`. Added both to `sluice.Meta`, populated in `List()`, and surfaced as new columns. `SendErr` can carry the upstream server's response text, so it is sanitized like the other operator-table fields (C22).

### C26 (LOW) — RecordSendAttempt status guard
`RecordSendAttempt` stamped `Sent`/`SendErr` onto any state — safe only by its single call site, a footgun for the new re-send verb. It now returns `ErrNotApproved` unless the message is `approved` (idempotent once `sent`).

### Tests
- Re-send recovers a message stranded by a transient failure (approved+SendErr → re-send → sent, error cleared); a pending or already-sent message is refused with `ErrNotResendable`.
- `RecordSendAttempt` refuses pending/rejected records and is idempotent on sent.
- `List()` surfaces `Inbox` and `SendErr`.

### Cross-package touch
`internal/admincfg` — one route→scope entry for `POST /queue/{id}/resend` (`queue:decide`), plus its line in the route-completeness test. The scope model requires every admin route to ship with its mapped scope.

Rebased onto `main` after #241 merged; the diff is exactly these 10 files. Verified locally: `gofmt`, `go build`, `go vet`, `go test -race ./...`, golangci-lint v2.11.4 — all clean.

### DIFF
--- a/internal/admin/service.go
+++ b/internal/admin/service.go
@@ -323,7 +323,7 @@ func (s *Service) Show(id string) (sluice.Message, error) { return s.store.Get(i
 
 // decide runs the approval chain for one message and commits the outcome —
 // nothing more (pure). Everything downstream lives in Approve/Reject.
-func (s *Service) decide(ctx context.Context, id string, human approver.Verdict) (sluice.Message, error) {
+func (s *Service) decide(ctx context.Context, id string, human approver.Verdict, asInbox string) (sluice.Message, error) {
 	msg, err := s.store.Get(id)
 	if err != nil {
 		return sluice.Message{}, err
@@ -353,7 +353,7 @@ func (s *Service) decide(ctx context.Context, id string, human approver.Verdict)
 	}
 	switch outcome.Disposition {
 	case approver.Approve:
-		return s.store.Approve(id, outcome.DecidedBy, outcome.Released)
+		return s.store.Approve(id, outcome.DecidedBy, outcome.Released, asInbox)
 	case approver.Reject:
 		return s.store.Reject(id, outcome.DecidedBy, outcome.Reason, outcome.Retryable)
 	default:
@@ -413,7 +413,7 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 		}
 	}
 
-	m, err := s.decide(ctx, id, approver.Verdict{Disposition: approver.Approve})
+	m, err := s.decide(ctx, id, approver.Verdict{Disposition: approver.Approve}, asInbox)
 	if err != nil {
 		return Outcome{}, err
 	}
@@ -431,7 +431,7 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 	}
 
 	sendErr := s.senderFor(sendInbox).Send(ctx, sendMsg)
-	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr)
+	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr, false)
 	if rerr != nil {
 		return Outcome{}, rerr
 	}
@@ -469,10 +469,21 @@ var ErrNotResendable = errors.New("admin: message is not awaiting re-send (needs
 // SendErr: decide() refuses non-pending messages, so a plain re-approve returns
 // ErrNotPending. Allowed only when Status==approved && SendErr!=""; it re-runs the
 // send and records the attempt (the RecordSendAttempt guard, C26, keeps this the
-// only way SendErr/Sent can be re-stamped). The message is delivered as stored —
-// the ADR 0023 slice 5 ApproveAs identity rewrite is send-time only and was never
-// persisted, so a re-send of an approved-as message goes through its stamped inbox
-// with its original From.
+// only way SendErr/Sent can be re-stamped).
+//
+// An approved-as message (AsInbox set) is re-sent AS APPROVED: the identity rewrite
+// is recomputed from the persisted inbox choice — the same computation the approve
+// path did at send time — so a re-send never silently reverts to the original From.
+// If that inbox no longer resolves to an identity (removed from config while the
+// message was stranded), the re-send is refused fail-closed rather than delivered
+// under the wrong identity.
+//
+// NOTE: the approved+SendErr check and the send are not atomic, so two concurrent
+// re-sends of the same id can both deliver (duplicate mail). The RecordSendAttempt
+// guard keeps the loser from corrupting the record (a failure can't stamp a sent
+// message), but full de-duplication needs an in-flight marker — tracked as a #232
+// follow-up; clearing SendErr before the attempt is NOT an option (it reintroduces
+// the stranded-invisible state).
 func (s *Service) ReSend(ctx context.Context, id string) (Outcome, error) {
 	m, err := s.store.Get(id)
 	if err != nil {
@@ -481,10 +492,28 @@ func (s *Service) ReSend(ctx context.Context, id string) (Outcome, error) {
 	if m.Status != sluice.StatusApproved || m.SendErr == "" {
 		return Outcome{}, fmt.Errorf("%w: message %s is %s", ErrNotResendable, id, m.Status)
 	}
-	slog.Info("resend", "message_id", m.ID, "inbox", inbound.NormInbox(m.Inbox))
 
-	sendErr := s.senderFor(m.Inbox).Send(ctx, m)
-	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr)
+	sendInbox, sendMsg := m.Inbox, m
+	var identity string
+	if m.AsInbox != "" {
+		identity = s.identities[inbound.NormInbox(m.AsInbox)]
+		if identity == "" {
+			// The chosen inbox lost its identity while the message was stranded;
+			// refuse rather than deliver from the original From (fail-closed).
+			return Outcome{}, fmt.Errorf("%w: approved-as inbox %q no longer resolves to an identity", ErrUnknownInbox, m.AsInbox)
+		}
+		rewritten, rwErr := rewriteFrom(sendBody(m), identity)
+		if rwErr != nil {
+			return Outcome{}, fmt.Errorf("admin: rewrite From to %q: %w", identity, rwErr)
+		}
+		sendInbox = m.AsInbox
+		sendMsg.From = identity      // envelope MAIL FROM
+		sendMsg.Released = rewritten // send-time rewrite only; the stored record is untouched
+	}
+	slog.Info("resend", "message_id", m.ID, "inbox", inbound.NormInbox(sendInbox), "identity", identity)
+
+	sendErr := s.senderFor(sendInbox).Send(ctx, sendMsg)
+	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr, true)
 	if rerr != nil {
 		return Outcome{}, rerr
 	}
@@ -509,7 +538,7 @@ func (s *Service) ReSend(ctx context.Context, id string) (Outcome, error) {
 // RejectID rejects a held message: commit, then deliver the DSN bounce
 // (downstream of commit, never mistaken for a reject failure, ADR 0006).
 func (s *Service) RejectID(ctx context.Context, id, reason string, retryable bool) (Outcome, error) {
-	m, err := s.decide(ctx, id, approver.Verdict{Disposition: approver.Reject, Reason: reason, Retryable: retryable})
+	m, err := s.decide(ctx, id, approver.Verdict{Disposition: approver.Reject, Reason: reason, Retryable: retryable}, "")
 	if err != nil {
 		return Outcome{}, err
 	}
--- a/internal/admin/service.go
+++ b/internal/admin/service.go
@@ -460,6 +460,52 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 	return out, nil
 }
 
+// ErrNotResendable is returned by ReSend when the message is not an approved
+// message carrying a recorded send error — the only state a manual re-send acts on.
+var ErrNotResendable = errors.New("admin: message is not awaiting re-send (needs an approved message with a recorded send error)")
+
+// ReSend retries the upstream delivery of an approved message whose previous send
+// failed (C4). It is the only recovery for a message stranded in `approved` with a
+// SendErr: decide() refuses non-pending messages, so a plain re-approve returns
+// ErrNotPending. Allowed only when Status==approved && SendErr!=""; it re-runs the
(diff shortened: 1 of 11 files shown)


yaad-rubigd (approved):
Reviewed against the un-stranding scope (C4/C21/C26); C25/C27 correctly deferred to the closing PR, so #232 stays open.

C4: the re-send verb (POST /queue/{id}/resend, queue:decide scope, client + CLI) recovers an approved message stranded with a send error — gated on Status==approved && SendErr!=empty, re-runs the send + RecordSendAttempt, and branches the outcome (delivered / stub no-op / permanent maps to bounce / transient stays approved for another try). ErrNotResendable maps to 409, distinct from the 404. Both the recovery and the non-resendable refusal are tested.

C21: Meta gains Inbox + SendErr, populated in List and shown in queue ls with the same sanitization as the other operator fields, so the stranded state is now visible.

C26: RecordSendAttempt refuses any status other than approved/sent (ErrNotApproved), so the re-send path cannot resurrect a pending or rejected record; both refusals are tested.

One non-blocking follow-up: a re-send of a message that was originally sent-as a different inbox goes via the stored inbox and original From, not the chosen send-as identity — the send-time rewrite is never persisted, so the operator's identity choice is silently dropped on re-send. It is documented in the code, but worth a follow-up to either persist the as-identity or surface on re-send that it will not be reapplied. Not a blocker: delivery still succeeds and there is no leak.

CI green. Approving.

internal/admin/service.go
```diff
@@ -460,6 +460,52 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
...
+// ReSend retries the upstream delivery of an approved message whose previous send
+// failed (C4). It is the only recovery for a message stranded in `approved` with a
+// SendErr: decide() refuses non-pending messages, so a plain re-approve returns
+// ErrNotPending. Allowed only when Status==approved && SendErr!=""; it re-runs the
+// send and records the attempt (the RecordSendAttempt guard, C26, keeps this the
+// only way SendErr/Sent can be re-stamped). The message is delivered as stored —
+// the ADR 0023 slice 5 ApproveAs identity rewrite is send-time only and was never
+// persisted, so a re-send of an approved-as message goes through its stamped inbox
+// with its original From.
+func (s *Service) ReSend(ctx context.Context, id string) (Outcome, error) {
```
sora-rubigd, line 487: **Blocking.** For an approved-as message this delivers the opposite of the operator's decision: the chosen identity rewrite was send-time only, so the re-send goes out with the original From (header and envelope — the sender prefers Released, and the stored Released is the original body) via the stamped inbox. The exact scenario where the choice mattered most — masking an internal identity — is the one where a transient failure plus a re-send silently unmasks it. The doc comment records this, but the operator surface gives no signal, and nothing in the record marks that an as-decision existed, so a later fix cannot distinguish these messages retroactively: every approved-as message that strands between this merge and a follow-up is permanently unrecoverable as approved.

Scoped fix that stays inside this PR and does not touch the stored-body invariant (the body stays as submitted; only decision metadata is recorded, exactly like DecidedBy already is):

1. Persist the as-choice at approve commit — e.g. `AsInbox string` on the record, stamped in the approve transition when non-empty.
2. Here, when `m.AsInbox != ""`: resolve the identity and recompute the rewrite exactly as the approve path's pre-compute does (`rewriteFrom(sendBody(m), identity)`), route via that inbox's sender, and set envelope From to the identity — a re-send then delivers what the operator actually approved.
3. If the inbox or its identity no longer resolves (removed from config while stranded), refuse with a clear error rather than falling back to the original identity — same fail-closed shape as the planned sender-refuse guard.
4. Tests: re-send of a stranded approved-as message delivers the rewritten From via the chosen inbox; approved-as whose inbox vanished refuses.

If recording the as-choice is judged to conflict with the send-time-rewrite-only design decision, that reading should be settled explicitly before this verb ships, because shipping first forecloses it.

internal/sluice/bbolt.go
```diff
@@ -225,6 +227,13 @@ func (s *bboltStore) RecordSendAttempt(id string, sendErr error) (Message, error
 		if err != nil {
 			return err
 		}
+		// A send is only ever recorded against an approved message (or re-recorded
+		// against an already-sent one, idempotently). Guard so the re-send verb
+		// (or any future caller) can never stamp Sent/SendErr onto a pending or
+		// rejected record (C26).
+		if rec.Status != StatusApproved && rec.Status != StatusSent {
```
sora-rubigd: Non-blocking: two concurrent re-sends of the same id both pass the approved+SendErr check (the check-and-send is outside the transaction) and both deliver — duplicate mail, and the audit shows two send_attempt events but nothing refused. The loser's failure case is quirkier: this guard admits StatusSent, so a failed second attempt stamps SendErr onto a sent record (listing then shows sent-with-error). Tightening the sent case to success-only (refuse `sendErr != nil` when already sent) removes that artifact cheaply. Full dedup needs an in-flight marker — note that naively clearing SendErr before the attempt would reintroduce the stranded-invisible state, so don't take that route.

internal/admin/service.go
```diff
@@ -460,6 +460,52 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
...
+	if err != nil {
+		return Outcome{}, err
+	}
+	if m.Status != sluice.StatusApproved || m.SendErr == "" {
+		return Outcome{}, fmt.Errorf("%w: message %s is %s", ErrNotResendable, id, m.Status)
+	}
+	slog.Info("resend", "message_id", m.ID, "inbox", inbound.NormInbox(m.Inbox))
+
+	sendErr := s.senderFor(m.Inbox).Send(ctx, m)
+	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr)
```
sora-rubigd: Non-blocking: the tamper-evident audit trail records this as the same send_attempt event as a first delivery — an operator-triggered re-delivery is indistinguishable in the audit log (the slog line is not the durable record). A distinct event name or a detail marker keeps re-sends auditable.

sora-rubigd (changes requested):
C21 and C26 verified sound: Meta gains Inbox/SendErr populated in List and surfaced sanitized+truncated, and the RecordSendAttempt guard closes the re-stamp footgun with tests for pending/rejected/idempotent-sent. The C4 verb's mechanics are also right (approved+SendErr precondition, 409 on non-resendable, outcome branching mirroring approve). One blocking gap in what the verb delivers — see the inline comment on the re-send path: for a message that was approved-as, the re-send silently reverts to the original identity, and because no marker of the as-decision is persisted, this cannot be warned about or refused later; every message approved between merge and a follow-up fix is permanently unrecoverable-as-approved. Two further non-blocking notes inline.

yaad-rubigd (changes requested):
Superseding my prior approval — the send-as identity drop on re-send is blocking, not a follow-up, and the reasoning is decisive: the record persists no marker that an as-decision was made. So (a) the re-send path cannot warn or refuse when the as-identity can't be honored, and (b) no later fix can retroactively identify approved-as messages — every approved-as message that strands between this merge and a follow-up would re-send from the wrong identity, silently and unrecoverably. Shipping the verb without persisting the as-choice forecloses the fix for exactly the messages it matters to.

Scoped fix, in-PR: persist the as-inbox choice at approve commit as decision metadata (the stored body stays as submitted, so the send-time-rewrite invariant holds), recompute the rewrite in ReSend via that inbox, and refuse the re-send if the inbox/identity no longer resolves; add tests for the honored-identity re-send and the unresolvable-identity refusal. C21 (Inbox/SendErr visibility) and C26 (RecordSendAttempt guard) and the verb mechanics are otherwise sound.

Also fold in the two inline notes while here: guard concurrent re-send double-delivery and the sent-with-SendErr stamp artifact, and make the audit trail distinguish a re-send from a first send.

sora-rubigd (approved):
Delta verified in full. The identity decision is now persisted as metadata at approve commit with the stored body untouched, the re-send recomputes the rewrite through the same computation as the approve path and routes via the chosen inbox, and a vanished identity refuses fail-closed (409) instead of delivering under the original From — with tests asserting the rewritten header and envelope, the untouched stored record, and the refusal path. The send-record guard now admits only an idempotent success against a sent message, so a losing concurrent attempt cannot produce sent-with-error, and re-sends are distinguishable in the durable audit log under their own event. The remaining concurrent double-delivery window is documented in-code with the unsafe shortcut explicitly ruled out and is tracked; deferring the crash-safe in-flight marker rather than rushing it is the right call for an operator-only surface. Every implementation and caller of the two changed store methods is updated; the only store implementation is the bbolt one.

yaad-rubigd (approved):
Re-reviewed the delta (6452975) — the blocker is closed. The ApproveAs choice is now persisted as decision metadata (AsInbox) at approve commit with the stored body unchanged, and ReSend recomputes the From rewrite from it and routes via that inbox, so a stranded approved-as message re-sends as the operator chose rather than silently reverting to the original From. If the inbox no longer resolves it refuses fail-closed (ErrUnknownInbox, 409) instead of delivering under the wrong identity. Two tests cover the honored-identity recompute (with the stored record kept original) and the vanished-inbox refusal.

The two folded notes are good: RecordSendAttempt now admits only an idempotent success against an already-sent message, so a losing concurrent second attempt can't turn a delivered message into sent-with-error; and a re-send audits under its own resend_attempt event.

Agree with scoping the full concurrent re-send de-duplication out to a follow-up: a crash-safe in-flight marker is a real design task, the residual is duplicate mail on a near-simultaneous manual double-fire (record corruption is already guarded, and clearing SendErr early would reintroduce the stranded-invisible state), so rushing it risks a stuck sending state. Tracking it on the issue.

CI green. Approving.

PR #244, opened 2026-08-08
fix: refuse send on removed inbox sender, reject header/envelope From mismatch

### LINKED ISSUE #232: Queue/outbound correctness: stranded approves + silent send loss (C4/C5/C12/C21/C22/C25/C26/C27)

forge-rubigd:
The final two findings of #232 — **C25** and **C27**. With this, #232 is complete.

`Closes #232`

> **Stacked on #243.** Base is `fix/232-unstrand-c4-c21-c26` so the diff shows only C25/C27 (C25 touches `senderFor`, which #243 also touched). Once #243 merges I'll rebase `--onto main` and retarget the base to `main`. Review the delta here; it will not merge before #243.

### C25 (LOW) — refuse a send whose inbox sender was removed
`senderFor` silently fell back to the default inbox's sender when a stamped inbox had none, so a message authored as inbox X — whose inbox was removed from config while it was held — went out through the *default* account. `newSenders` builds one sender per configured inbox, so a miss unambiguously means removed-from-config. `senderFor` is now strict (returns `ok=false` on a miss), and the approve and re-send paths route through a `sendVia` helper that refuses with `errNoSender`: the message stays approved, visible (`SendErr`), and re-sendable once the inbox is restored — never bounced, never delivered under the wrong account.

### C27 (LOW) — reject header/envelope From divergence at submit
Routing and ADR 0027 send-scoping key on the envelope MAIL FROM, but recipients see the header From, so a send-granted agent could present a header `From:` for an identity it holds no grant on — and the operator's `queue ls` shows the envelope From, so the spoof isn't visible at the decision point.

The submit face now rejects at DATA when the header From address diverges from the envelope sender. Chosen shape: **reject at submit**, not flag-for-approver. Flagging would spend the operator attention C22 just hardened, and the legitimate "send as a different identity" path already exists as the operator's **ApproveAs** (now correctly persisted post-#243). Reject adds **no new state**, so no re-send path has to honor it (the #243 lesson).

Checklist honored:
- normalized RFC 5321 address comparison through the **single** provenance normalizer (`NormalizeAddress`) — `From`, this check, and the ADR 0031 trust chokepoint all share it; no second path.
- address-only comparison — display names are free.
- ambiguous From → reject fail-closed: unparseable header block, **more than one From header**, or a From that isn't exactly one address.
- absent From → allowed (nothing to compare).
- **Scope is From-only by decision** — Reply-To / Sender are a softer reply-path vector, consciously out of scope (a decision, not an omission).

### Tests
- approve and re-send both refuse a removed-sender inbox and stay approved.
- `SubmitFromAddress`: single/normalized/display-name pass; absent allowed; multiple-From, multi-address, and unparseable-block all rejected.
- submit face rejects a header/envelope mismatch end-to-end while accepting a display-named match and an absent From.

Verified locally: `gofmt`, `go build`, `go vet`, `go test -race ./...`, golangci-lint v2.11.4 — all clean.

### DIFF
--- a/internal/admin/service.go
+++ b/internal/admin/service.go
@@ -151,13 +151,22 @@ func (s *Service) SetSenders(senders map[string]backend.Sender) {
 	}
 }
 
-// senderFor returns the upstream sender for a message's inbox, falling back to the
-// default inbox's sender when the message carries no/unknown inbox (ADR 0023).
-func (s *Service) senderFor(inbox string) backend.Sender {
-	if snd := s.senders[inbound.NormInbox(inbox)]; snd != nil {
-		return snd
-	}
-	return s.senders[inbound.DefaultInbox]
+// errNoSender marks a refused send: the message's stamped inbox has no configured
+// sender, so it was removed from config while the message was held (C25). It is a
+// config fault, never permanent — the message stays approved and re-sendable once
+// the inbox is restored, and it never bounces the agent.
+var errNoSender = errors.New("admin: inbox has no configured sender")
+
+// senderFor returns the upstream sender for a message's inbox (ADR 0023). ok=false
+// means the inbox has no configured sender: a message stamped with a named inbox
+// that was removed from config while held (C25). The caller must refuse the send
+// rather than silently deliver it through the default account under the wrong
+// identity. The default inbox is keyed under DefaultInbox and NormInbox folds the
+// empty inbox to it, so a default-routed message resolves here directly — the old
+// fall-through to the default sender only ever masked the removed-inbox case.
+func (s *Service) senderFor(inbox string) (backend.Sender, bool) {
+	snd := s.senders[inbound.NormInbox(inbox)]
+	return snd, snd != nil
 }
 
 // SetInboundHolds wires the inbound hold-for-human queue (ADR 0021/0023): the
@@ -430,7 +439,7 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 		slog.Info("approved-as", "message_id", m.ID, "inbox", inbound.NormInbox(asInbox), "identity", identity)
 	}
 
-	sendErr := s.senderFor(sendInbox).Send(ctx, sendMsg)
+	sendErr := s.sendVia(ctx, sendInbox, sendMsg)
 	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr, false)
 	if rerr != nil {
 		return Outcome{}, rerr
@@ -446,6 +455,8 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 		out.Detail = sentDetail
 	case isSendPending(sendErr):
 		out.Detail = "approved; no real Sender configured — nothing left Darbaan"
+	case errors.Is(sendErr, errNoSender):
+		out.Warn = "inbox has no configured sender (removed from config?); not sent — stays approved"
 	case backend.IsPermanent(sendErr):
 		// Permanent failure: bounce the agent with a generic reason (never the
 		// upstream body). The verdict still committed.
@@ -458,6 +469,18 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 		out.Warn = "upstream send failed (transient); stays approved for re-send"
 	}
 	return out, nil
+}
+
+// sendVia delivers sendMsg through the given inbox's sender, or returns errNoSender
+// when that inbox has no configured sender — refusing rather than silently routing
+// through the default account (C25). Recording it as a send error keeps the message
+// approved, visible (SendErr), and re-sendable once the inbox is restored.
+func (s *Service) sendVia(ctx context.Context, inbox string, sendMsg sluice.Message) error {
+	snd, ok := s.senderFor(inbox)
+	if !ok {
+		return fmt.Errorf("%w: %q", errNoSender, inbound.NormInbox(inbox))
+	}
+	return snd.Send(ctx, sendMsg)
 }
 
 // ErrNotResendable is returned by ReSend when the message is not an approved
@@ -512,7 +535,7 @@ func (s *Service) ReSend(ctx context.Context, id string) (Outcome, error) {
 	}
 	slog.Info("resend", "message_id", m.ID, "inbox", inbound.NormInbox(sendInbox), "identity", identity)
 
-	sendErr := s.senderFor(sendInbox).Send(ctx, sendMsg)
+	sendErr := s.sendVia(ctx, sendInbox, sendMsg)
 	final, rerr := s.store.RecordSendAttempt(m.ID, sendErr, true)
 	if rerr != nil {
 		return Outcome{}, rerr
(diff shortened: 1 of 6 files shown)


yaad-rubigd (dismissed):
Reviewed the C25/C27 delta (stacked on the #243 branch; CI pends the rebase onto main).

C25: senderFor now returns (Sender, ok); a message stamped with an inbox whose sender was removed from config no longer silently falls back to the default account. Both approve and re-send route through sendVia, which returns errNoSender (a non-permanent config fault) so the message stays approved, records the error (visible + re-sendable once the inbox is restored), and never bounces. The note that the old default fall-through only ever masked this removed-inbox case is correct, since default-routed messages resolve directly through the default key. Two tests assert the refusal and that the default sender is not used.

C27: the submit face rejects at DATA when the header From address diverges from the envelope MAIL FROM. SubmitFromAddress enforces an unambiguous From first — unparseable block, more than one From header, or not exactly one address all reject fail-closed (550 5.7.1); absent From is allowed; the compare is address-only through a single NormalizeAddress normalizer now shared by From, this check, and the ADR 0031 trust chokepoint (no second path that could drift). The test covers mismatch-rejected, display-name-match-accepted, and absent-From-accepted. Reply-To/Sender named out of scope. Matches the agreed design and does not false-positive on the fleet's own sends (envelope address equals header address there).

Code approves. Final merge waits on #243 merging, the rebase onto main, and CI green there — I'll confirm the re-based diff is unchanged and CI before merging.

internal/admin/service.go
```diff
@@ -430,7 +439,7 @@ func (s *Service) approve(ctx context.Context, id, asInbox string) (Outcome, err
 		slog.Info("approved-as", "message_id", m.ID, "inbox", inbound.NormInbox(asInbox), "identity", identity)
 	}
 
-	sendErr := s.senderFor(sendInbox).Send(ctx, sendMsg)
+	sendErr := s.sendVia(ctx, sendInbox, sendMsg)
```
sora-rubigd, line 442: Non-blocking asymmetry with the pre-commit validation above: the identity is resolved before the verdict commits, but sender presence is not — an approve-as targeting an inbox that still has an identity but whose sender was removed commits the verdict first, and the message then soft-strands in approved-with-SendErr. It is visible and re-sendable once the config is restored, but the operator can no longer change the identity choice (the record is out of pending, and a re-send is pinned to the persisted as-inbox). A one-line sender-presence check in the pre-commit block (same shape as the identity check) would refuse before the commit, leaving the message pending and the choice still open. Fine as a follow-up ride-along since the issue's prescribed stay-approved semantics are what shipped.

sora-rubigd (dismissed):
Both fixes verified against the checklist and the surrounding code. C27: the normalizer is genuinely single-path (SubmitFromAddress and From both return through NormalizeAddress, and the envelope side is normalized through the same function at the compare); every ambiguous shape rejects fail-closed — unparseable block, multiple From headers, multi-address or empty-group From — with unit tests for each, plus listener-level tests for mismatch/display-name/absent-From; the SMTP listener is the only non-test Enqueue caller, so the check covers every agent submission path; scope boundaries are documented in code and body. C25: NormInbox folds the empty inbox to the default key so default-routed legacy records resolve directly, and the strict form also fixes a latent crash — under a named-inbox config with no default entry, the old fallback returned a nil interface and the send would have panicked; refusing with a recorded send error is strictly better. The errNoSender outcome branch is ordered before the permanent-failure branch so a refused send can never bounce the agent. One non-blocking note inline. Approval is on the stacked delta; the merge gate still includes CI green after the rebase onto the default branch.

yaad-rubigd (approved):
Re-approving the rebased HEAD (base=main): confirmed the rebased diff is byte-identical to the reviewed delta (C25 strict senderFor + sendVia refusal, C27 header/envelope-From reject-at-submit via the shared NormalizeAddress). CI green (Analyze actions/go, CodeQL, check). Approval carries; clear to merge.

sora-rubigd (approved):
Re-approval after the rebase onto the default branch: the rebased diff was verified byte-identical to the previously approved delta (full patch text compared), so the prior review's findings carry unchanged.

PR #242, opened 2026-08-08
chore(main): release 0.17.0

yaad-rubigd:
:robot: I have created a release *beep* *boop*
---

### Upgrade notes

**Expect repeated label-write deferrals for messages stored before this series, and do not chase them as a fault.** The X-GM-LABELS writer now refuses a label write whenever the mailbox validity is unknown on either side. Messages recorded before the validity field existed therefore defer their label writes instead of replicating them, and because reconcile clears the dirty flag only on success, each one re-attempts and logs a deferral on every sync. That is the intended trade: writing into an unconfirmed UID space is the worse outcome. The deferrals stop once those records carry a validity; a backfill is tracked separately.

**Assessment holds may differ slightly if you previously lowered `threshold`.** Recipient-skew scoring was corrected, so a deployment tuned around the old behaviour will hold marginally less than before. The `assessment:` settings are the supported way to re-tune.

### Features

* wire assessment tunables, agent advisory, identity-less scoring ([#233](https://github.com/yaad-index/darbaan/issues/233) part 3) ([#252](https://github.com/yaad-index/darbaan/issues/252)) ([9ac2b75](https://github.com/yaad-index/darbaan/commit/9ac2b75bb88258a2ec5c340b74af97386bd7b244))

### Bug Fixes

* add outbound re-send verb, surface send failures, guard send-attempt state ([#243](https://github.com/yaad-index/darbaan/issues/243)) ([6568faa](https://github.com/yaad-index/darbaan/commit/6568faab77748a7e655cdffbbb669c0646853f80))
* audit tail-truncation detection, inbound verdicts + attribution ([#236](https://github.com/yaad-index/darbaan/issues/236)) ([#246](https://github.com/yaad-index/darbaan/issues/246)) ([35bcccb](https://github.com/yaad-index/darbaan/commit/35bcccbd4792d74b11d382d51bf71eaa284f1e82))
* bound X-GM-LABELS writes to mailbox validity and honor ctx in sync ([#235](https://github.com/yaad-index/darbaan/issues/235) part 2) ([#256](https://github.com/yaad-index/darbaan/issues/256)) ([672cc3d](https://github.com/yaad-index/darbaan/commit/672cc3d67607fa7c5bc2abb086d80f3936fe755d))
[...]

PR #246, opened 2026-08-09
fix: audit tail-truncation detection, inbound verdicts + attribution (#236)

### LINKED ISSUE #236: Audit integrity: tail-truncation undetectable, inbound verdicts + attribution unaudited (C15/C16/C28/C29/C30)
Audit-integrity gaps: ADR 0011 promises "every verdict, tamper-evident" but the tail-truncation check is missing, inbound verdicts aren't audited at all, and attribution is dropped.

### C15 — MEDIUM — `AuditLog.Verify` cannot detect tail truncation, and nothing in production ever calls it
`internal/audit/bbolt.go:80-108`, `internal/audit/audit.go:54`. (a) `Verify` checks hash links but never `Seq` continuity or key↔Seq agreement — deleting the last N entries verifies clean, and later appends chain onto the surviving head (truncate-then-append undetectable), contradicting ADR 0011's "tampering or truncation is evident". (b) No CLI/endpoint invokes `Verify`.
**Fix:** assert `Seq == prev+1` (and Seq matches the key) in `Verify`; add a `darbaan audit verify` CLI command.

### C16 — MEDIUM — The ApproveAs identity rewrite is never recorded in the audit log
`internal/admin/service.go:368-382`. ADR 0011 requires "who/what approved" and original-vs-edited fidelity; the chosen identity exists only in a non-tamper-evident `slog.Info`. The trail cannot reconstruct that a message was sent as a different sender than stored.
**Fix:** pass the as-identity down so the approve/send-attempt audit `Detail` includes it (e.g. `"sent as work@x.test"`).

### C29 — LOW — Inbound hold verdicts (Expose/Drop) are not audited at all
`internal/admin/service.go:255-279`. ADR 0011's "every verdict" is implemented only for outbound.
**Fix:** give `admin.Service` the audit log; append `expose`/`drop` records in `setHold`.

### C30 — LOW — The acting admin client's name is discarded after token resolution
`internal/admin/http.go:20-23, 158-169`. The `credential` struct drops `ScopedClient.Name`, so verdicts audit as `DecidedBy:"manual"` regardless of which client acted — ADR 0029's named clients make attribution possible; the code throws it away.
**Fix:** carry the name into `credential`; log/audit it on state-changing routes.

### C28 — LOW — Permanent (5xx) rejections aren't distinguishable in the audit log
`internal/sluice/bbolt.go:211-218`. `Retryable` isn't recorded and no distinct operator surface exists, though ADR 0006 calls them a security signal.
**Fix:** include the flag in the audit `Record`; WARN on permanent rejects.

forge-rubigd:
Closes #236. Audit-integrity hardening from the review backlog — ADR 0011 promises "every verdict, tamper-evident", but tail truncation was undetectable, inbound verdicts were unaudited, and operator/identity attribution was dropped.

## Findings addressed

**C15 (MEDIUM) — tail-truncation undetectable + `Verify` never invoked.** `AuditLog.Verify` now asserts, per entry, that the key equals the entry's own `Seq` (key↔Seq agreement) and that `Seq` is a gap-free run from 1; after the walk it checks the last entry's `Seq` against the bucket's sequence counter, which is monotonic and survives deletes. A plain tail deletion leaves the counter ahead of the head; a truncate-then-append resumes at the higher counter value and gaps the run — both are now caught. Adds `darbaan audit verify`, which opens the log read-only and exits non-zero on any violation (usable as an incident-response / monitoring gate).

**C16 (MEDIUM) — ApproveAs identity unrecorded.** The chosen send identity is threaded into the send-attempt audit `Detail` (`sent as <identity>`, and `… (as <identity>)` on failure), so the trail reconstructs that a message was sent as a different sender than stored.

**C29 (LOW) — inbound verdicts unaudited.** `admin.Service` gains the audit log (the same one the message store writes to); `setHold` appends an `expose`/`drop` record. Best-effort and non-fatal, matching the store's own audit contract; a nil sink (tests) is a no-op.

**C30 (LOW) — acting client name discarded.** The credential now carries the ADR 0029 client name; the `scoped` middleware attaches it to the request context, and it flows to the new `audit.Record.Actor` field on every state-changing verdict (approve/reject/send/resend/expose/drop). Verdicts now attribute who decided, not just the approver name.

**C28 (LOW) — permanent rejects indistinguishable.** A reject's permanence is recorded in `audit.Record.Retryable` (a pointer — set on reject rows, nil elsewhere), and a permanent reject logs a WARN.

## Cross-package touch

The `sluice.MessageStore` interface gains an `actor` argument on `Approve`/`Reject` and `actor, asIdentity` on `RecordSendAttempt` (the store is the single writer of outbound verdict audit rows, so attribution belongs there). Callers are `admin.Service` only; the churn elsewhere is mechanical test-call updates.

## Testing

`go test -race ./...`, `go vet ./...`, golangci-lint v2 — all clean. New tests cover tail truncation, truncate-then-append, interior gap, key↔Seq mismatch, read-only verify (audit); ApproveAs identity + actor in the send audit and the reject Retryable flag (sluice); inbound expose/drop auditing (service); and end-to-end that a named scoped client is recorded as the verdict actor (http). CLI smoke-tested (`audit verify` on null / missing / populated logs).

This is a `fix:` so it accrues to the release-please batch (#242).

### DIFF
--- a/internal/audit/audit.go
+++ b/internal/audit/audit.go
@@ -18,13 +18,21 @@ import (
 
 // Record is the caller-supplied content of an audit entry. Agent + Inbox are the
 // acting principal and the inbox it acted as (ADR 0027): every row answers "which
-// agent did what, as which inbox".
+// agent did what, as which inbox". Actor names the operator client that made an
+// operator decision (ADR 0029 named clients), distinct from the message's own
+// agent — it is empty for agent-originated events (e.g. enqueue).
 type Record struct {
 	Event     string `json:"event"`
 	Agent     string `json:"agent"`
 	Inbox     string `json:"inbox,omitempty"` // the inbox the action was scoped to (ADR 0023/0027)
+	Actor     string `json:"actor,omitempty"` // the operator client that decided (ADR 0029); "" for agent-originated events
 	MessageID string `json:"message_id"`
 	Detail    string `json:"detail,omitempty"` // e.g. reject reason, send-attempt result
+	// Retryable records a reject's permanence (ADR 0006): a permanent (non-retryable)
+	// reject is an operator security signal. It is a pointer so it is meaningful only
+	// where it applies — nil (omitted) for every non-reject event, rather than a bare
+	// false that would read as "permanent" on rows the flag does not describe.
+	Retryable *bool `json:"retryable,omitempty"`
 }
 
 // Entry is a persisted, hash-chained audit entry. Hash binds PrevHash and the
@@ -43,12 +51,18 @@ type Entry struct {
 type AuditLog interface {
 	// Append records one entry. For the bbolt log it links into the hash chain.
 	Append(Record) error
-	// Verify reports the first integrity violation in the log, if any.
+	// Verify reports the first integrity violation in the log, if any. It proves
+	// the entries are intact AND that none were removed from the log: the
+	// prev_hash links, the per-entry hash, a gap-free Seq run keyed to the record,
+	// and — via the store's monotonic sequence counter, which survives deletes —
+	// that the tail was not truncated. A truncate-then-append resumes at a higher
+	// counter value, so it breaks the Seq run and is no longer undetectable
+	// (ADR 0011).
 	//
-	// Integrity is NOT completeness: Verify proves the prev_hash links between
-	// the entries that ARE present are intact, but it cannot detect entries that
-	// were never written (the best-effort gap, since audit is written after the
-	// message-store commit). Detecting such gaps would be a separate
+	// Integrity is NOT completeness across stores: Verify cannot detect an entry
+	// that was never written in the first place (the best-effort gap, since audit
+	// is written after the message-store commit — a missing entry never enters
+	// this log, so there is no counter to be short). Detecting that is a separate
 	// message-store-to-audit cross-reference, not Verify's job — see the seam in
 	// the bbolt implementation. An empty or disabled log verifies clean.
 	Verify() error
--- a/internal/sluice/bbolt.go
+++ b/internal/sluice/bbolt.go
@@ -198,8 +198,8 @@ func (s *bboltStore) Get(id string) (Message, error) {
 	return s.withRaw(rec)
 }
 
-func (s *bboltStore) Approve(id, decidedBy string, released []byte, asInbox string) (Message, error) {
-	return s.transitionFromPending(id, "approve", decidedBy, func(m *Message) {
+func (s *bboltStore) Approve(id, decidedBy string, released []byte, asInbox, actor string) (Message, error) {
+	return s.transitionFromPending(id, "approve", decidedBy, actor, nil, func(m *Message) {
 		m.Status = StatusApproved
 		m.DecidedBy = decidedBy
 		// An amended body (ADR 0004) is recorded inline; it is nil in v1 (no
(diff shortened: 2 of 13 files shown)


internal/sluice/bbolt.go
```diff
@@ -198,8 +198,8 @@ func (s *bboltStore) Get(id string) (Message, error) {
 	return s.withRaw(rec)
 }
 
-func (s *bboltStore) Approve(id, decidedBy string, released []byte, asInbox string) (Message, error) {
-	return s.transitionFromPending(id, "approve", decidedBy, func(m *Message) {
+func (s *bboltStore) Approve(id, decidedBy string, released []byte, asInbox, actor string) (Message, error) {
```
sora-rubigd, line 201: Non-blocking: the approve row itself never records that an identity choice existed — asInbox arrives here but only DecidedBy goes into the audit Detail, and the as-identity first reaches the tamper-evident log at the send attempt. In the crash window between the approve commit and the send attempt (the window the best-effort contract already acknowledges), the trail shows a plain approve while the store record carries AsInbox — reconstructable, but only from the non-tamper-evident store. Appending the as-inbox to this row's Detail when non-empty (e.g. `"manual, as work"`) closes it for one line and keeps the verdict row self-describing.

internal/audit/audit.go
```diff
@@ -43,12 +51,18 @@ type Entry struct {
 type AuditLog interface {
 	// Append records one entry. For the bbolt log it links into the hash chain.
 	Append(Record) error
-	// Verify reports the first integrity violation in the log, if any.
+	// Verify reports the first integrity violation in the log, if any. It proves
```
sora-rubigd, line 54: Non-blocking, doc precision: "none were removed" holds against partial tampering — deletion, edit, relocation, truncate-then-append — but not against an adversary with write access to the audit database who re-chains the entire log from genesis and resets the bucket's sequence counter (bbolt exposes SetSequence). That is inherent to any self-contained unkeyed hash chain, not a defect of this implementation; the eventual hardening is anchoring the head hash externally (periodic export to a log line, a monitoring system, or a signed checkpoint), at which point Verify's guarantee becomes as strong as the anchor. Worth a sentence here so the guarantee's boundary is explicit, and possibly a line in the follow-ups issue.

sora-rubigd (approved):
All five findings verified against the full code paths, not just the hunks. C15: Verify now checks all four properties — key↔Seq agreement (with Seq bound into the hash payload, so an entry cannot be renumbered to fake continuity without breaking its hash), a gap-free run from 1, the prev_hash/hash chain, and the bucket's delete-surviving sequence counter against the last entry's Seq. That last check is the one that matters most: plain tail deletion (no re-append) leaves a clean chain AND contiguous Seqs, so only the counter comparison catches it — and it is present, with a dedicated test per tampering shape (pure truncation, truncate-then-append, interior gap, relocated key). Chain compatibility on upgrade also holds: the new Record fields are omitempty, so pre-upgrade entries re-marshal byte-identically and old chains still verify. C16: the identity reaches the send-attempt Detail on both the approve and the re-send path, and is empty for as-stamped sends, so plain sends never mis-audit as identity-rewritten. C29: setHold audits after the store commit, best-effort and non-fatal, and serve wires the same single audit instance the message store writes to (one bbolt open; a second open would deadlock on the exclusive lock). C30: the credential name flows middleware → context → store verdict rows, with an end-to-end scoped-client test; the root token audits as "root". C28: Retryable rides as a pointer only on reject rows, nil elsewhere, plus the WARN on permanent rejects.

One scope observation for the record, not blocking: `audit verify` under audit-type=null prints a notice but exits 0, so if a deployment's config ever drifts from bbolt to null, a monitoring gate built on the exit code goes permanently green rather than flagging that there is no longer a chain to verify. A non-zero or distinguishable exit for "nothing to verify" would make the gate fail toward attention. Fine as-is for incident response; worth considering if this becomes a scheduled check. Two further non-blocking notes inline. Approval is on head 43b7c09; the merge gate still includes CI green.

yaad-rubigd (approved):
Reviewed the full diff and independently traced the audit-integrity changes; CI green.

**C15 (tail-truncation)** — Verify now enforces key↔Seq agreement, a gap-free Seq run from 1, and a tail check against the bucket's monotonic sequence counter (which survives deletes). This closes the truncate-then-append gap for the normal delete/append path, and the tests exercise all four shapes: plain truncation, truncate-then-append, interior gap, and key relocation. Confirmed it stays clean on an untampered chain, so existing on-disk logs keep verifying after upgrade.

Non-blocking limitation worth recording: the tail check trusts the sequence counter. An adversary with raw write access to the database file can reset that counter to match a truncated head (and could equally recompute the whole hash chain), so truncation is evident against normal-path tampering but not against arbitrary-file rewrite. That is inherent to a self-contained log with no external anchor (offsite head-hash export or external signing). The godoc is honestly scoped and does not overclaim; suggest a one-line note there, or a follow-up, that full tamper-evidence needs an external anchor.

**C16 / C28 / C29 / C30** — the ApproveAs identity lands in the send-attempt Detail; Retryable is a pointer so it reads as nil on non-reject rows rather than a false "permanent", with a WARN on permanent rejects; the acting client name is context-carried and correctly empty for agent-originated enqueue; inbound expose/drop verdicts audit best-effort and non-fatal, matching the store's existing audit contract. Signatures are threaded consistently through the interface, call sites, and tests.

Approving.

Issue #237, opened 2026-08-08
ADR hygiene + unimplemented mandates (A1-A17, C11 retry cap, C18 secrets-at-rest)


yaad-rubigd:
The ADRs are the definitive flow, so defects here are spec defects. Most of this is a mechanical hygiene pass; a few items need an actual design decision (flagged **DECISION**) and are why this carries `needs-decision`.

### High
- **A1 — ADR 0031's Authentication-Results gate is spoofable exactly where it's recommended.** It reads AR headers whose `authserv-id` matches the configured value, but a sender can inject an AR header bearing that exact authserv-id; safe only if the upstream MTA strips forged AR headers claiming its own identity (RFC 8601 §5) — the behavior the "weaker upstreams" this targets may lack. No tiebreak for multiple matching AR blocks. **Fix:** state the strip assumption + restrict to the topmost AR block(s) above the upstream's own `Received`, or demote the "unspoofable" claim and document the real prerequisite.

### Medium (mechanical)
- **A2** — 0032 fail-safe contradicts itself on assessor-absent (hold vs cheap-terms auto-pass). Distinguish *configured-but-failing* (→hold) from *not configured* (→cheap-terms), state which the code does.
- **A3** — 0032 "the agent never sees the raw payload" is contradicted twice in the same ADR. Scope the claim to held-and-undecided/rejected.
- **A4** — 0032 band cutoffs `0–33/33–66/66–100` double-assign 33 and 66; default "≈70" ≠ high-band start 66; "≈70" not implementable. Define half-open intervals `[0,33)`,`[33,66)`,`[66,100]`, set default threshold to exactly the high cutoff; code uses 70 — align ADR or code.
- **A5** — 0032 Amendment 1 asserts single-write atomicity 0018's blob/metadata split gave up. Reword to the ordering invariant (metadata written only after body + disposition). *(Already satisfied in code by the eager rework; the ADR text still needs it.)*
- **A9** — Status hygiene broken set-wide (0005/0010/0013 unmarked-superseded; 0017 Proposed while Accepted ADRs depend on it; 0021/0019/0026/0028 Proposed but cited as implemented). One hygiene pass: supersession pointers + promote implemented ADRs to Accepted.
- **A10–A15** — Miscitations (A10), invalid example config (A11), token-compare parity (A12, also code C31), status inconsistency (A13), immutability rule violated incl. an in-place rewrite (A14), hold-vs-rule-edit precedence unspecified (A15). Mechanical corrections.

[...]

Issue #233, opened 2026-08-08
Injection-assessment hardening: extraction gaps, detector bypasses, unwired tunables (C17/C19/C20/C23/C24/C36/C38/C39/C44)


yaad-rubigd:
Injection-assessment hardening: extraction gaps that let a crafted message hide its payload from the assessor, detector bypasses, the unwired operator tunables, and the missing agent-facing advisory. C41/C43/C42/C6 are tracked separately and out of scope here.

### C19 — MEDIUM — `mailtext.walk` swallows mid-stream MIME errors without setting `truncated`
`internal/mailtext/mailtext.go:132-141`. Any `mr.NextPart()` error (not just `io.EOF`) breaks silently with `Truncated==false`; a malformed structure after a benign first part gets later parts skipped, the assessment reads clean, and the raw (with the unscanned payload) is still served.
**Fix:** `if errors.Is(err, io.EOF) { break }`; otherwise set `st.truncated = true` before breaking. Add a malformed-multipart test.

### C20 — MEDIUM — An undecodable part is only flagged `Truncated`, which scores zero
`internal/mailtext/mailtext.go:228-235` + `internal/riskscore/score.go:155-182`. A bogus `charset=` makes go-message error on decode (assessor never sees the text) while plain-ASCII under the fake label reads fine for an agent consuming raw MIME. Truncation adds no points and never holds.
**Fix:** make truncation/undecodable-part a scored factor, or treat per-part decode failure as not-cleared per Amendment 1's "extraction hard-fails → held".

### C17 — MEDIUM — ADR 0032 §7 operator tunables are unwired; `riskscore.Parse` is dead code
`cmd/darbaan/main.go:1144`, `internal/riskscore/config.go:88`. Serve hardcodes `riskscore.DefaultConfig()`; the YAML overlay built for exactly this has zero production callers.
**Fix:** parse an `assessment:` section from `configBytes()` via `riskscore.Parse` into `buildAssessHook`; document in config.example.yaml.

### C23 — MEDIUM — The agent-visible advisory (summary/score/factors for below-threshold mail, ADR 0032 §6) is unimplemented
`internal/listener/imap.go`. The persisted assessment is never surfaced to the agent.
**Fix:** stamp system-authored headers (`X-Darbaan-Risk`, `X-Darbaan-Risk-Factors`) inside the already-stripped-then-stamped `X-Darbaan-` namespace at serve time (Layer-1 strip makes them spoof-proof).

[...]

PR #249, opened 2026-08-10
fix: hold messages whose content cannot be decoded for assessment (#233 part 1)

forge-rubigd:
Part 1 of 3 for #233 (injection-assessment hardening — the code-hardening cluster; ADR/decision-bearing items stay in #237, which is operator-gated). Theme: **the payload can't hide from the assessor**. ADR 0032 is live in prod, so this is scoped tight to tune-don't-break.

## Findings addressed

**C19 (MEDIUM) — `mailtext.walk` swallowed mid-stream MIME errors.** The multipart loop broke on *any* `NextPart()` error, so a malformed structure after a benign first part skipped every later part unseen: the assessment read clean while the raw (with the unscanned payload) was still served. It now treats `io.EOF` as the normal end and any other error as an extraction hard-fail.

**C20 (MEDIUM) — an undecodable part scored zero and never held.** A part go-message can't decode (e.g. a bogus `charset=`) previously only set `Truncated`, which adds no points and never holds — the assessor never saw the text, but an agent reading raw MIME did. Extraction now carries a distinct `Content.Undecodable`, set on a genuine per-part read/decode failure or a mid-stream structural break, kept **separate** from the benign `Truncated` cap flag. The screener holds `Undecodable` content fail-safe (`NotCleared`, ADR 0032 Amendment 1) and never reaches the assessor; a benign cap still scores normally, so large-but-readable mail is **not** newly held (no over-hold on a live feature). A soft per-part decode error (unknown charset, part still usable raw) flags `Undecodable` without abandoning the message's other readable parts.

## Design note

`Truncated` (benign bound — real content, just more of it) and `Undecodable` (hard-fail — content never reached the assessor) are deliberately distinct so only genuine failures hold. go-message surfaces an unknown charset/encoding as a non-fatal error alongside a usable raw entity; that error is captured at both the top-level `Read` and per-part `NextPart`.

## Tests (C24, extraction paths)

Malformed multipart -> `Undecodable`; undecodable charset -> `Undecodable`; a cap hit -> `Truncated` not `Undecodable`; a bad part keeps readable siblings; screener holds on `Undecodable` and skips the assessor, but proceeds on a benign truncation.

`go test -race ./...`, `go vet`, golangci-lint v2 — all green. `fix:` -> accrues to release-please #242. Next: PR-2 (detector hardening C38/C39/C44), PR-3 (tunables + advisory + identity C17/C23/C36).

### DIFF
--- a/internal/mailtext/mailtext.go
+++ b/internal/mailtext/mailtext.go
@@ -23,6 +23,7 @@ package mailtext
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
@@ -43,8 +44,15 @@ type Content struct {
 	// their decoders are a follow-up.
 	Attachments []Attachment
 	// Truncated is true if any cap (per-part, total, part-count, or depth) was hit,
-	// so a caller knows the extraction is partial.
+	// so a caller knows the extraction is bounded — the content it holds is real,
+	// there is just more of it beyond the limits.
 	Truncated bool
+	// Undecodable is true if a part could not be decoded or read at all, or the MIME
+	// structure errored mid-stream, so text the message carries never reached the
+	// assessor (C19/C20). Unlike Truncated (a benign bound on readable content), this
+	// is an extraction hard-fail: the caller must not clear the message on it — the
+	// assessor cannot vouch for content it never saw (ADR 0032 fail-safe, Amendment 1).
+	Undecodable bool
 }
 
 // Attachment is one attachment part.
@@ -107,21 +115,30 @@ func Extract(raw []byte, lim Limits) (Content, error) {
 		return Content{}, fmt.Errorf("mailtext: unreadable message")
 	}
 	st := &walkState{lim: lim.withDefaults()}
+	if err != nil {
+		// go-message returns a usable entity alongside a non-nil error for a soft
+		// decode failure — an unknown charset or transfer-encoding — where the body
+		// is only available raw/undecoded (C20). The assessor's decoded view is then
+		// unreliable, so flag the extraction undecodable; the caller holds fail-safe.
+		st.undecodable = true
+	}
 	st.walk(ent, 0)
 	return Content{
 		Body:        strings.TrimSpace(st.body.String()),
 		Attachments: st.atts,
 		Truncated:   st.truncated,
+		Undecodable: st.undecodable,
 	}, nil
 }
 
 type walkState struct {
-	lim       Limits
-	parts     int
-	totalText int
-	truncated bool
-	body      strings.Builder
-	atts      []Attachment
+	lim         Limits
+	parts       int
+	totalText   int
+	truncated   bool
+	undecodable bool
+	body        strings.Builder
+	atts        []Attachment
 }
 
 func (st *walkState) walk(ent *gomessage.Entity, depth int) {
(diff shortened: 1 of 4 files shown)


internal/mailtext/mailtext.go
```diff
@@ -220,18 +254,20 @@ func (st *walkState) budgetText(text string) string {
...
+// error — go-message decodes transfer-encoding and charset here, so a bogus
+// charset= surfaces as failed with the text never seen (C20). The caller flags a
+// cap as Truncated but a failure as Undecodable, so a hard-fail can hold while a
+// bounded read does not.
+func (st *walkState) readCapped(r io.Reader) (text string, capped, failed bool) {
 	limit := st.lim.MaxPartText
 	b, err := io.ReadAll(io.LimitReader(r, int64(limit)+1))
 	if len(b) > limit {
-		return string(b[:limit]), true
+		return string(b[:limit]), true, false
```
sora-rubigd, line 268: Non-blocking: the early return on a cap hit masks a concurrent read error — io.ReadAll can return limit+1 bytes AND a non-nil error when the underlying decoder produces bytes and then fails in the same Read, and this path reports that as capped-benign with failed swallowed. The reachable residual is equivalent to the accepted beyond-cap blindness (an attacker gains nothing over padding past the cap benignly), so this is robustness rather than a live bypass, but the two signals are independent facts and cheap to report independently: `return string(b[:min(len(b),limit)]), len(b) > limit, err != nil`. Fine as a ride-along in a later part of this series.

sora-rubigd (approved):
Verified against the full extraction and screening paths. The load-bearing design choice is correct: a mid-stream MIME break (C19) sets the new Undecodable hard-fail flag, not the benign Truncated bound — this matters because Truncated deliberately does not hold, so routing the C19 case there would have left the original attack scoring clean. The screener checks Undecodable BEFORE invoking the assessor and holds not-cleared regardless of sender trust, with the spy-detector test proving the assessor is never consulted on content it could not have fully seen. The soft-error branch is also right: an unknown-charset part flags the extraction but keeps processing readable siblings, so one bad part does not abandon the rest of the message's evidence, while a structural break (nil part) stops since everything after it is unreachable anyway. The top-level gomessage.Read soft-error case (usable entity + non-nil error) is covered too. Semantics migration is clean: the only non-test consumer of Truncated is the assessor's summary, which undecodable content now never reaches. Tests cover all three shapes the issue demanded — malformed multipart, bogus charset, and cap-vs-failure separation. One non-blocking robustness note inline. Approval on head 534de4d; merge gate still includes CI green.

yaad-rubigd (approved):
Reviewed PR-1/3; CI green. The C19/C20 fix is correct and the Truncated-vs-Undecodable split is the right call.

Traced the extraction-hold invariant end to end:
- C19: a mid-stream `NextPart` error now sets `Undecodable` (a structural break with `p == nil` stops the walk; a soft per-part error keeps readable siblings), and the screener fail-safe-holds on `Undecodable`. So a malformed-multipart message with skipped-but-still-served parts no longer reads clean. Confirmed by the malformed-multipart test.
- C20: a whole-message soft decode error and a per-part read/decode failure both set `Undecodable` → hold; a bogus `charset=` is covered. `readCapped` correctly separates a benign cap (`Truncated`, real bounded content, does not hold) from a hard-fail (`Undecodable`, holds), resolving the "truncation scores zero and never holds" gap without over-holding large-but-readable mail.
- Tests cover malformed multipart, undecodable charset, sibling preservation, and the cap-vs-fail distinction.

One non-blocking note: cap-based truncation (part-count / depth / total-text limits) is treated as benign and does not hold, but parts beyond those caps are unseen by the assessor while still served to the agent in raw — the same hide-the-payload shape as C19, just via a defensive limit rather than a parse error. Worth deciding whether a high part-count/depth truncation should score or hold, weighed against the over-hold cost on legitimate many-part mail; could be a follow-up. Not blocking this PR, which correctly implements the chosen approach.

Approving.

PR #250, opened 2026-08-10
fix: harden injection detector against evasion (#233 part 2)

forge-rubigd:
Part 2 of 3 for #233 (injection-assessment hardening — code-hardening cluster; ADR/decision items stay in #237, operator-gated). Theme: **the detector is harder to slip past**. Best-effort/defense-in-depth — the human send-gate and sender baseline remain the real backstops. ADR 0032 is live in prod, so behaviour changes are called out below.

## Findings addressed

**C44 (LOW-MEDIUM) — fence neutralization was case-sensitive.** Only exact-case `[BEGIN UNTRUSTED`/`[END UNTRUSTED` were neutralized, so a mixed-case spoof (`[End Untrusted email body]`) survived and could visually terminate the fence for a human — or an LLM summarizing the alert. Now case-insensitive (`(?i)\[(BEGIN|END) UNTRUSTED` -> `[${1}_UNTRUSTED`).

**C39 (LOW) — secrets factor fired on a noun alone.** "password" anywhere + the unknown-sender baseline (30) + secrets (40) = 70 = threshold, so receipts, resets, and newsletters were held and trained the operator to rubber-stamp `expose`. It now requires a request verb within a bounded, same-clause span of the secret noun — keeping the genuine "send me your password" vector while dropping bare-mention false positives. **Behaviour note:** this is a net *reduction* in holds (the safer direction for over-hold on a live feature); reviewers may want to eyeball the verb list for coverage.

**C38 (LOW) — cheap detector bypasses.** (a) Zero-width / soft-hyphen code points are stripped before matching, so a keyword split by an invisible character no longer defeats the word-anchored patterns. Stripping is **match-only** — the served content the agent/operator sees is untouched (mailtext keeps invisibles on purpose, since hiding text is itself a signal). (b) A `message/rfc822` attachment is now recursed into (bounded by the existing depth cap), so a directive smuggled in an attached `.eml` is extracted and scanned instead of left as opaque attachment metadata.

## Cross-package touch (`internal/mailtext`)

The C38 rfc822 recursion lives in mailtext extraction, and I folded in the **PR-1 review ride-along** here: `readCapped` now returns the cap hit and the read error independently, so a cap hit can no longer swallow a concurrent decode failure (`io.ReadAll` can return `limit+1` bytes AND an error in one call). The other PR-1 ride-along — whether high part-count/depth cap-truncation should score or hold — is a disposition decision on a live feature, filed as a follow-up rather than changed here.

## Tests

Mixed-case + exact-case fence neutralization; zero-width keyword still matches; secrets requires a request verb (bare noun does not fire, genuine request does); rfc822 attachment directive is extracted into the scannable body.

`go test -race ./...`, `go vet`, golangci-lint v2 — all green. `fix:` -> accrues to release-please #242. Next: PR-3 (tunables C17 + agent advisory C23 + identity skew C36).

### DIFF
--- a/internal/assessor/fence.go
+++ b/internal/assessor/fence.go
@@ -1,6 +1,15 @@
 package assessor
 
-import "strings"
+import (
+	"regexp"
+	"strings"
+)
+
+// fenceMarker matches a fence begin/end marker in any case, so a mixed-case
+// spoof ("[End Untrusted email body]") is neutralized just like the exact-case
+// form (C44). ${1} is brace-delimited so the group ref is not read as the
+// identifier "1_UNTRUSTED".
+var fenceMarker = regexp.MustCompile(`(?i)\[(BEGIN|END) UNTRUSTED`)
 
 // Fence wraps untrusted text so it crosses the boundary to the privileged agent
 // as clearly-delimited, inert data — never as instructions (the ADR 0006
@@ -17,9 +26,9 @@ func Fence(label, text string) string {
 	label = sanitizeLabel(label)
 	begin := "[BEGIN UNTRUSTED " + label + "]"
 	end := "[END UNTRUSTED " + label + "]"
-	// Neutralize any spoofed markers so the payload cannot terminate the fence.
-	text = strings.ReplaceAll(text, "[BEGIN UNTRUSTED", "[BEGIN_UNTRUSTED")
-	text = strings.ReplaceAll(text, "[END UNTRUSTED", "[END_UNTRUSTED")
+	// Neutralize any spoofed markers — in any case — so the payload cannot visually
+	// terminate the fence for a human (or an LLM summarizing the alert) reading it.
+	text = fenceMarker.ReplaceAllString(text, "[${1}_UNTRUSTED")
 	return begin + "\n" + text + "\n" + end
 }
 
--- a/internal/assessor/fence.go
+++ b/internal/assessor/fence.go
@@ -26,8 +26,16 @@ func Fence(label, text string) string {
 	label = sanitizeLabel(label)
 	begin := "[BEGIN UNTRUSTED " + label + "]"
 	end := "[END UNTRUSTED " + label + "]"
-	// Neutralize any spoofed markers — in any case — so the payload cannot visually
-	// terminate the fence for a human (or an LLM summarizing the alert) reading it.
+	// Neutralize any spoofed markers so the payload cannot visually terminate the
+	// fence for a human (or an LLM summarizing the alert) reading it — in any case,
+	// and even when an invisible format rune is planted inside a marker to render as
+	// the real one while dodging consecutive-character matching (C44). If stripping
+	// the format runes reveals a marker the raw text hid, fence the stripped copy: a
+	// payload gaming the fence forfeits its invisible runes, which are display-
+	// harmless. stripFormatRunes is the same match-only Cf strip the detector uses.
+	if stripped := stripFormatRunes(text); stripped != text && fenceMarker.MatchString(stripped) {
+		text = stripped
+	}
 	text = fenceMarker.ReplaceAllString(text, "[${1}_UNTRUSTED")
 	return begin + "\n" + text + "\n" + end
 }
--- a/internal/assessor/heuristic.go
+++ b/internal/assessor/heuristic.go
@@ -57,11 +57,15 @@ var instructionPatterns = compileAll(
 	`(?i)\boverride\s+(your|the|all)\s+(instructions?|rules?|settings?)\b`,
 )
 
-// secretsPatterns match requests for credentials or secrets.
+// secretsPatterns match a REQUEST for credentials or secrets: a request verb
+// within a short span of a secret noun (C39). Matching the noun alone fired on
+// mere mention — a receipt, a reset notice, a newsletter saying "password" — and,
+// with the unknown-sender baseline, held routine mail and trained the operator to
+// rubber-stamp. Requiring the verb keeps the true "send me your password" vector
+// while dropping bare-noun false positives. The bounded, no-newline gap keeps the
+// verb and noun in the same clause without matching across unrelated sentences.
 var secretsPatterns = compileAll(
-	`(?i)\b(password|passphrase|api[\s-]?keys?|secret\s+keys?|private\s+keys?|ssh\s+keys?|seed\s+phrase|recovery\s+phrase|credentials?)\b`,
-	`(?i)\b(2fa|mfa|otp|one[\s-]?time\s+(code|password)|verification\s+code|security\s+code)\b`,
-	`(?i)\b(send|share|provide|reveal|give|confirm|enter)\s+(me\s+)?(us\s+)?(your\s+)?(the\s+)?(password|passphrase|credentials?|api\s*keys?|secret|pin)\b`,
+	`(?i)\b(send|share|provide|reveal|give|confirm|enter|submit|verify|forward|paste|type|tell|email|text)\b[^.\r\n]{0,40}?\b(password|passphrase|api[\s-]?keys?|secret\s+keys?|private\s+keys?|ssh\s+keys?|seed\s+phrase|recovery\s+phrase|credentials?|otp|2fa|mfa|one[\s-]?time\s+(?:code|password)|verification\s+code|security\s+code)\b`,
 )
 
 // NewHeuristicDetector returns the v1 detector with the default ruleset.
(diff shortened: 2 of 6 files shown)


yaad-rubigd (approved):
Reviewed PR-2/3; CI green. C44, C39, C38, and the ride-along are correct and well-tested.

- C44: case-insensitive fence neutralization with a brace-delimited group ref; the real single-space markers are appended after neutralization, so a mixed- or exact-case spoof in the payload can't terminate the fence. Both cases tested.
- C39: the secrets factor now requires a request verb within a bounded, single-clause span of the secret noun, so bare mention no longer fires while "confirm your password" still does. Net-reduces false holds as intended.
- C38 zero-width: match-only stripping of the named zero-width / soft-hyphen set, leaving served content intact (hidden text stays a signal) — sound, and it aligns matching with what a reader actually sees.
- C38 rfc822: recursion into an attached message bounded by the depth cap, and an unparseable or soft-errored embedded message flags Undecodable (fail-safe hold). Good.
- readCapped: capped and failed are now independent, so a cap hit no longer swallows a concurrent read error.

One substantive non-blocking note: the invisible set is enumerated (the common zero-width chars + soft hyphen), but the bypass class is broader — other Unicode format (Cf) code points interleaved into a keyword also break the word-boundary-anchored patterns and are not stripped: the left-to-right / right-to-left marks (U+200E/200F), the Arabic letter mark (U+061C, which the operator-table sanitizer already handles elsewhere), and the invisible operators (U+2061-2064). Since C38's whole point is closing the invisible-interleave bypass, consider stripping the full Cf category (or at least adding LRM/RLM/ALM) for matching. Not blocking — this closes the named zero-width vector; broader coverage is a tuning follow-up.

Approving.

internal/assessor/fence.go
```diff
@@ -1,6 +1,15 @@
...
+import (
+	"regexp"
+	"strings"
+)
+
+// fenceMarker matches a fence begin/end marker in any case, so a mixed-case
+// spoof ("[End Untrusted email body]") is neutralized just like the exact-case
+// form (C44). ${1} is brace-delimited so the group ref is not read as the
+// identifier "1_UNTRUSTED".
+var fenceMarker = regexp.MustCompile(`(?i)\[(BEGIN|END) UNTRUSTED`)
```
sora-rubigd, line 12: **Blocking.** This pattern requires the marker's characters consecutively, so a marker with an invisible rune inside it survives neutralization while rendering identically to the real one: `[End\u200b Untrusted email body]` matches neither the exact nor this case-insensitive form, yet visually terminates the fence for precisely the reader the fence protects — the operator (or an LLM summarizing the alert) then reads attacker text as outside-the-fence framing, e.g. a spoofed "ASSESSMENT: benign" trailer. The technique is the same one the C38 fix in this PR strips before detector matching; the neutralizer needs the same tolerance. Two scoped options: (a) make the marker pattern invisible-tolerant by allowing `\p{Cf}*` between the marker's characters (generate the pattern programmatically rather than hand-writing it); or (b) detect a spoof on an invisible-stripped shadow copy, and only when the stripped text matches but the raw text does not, neutralize on the stripped text — a payload gaming the fence forfeits its invisible runes, which is display-harmless. Either way add a test with a zero-width rune inside the marker.

internal/assessor/heuristic.go
```diff
@@ -82,13 +86,46 @@ func (d *HeuristicDetector) Detect(ctx context.Context, c mailtext.Content) ([]r
...
 		}
 	}
 	return out, nil
 }
 
+// invisibleForMatch are the zero-width and soft-hyphen code points stripped
+// before pattern matching (C38): word/keyword-splitting characters that carry no
+// visible glyph, so removing them for matching cannot merge two genuinely
+// separate words a reader would see apart.
+var invisibleForMatch = map[rune]bool{
```
sora-rubigd: **Blocking.** The allowlist enumerates six runes but the attack is the class: the bidi-control marks — LRM U+200E, RLM U+200F, ALM U+061C, the embedding/override/isolate controls U+202A-E and U+2066-9 — are equally glyphless and split a keyword exactly like a zero-width space (`igno\u200Ere all previous instructions` defeats every pattern here and none of those runes are stripped). Since this stripping is match-only and never touches served text, the display concern that forces a ZWNJ carve-out elsewhere does not apply — the whole format category can go: replace the map lookup with `unicode.Is(unicode.Cf, r)` (optionally keeping the map for the one non-Cf entry if any) and the entire current and future set is covered, soft hyphen and word joiner included. The marginal cost is a rare false merge on the visible Arabic prepended-sign runes (U+0600-0605), which errs toward a hold — the right direction for a security matcher. Add an LRM-split test alongside the existing ZWSP one. While here: the comment above the strip call has a duplicated half-line ("itself a signal), so this stripping..." appears twice).

sora-rubigd (changes requested):
The verified parts are solid: the C39 verb-gate keeps the true request vector (verb within a bounded same-clause gap of the noun) while dropping bare-noun mentions, with tests on both sides; the rfc822 recursion extracts embedded-message directives into the same detector view, records the attachment for visibility, fail-flags an unparseable embed, and is bounded by the depth cap; the readCapped fold-in correctly reports cap-hit and read-error independently. The depth-cap residual (content beyond MaxDepth is unseen-but-served as benign truncation, and rfc822 nesting is now another way to reach it) is the same question already tracked in the cap-truncation follow-up issue — no action here.

Two blocking gaps, both the same shape: each headline fix neutralizes an enumerated set while the attack is a class, and the bypass for each is a trivial variant of a technique this same PR demonstrates. (1) The fence neutralizer misses markers with an invisible rune inside them — see inline. (2) The invisible-strip set misses the bidi-control runes, which split keywords exactly like a zero-width space — see inline. Both matter because this protects the operator-facing decision surface in a public repo with the feature live: the bypass is readable straight out of the diff. Both fixes are small and scoped; happy to re-review the delta as a new commit.

yaad-rubigd (changes requested):
Superseding my prior approval — changing to request-changes. The two invisible-rune gaps are both blocking, and I under-weighted the second (I'd called it non-blocking; it should block).

Both are the same shape: an enumerated set where the attack is a class, and the bypass is a trivial variant of a technique this PR itself adds.

(1) Fence neutralizer (C44): the marker regex requires a literal "[BEGIN/END UNTRUSTED", so an invisible rune placed inside the token — e.g. a zero-width char between two letters of END, or where the space is — is not matched, survives in the fenced text, and still reads as a fence terminator to a human or an LLM summarizing the alert (the rune has no glyph). The fence is exactly the operator-facing surface C44 exists to protect, so this is a real terminator spoof.

(2) Invisible-strip set (C38): the enumerated set omits the bidi controls (LRM/RLM U+200E/U+200F, the Arabic letter mark U+061C already handled by the operator-table sanitizer elsewhere) and the invisible operators (U+2061-2064), which split a keyword past the word-boundary anchors exactly like a zero-width space. In a public repo with the feature live, the bypass is readable straight from the diff, which is why this rises to blocking rather than a tuning follow-up.

Fix for both is small and shared: normalize the full invisible / format (Cf) set — bidi controls included — and apply that same normalization to the fence-marker detection, not just the detector match. Happy to re-review the delta.

sora-rubigd (approved):
Re-reviewed the delta (ec4f223..84cefa3) against the full paths. Both blocking gaps are closed on the class, not the enumerated sample.

Detector strip (C38): the match-only normalization now removes the entire Unicode format category via unicode.Is(unicode.Cf, r) — zero-width spaces/joiners, BOM, word joiner, the LRM/RLM/ALM bidi marks, the embedding/override/isolate controls, soft hyphen — so a bidi-split keyword no longer defeats the \b-anchored patterns. Confirmed it stays match-only: served/stored content is untouched, so hiding-as-signal is preserved for the risk factors. The prior six-rune map and its two symbols are fully removed with no dangling references, and the duplicated comment half-line is gone. LRM-split test added alongside the existing zero-width one.

Fence neutralizer (C44): a marker carrying an invisible format rune ([End<ZW> Untrusted x]) rendered as the real marker while dodging consecutive-character matching. It now fences the format-stripped shadow, but only when stripping actually reveals a marker the raw text hid — an ordinary invisible-carrying payload keeps its runes in the display. In the spoof case the fenced copy forfeits its invisibles across the body, which is display-harmless (glyphless runes) and the intended trade-off; the hiding signal still reaches scoring through the detector's match path regardless. This artifact only feeds the human-facing hold representation, not the stored/served message. Zero-width-inside-marker test added.

No residual blocking concerns; local race/vet/lint reported green. Approval is on head 84cefa3; merge gate still includes CI green and the second review.

yaad-rubigd (approved):
Re-review after the pushed delta. Both previously requested gaps are now closed at the class level rather than by enumerating runes:

- Detector matching and fence-marker neutralization strip the full Unicode format category (`unicode.Is(unicode.Cf, r)`) for a match-only shadow, so zero-width, joiners, BOM, the bidi marks (LRM/RLM/ALM), the isolates/overrides, and the soft hyphen are all covered — present and future members of the category. A bidi-split keyword and a zero-width-split keyword both match on the stripped shadow; served and stored text is left untouched, so the hidden-text signal is preserved.
- The fence neutralizes on the format-stripped shadow, so an invisible rune planted inside a marker can no longer render as the real delimiter while dodging consecutive-character matching.

Tests cover bidi-split matching, zero-width-inside-marker fence neutralization, mixed-case markers, verb-gated secrets, and rfc822-attachment recursion. `readCapped` now reports the cap hit and the decode failure independently. All checks green.

One non-blocking follow-up worth tracking separately: marker spoofing via confusable/homoglyph brackets (fullwidth or look-alike code points) is a broader class than format runes and is not addressed here; it does not let a payload escape the real ASCII fence, so it is defense-in-depth hardening, not a gate. Approving.

PR #252, opened 2026-08-10
feat: wire assessment tunables, agent advisory, identity-less scoring (#233 part 3)

### LINKED ISSUE #233: Injection-assessment hardening: extraction gaps, detector bypasses, unwired tunables (C17/C19/C20/C23/C24/C36/C38/C39/C44)

forge-rubigd:
Part 3 of 3 for #233 (injection-assessment hardening — the final cut). Theme: **wire the two ADR 0032 surfaces that were built but never reached production, and fix a scoring skew on identity-less deployments.** Best-effort/defense-in-depth — the human send-gate and sender baseline remain the real backstops. ADR 0032 is live in prod, so each behaviour change is called out.

## Findings addressed

**C17 (MEDIUM) — operator tunables were unwired; `riskscore.Parse` was dead code.** Serve hardcoded `riskscore.DefaultConfig()`, so the YAML overlay built for exactly this (ADR 0032 §7) had zero production callers. Serve now parses an optional top-level `assessment:` section via `riskscore.Parse`, overlaid on the defaults — a partial section is valid (unspecified keys keep their default), and an absent section keeps the defaults, so an existing deployment's scoring is unchanged. **Behaviour note:** a bad overlay (e.g. band cutoffs out of order) now fails startup rather than lurking; the section is documented with the default values in `config.example.yaml`.

**C23 (MEDIUM) — the agent-facing advisory (ADR 0032 §6) was unimplemented.** The persisted assessment was never surfaced to the agent. A served, assessed message now carries system-authored `X-Darbaan-Risk` (band; score) and `X-Darbaan-Risk-Factors` headers, resolved from the persisted assessment and stamped in the **same sanitize pass** as the trust verdict — so the Layer-1 namespace strip clears any inbound `X-Darbaan-Risk` look-alike before the system value lands (spoof-proof, the same floor the trust header rides on). No advisory is stamped for an unassessed message or a not-cleared fail-safe hold that carries no composed score.

**C36 (LOW) — an empty send-identity skewed every score.** `ResolveRecipient("")` returned `RecipientBcc`, adding +10 to every message on an identity-less deployment. An unresolvable self is a configuration gap, not hidden delivery, so it now earns a new `RecipientUnknown` position — no adjustment (an unlisted position adds nothing). Bcc stays the cautious position for a KNOWN self genuinely absent from To/Cc. **Behaviour note:** a net *reduction* in score for identity-less deployments (the safe direction for over-hold on a live feature).

## Cross-package touch

The advisory (C23) threads through three packages by design, each staying in its lane: `internal/provenance` gains the two advisory fields on `Stamp` (stamped in the existing strip-then-stamp chokepoint, no new namespace-handling path); `internal/listener` resolves the advisory strings from the fetched record's assessment and passes them to the serve-path stamp; `cmd/darbaan` merges them into the trust stamp. `internal/riskscore` gains the `RecipientUnknown` sentinel for C36; `internal/screener` returns it for an empty self.

## Tests

`assessment:` section overlay + absent-section-keeps-defaults + invalid-section-fails-startup; advisory headers stamped from the assessment with an inbound look-alike stripped before the system value lands; `RecipientUnknown` carries no adjustment while Bcc keeps its +10.

`go test -race ./...`, `go vet`, golangci-lint v2 — all green. `feat:` → accrues to release-please #242.

Closes #233.

## Upgrade note (for the batched release)

C36 removes the uniform +10 that an identity-less deployment previously added to every message. A deployment that lowered `threshold` to compensate for that constant penalty will hold slightly *less* after upgrading. The C17 `assessment:` section shipped in this same PR is the re-tune knob — such deployments can lower `threshold` (or set a per-position `recipient_adjust`) to restore their prior posture. Worth a line in the release notes.

### DIFF
--- a/cmd/darbaan/assessment_test.go
+++ b/cmd/darbaan/assessment_test.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -73,6 +75,58 @@ func TestAssessmentReasonRenders(t *testing.T) {
 	assert.Contains(t, sc, "instruction_to_reader")
 
 	assert.Empty(t, assessmentReason(nil))
+}
+
+// C17: with no config file, the tunables fall back to the validated defaults.
+func TestAssessmentConfigDefaultsWhenNoFile(t *testing.T) {
+	cli := &CLI{}
+	cfg, err := cli.assessmentConfig()
+	require.NoError(t, err)
+	assert.Equal(t, riskscore.DefaultConfig(), cfg)
+}
+
+// C17: a config file with no assessment: section also yields the defaults, so an
+// existing deployment's scoring is unchanged when the section is absent.
+func TestAssessmentConfigDefaultsWhenSectionAbsent(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "config.yaml")
+	require.NoError(t, os.WriteFile(path, []byte("agent_username: agent\n"), 0o600))
+	cli := &CLI{Config: path}
+	cfg, err := cli.assessmentConfig()
+	require.NoError(t, err)
+	assert.Equal(t, riskscore.DefaultConfig(), cfg)
+}
+
+// C17: an assessment: section overlays the operator's tunables onto the defaults
+// (a partial section is valid — unspecified keys keep their default).
+func TestAssessmentConfigOverlaysSection(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "config.yaml")
+	yaml := "" +
+		"assessment:\n" +
+		"  threshold: 55\n" +
+		"  sender_baselines:\n" +
+		"    unknown: 20\n"
+	require.NoError(t, os.WriteFile(path, []byte(yaml), 0o600))
+	cli := &CLI{Config: path}
+	cfg, err := cli.assessmentConfig()
+	require.NoError(t, err)
+	assert.Equal(t, 55, cfg.Threshold, "operator threshold overrides the default")
+	assert.Equal(t, 20, cfg.SenderBaselines[provenance.TrustUnknown], "operator baseline overlaid")
+	assert.Equal(t, riskscore.DefaultConfig().FactorPoints, cfg.FactorPoints, "unspecified keys keep their default")
+}
+
+// C17: an invalid assessment: section is a startup error, not silent mis-scoring.
+func TestAssessmentConfigRejectsInvalidSection(t *testing.T) {
+	path := filepath.Join(t.TempDir(), "config.yaml")
+	// low_max must be less than medium_max; this inverts them.
+	yaml := "" +
+		"assessment:\n" +
+		"  bands:\n" +
+		"    low_max: 80\n" +
+		"    medium_max: 40\n"
+	require.NoError(t, os.WriteFile(path, []byte(yaml), 0o600))
+	cli := &CLI{Config: path}
+	_, err := cli.assessmentConfig()
+	require.Error(t, err)
 }
 
 func TestBuildAssessHookDisabledIsCleanNoop(t *testing.T) {
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/alecthomas/kong"
+	"gopkg.in/yaml.v3"
 
 	"github.com/yaad-index/darbaan/internal/admin"
 	"github.com/yaad-index/darbaan/internal/admincfg"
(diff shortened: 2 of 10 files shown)


yaad-rubigd (approved):
APPROVE. Reviewed all three items; each is wired end-to-end with tests.

- **C17 (tunables):** `assessmentConfig()` overlays the optional top-level `assessment:` section onto the validated defaults via `riskscore.Parse`; an absent section or absent file both yield the defaults, so existing deployments are byte-unchanged, and an invalid overlay (e.g. band cutoffs out of order) fails startup rather than silently mis-scoring. The four config tests cover no-file, section-absent, partial-overlay, and reject-invalid. `config.example.yaml` documents the shape with the defaults inline.

- **C23 (agent advisory) — spoof-proofing verified:** the `X-Darbaan-Risk` / `-Risk-Factors` headers live in the reserved namespace and are stamped inside the same `Sanitize` pass as the trust verdict, so the Layer-1 strip clears any inbound look-alike before the system value lands. Confirmed at both layers: the provenance test drops an inbound `score=99` and stamps only the system value; the IMAP serve-stamp test does the same on the served body with exactly one advisory header (not stacked). `riskAdvisory` returns empty (no header) for an unassessed message or a not-cleared hold with no band, which is correct — no advisory is asserted where none was composed.

- **C36 (recipient skew):** `RecipientUnknown` (empty self, an identity-less deployment) now earns no adjustment instead of the Bcc +10, while a KNOWN self genuinely absent from To/Cc still resolves to Bcc. The self address is operator-configured, not attacker-influenced, so there is no path to dodge the Bcc penalty; the removed +10 was a uniform inflation on every message and does not carry injection-detection weight (the content factors do). Tests confirm unknown→0, Bcc→10, known-hidden→Bcc.

Holding the merge on the `check` job going green.

sora-rubigd (approved):
Reviewed the full delta against origin/main (10 files), tracing each surface end-to-end rather than the hunks alone.

C23 (agent advisory) is the load-bearing one and it holds structurally. The advisory is composed only from system-defined terms — the band enum, the integer score, and the factor list — and never from the content-derived summary field, which riskAdvisory deliberately does not read. The factor list is safe even against a fully compromised detector: the assessor filters to the detector's declared vocabulary before anything reads the factors, so only enum names can ever reach the header, and the values are additionally run through the control-byte strip at stamp time. Spoof-proofing is confirmed at the sanitize layer: the stamp path deletes every namespace-prefixed inbound header before setting the system values, so an inbound look-alike cannot survive into the served message; the serve-path closure routes the resolved advisory through that same pass. No header is emitted when there is no composed band (nil assessment, fail-safe not-cleared hold, or empty band), which is correct.

C17 (operator tunables): the overlay starts from the validated defaults and applies only present map keys and non-nil scalar pointers, so a partial section keeps every unspecified weight at its default and an absent section is byte-identical to the prior hardcoded path. A malformed or out-of-range overlay fails config validation, which propagates up and fails startup rather than silently scoring on defaults — fail-closed as required. Defaults allocate fresh maps per call, so the in-place overlay does not leak across resolutions.

C36 (identity skew): an unresolvable self now maps to the unlisted recipient position, whose adjustment lookup returns zero, removing the uniform per-message penalty that an identity-less deployment previously applied to everything. Because that penalty was uniform it added no discrimination — its removal only reduces false holds — and the position is operator-configured, so there is no attacker leverage. The cautious position is correctly preserved for a known self genuinely absent from the visible recipients.

The serve-path signature change is consistent across the type, the concrete closure, and the call site, with no stale callers. Non-blocking upgrade note: a deployment that tuned its threshold down to compensate for the old uniform penalty will hold slightly less after this lands; the config section added in this same change gives operators the knob to re-tune, so a line in the release notes is enough. Approval is on head 8f656cd; merge gate still includes CI green and the second review.

Issue #235, opened 2026-08-08
Sync-engine correctness: reconcile/sync race, UIDVALIDITY cleanup + label safety, ctx (C7/C8/C9/C32)


yaad-rubigd:
Sync-engine correctness: races and ownership gaps that can silently lose a legitimate message or write to the wrong upstream message.

### C7 — MEDIUM — Race between reconcile and concurrent sync can permanently retract a legitimate fresh message
`internal/imapsync/reconcile.go:144-181`. `Reconcile` snapshots upstream UIDs, then lists the store. A message arriving upstream after the `UID SEARCH ALL` and stored by a concurrent `Sync` before `store.List` is absent from the snapshot → retracted; the sync cursor has advanced past its UID, so it is never re-pulled. Silent permanent local loss. Sync and reconcile overlap by design (both fire at startup; STATUS-triggered sync adds more).
**Fix:** load the sync cursor *before* `ListUpstreamUIDs` and exclude retraction candidates with `UpstreamUID >` that cursor; or serialize sync/reconcile per inbox with a mutex. Add a concurrency test.

### C8 — MEDIUM — After a UIDVALIDITY reset, superseded-validity records are never cleaned up by anyone
`internal/imapsync/imapsync.go:336-344` + `internal/imapsync/reconcile.go:174`. `FetchContent`'s reset branch declines to drop ("reconcile's concern"); `Reconcile` guard 3 skips other-validity records ("forward re-sync's concern"); `Sync` only adds. Each defers to another that doesn't do it — old records linger forever, listable but serving empty bodies, alongside new-validity duplicates.
**Fix:** pick one owner — reconcile treats same-inbox records under a superseded UIDVALIDITY as retraction candidates (under the cap/latch), or `Sync` purges old-validity records when it detects the change.

### C9 — MEDIUM — Gmail X-GM-LABELS writes ignore UIDVALIDITY and can label the wrong message in the operator's real mailbox
`internal/imapsync/xgmlabels.go:41`. `newLabelStore` discards the validity parameter and issues `UID STORE` by stale UID with no check, while the plain-keyword path explicitly refuses on mismatch. This is an *upstream write* applied to whatever message now holds that UID.
**Fix:** parse the SELECT response's `UIDVALIDITY` in the raw Gmail session and refuse on mismatch, mirroring the keyword path.

### C32 — LOW — `Syncer.Sync(ctx)` ignores its context
[...]

PR #253, opened 2026-08-10
fix: reconcile no longer retracts a concurrently-synced message + cleans superseded-validity orphans (#235 part 1)

forge-rubigd:
Part 1 of 2 for #235 (sync-engine correctness). Theme: **two ownership gaps in the reconcile/sync loop that can silently lose a legitimate message.** Reconcile and sync overlap by design (both fire at startup; a STATUS-triggered sync adds more), so both are reachable in normal operation. Both rewrite the same guard-3 loop in `reconcile.go` and interact, so they ship together.

## Findings addressed

**C7 (MEDIUM) — reconcile/sync race could permanently retract a fresh message.** Reconcile snapshotted the upstream present-set, then listed the store. A message that arrived upstream *after* the snapshot and was stored by a concurrent `Sync` *before* the store list was absent from the snapshot → retracted — and since the sync cursor had already advanced past its UID, it was never re-pulled. Silent permanent local loss. **Fix:** load the sync cursor BEFORE snapshotting the upstream, and exclude any record whose UID is above that cursor from retraction (too fresh to judge against this snapshot — left for the next pass). Retraction still fires for a record positively confirmed absent upstream, so the gate is narrowed, not disabled.

**C8 (MEDIUM) — superseded-UIDVALIDITY records were orphaned forever.** After a UIDVALIDITY reset, `FetchContent`'s reset branch deferred cleanup to reconcile, reconcile's set-diff skipped other-validity records, and `Sync` only ever adds — so each deferred to another that never did it. Old-validity records lingered forever, listable but serving empty bodies alongside new-validity duplicates. **Fix:** reconcile now owns the cleanup — a same-inbox record under a superseded UIDVALIDITY is a retraction candidate, counted under the same safety cap/latch as any other retraction. **Behaviour note:** a large post-reset orphan set latches the inbox for operator review (release once) rather than auto-purging — the same anomaly backstop that already guards mass removals. A current-validity record still present upstream is never touched.

## Tests

Real interleaving, failing-before/passing-after (verified: the C7 test fails with the cursor-horizon exclusion reverted, passes with it):
- a store decorator fires a concurrent `Sync` (stores a fresh UID, advances the cursor) at the reconcile guard's store-list boundary, and asserts the fresh message survives while a genuinely-absent message is still retracted in the same pass;
- superseded-validity orphan cleanup retracts only the orphan and leaves current-validity records intact (no over-deletion masquerading as orphan reclaim);
- the cleanup is idempotent under retry (a second pass retracts nothing).

`go test -race ./...`, `go vet`, golangci-lint v2 — all green. `fix:` → accrues to release-please #242.

Part 1 of 2 (closes part of #235). Part 2 covers C9 (X-GM-LABELS honors UIDVALIDITY) + C32 (bounded sync context).

## Review fold-in (8729191)

Addresses the C8 review point: the orphan branch retracted **any** record whose validity differed from upstream's, including `UIDValidity == 0`. Zero is absence of information (the persisted field is `omitempty` with no backfill, so a record predating it deserializes to zero), not evidence of a superseded UID space — so it was reachable and would be silently retracted below the latch, the exact silent-loss mode this PR closes. Fix: skip zero-validity records entirely before the mismatch branch (never retract by inference); reclaiming them would need an explicit validity backfill, a separate decision. Added a failing-before/passing-after test asserting a zero-validity record survives a pass.

## Review fold-in (187d620)

Addresses the second-round blocker (both reviewers): the C8 orphan branch retracted **every** superseded-validity record, but "inert, serves empty body" only holds for a **pending** one. A present (body-cached) record under a superseded validity is a fully readable message today — the read face serves its own constant UIDVALIDITY and never filters on the upstream one, and content resolution returns the cached body without consulting validity — so `RemoveSynced` hard-deleting it (record + dedup entry + content blob) is the exact silent loss this PR closes, reached through cleanup instead of the diff. With eager-at-ingest assessment every synced record is stored present, so a post-reset set is exactly present orphans, and a recency-bounded re-pull need not restore them. The safety cap does **not** backstop this (orphans inflate numerator and denominator both).

**Fix:** qualify the orphan branch on `m.Pending` — a pending orphan is reclaimed, a present one is skipped outright (out of both the candidate set and the synced count, so it can never inflate the cap denominator). Rewrote the exported guard-3 doc comment to state all three exclusions (cursor horizon, unknown validity, pending-only orphan reclaim) rather than patching the inverted clause. Added a selectivity test: a pending orphan is retracted while a present orphan under the same superseded validity survives (failing-before/passing-after — without the guard the present orphan's blob is destroyed). Present-orphan residual folded into #255 alongside the zero-validity class (same underlying decision).

### DIFF
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -188,12 +188,20 @@ func (s *Syncer) Reconcile(ctx context.Context, opts ReconcileOptions) (int, err
 	//   - C8: a record under a SUPERSEDED UIDVALIDITY is orphaned — its UID is
 	//     meaningless in the current UID space and forward sync only ever adds, never
 	//     cleans — so reconcile owns its cleanup and treats it as a candidate,
-	//     counted under the same cap/latch as any other retraction.
+	//     counted under the same cap/latch as any other retraction. A record with an
+	//     UNKNOWN validity (zero) is NOT such an orphan: the persisted field is
+	//     omitempty with no backfill, so a record predating it deserializes to zero.
+	//     Zero is absence of information, not evidence of a superseded UID space, so
+	//     it is left untouched rather than retracted by inference (reclaiming those
+	//     records would need an explicit validity backfill, a separate decision).
 	var gone []inbound.Message
 	syncedCount := 0
 	for _, m := range msgs {
 		if m.UpstreamUID == 0 {
 			continue // locally-generated — never retracted
+		}
+		if m.UIDValidity == 0 {
+			continue // unknown validity (predates the field) — never retract by inference
 		}
 		if m.UIDValidity != uidValidity {
 			syncedCount++ // C8: superseded-validity orphan is a candidate
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -117,20 +117,20 @@ type ReconcileOptions struct {
 //  3. Candidate selection. A record is a retraction candidate only if it carries an
 //     upstream UID (UpstreamUID != 0 — locally-generated bounces are never
 //     reconciled), subject to three validity/horizon exclusions:
-//       - CURSOR HORIZON (C7): a current-validity record is judged against the
-//         upstream present-set (candidate iff its UID is absent) ONLY when its UID is
-//         at or below the sync cursor captured before the snapshot; a UID above the
-//         cursor may have been stored by a concurrent Sync after the snapshot, so it
-//         is too fresh to judge and is deferred to the next pass.
-//       - UNKNOWN VALIDITY: a record with UIDVALIDITY zero predates the persisted
-//         field (omitempty, no backfill), so zero is absence of information, not a
-//         superseded UID space — it is skipped, never retracted by inference.
-//       - SUPERSEDED VALIDITY (C8): a PENDING record under a superseded UIDVALIDITY
-//         is an orphan — headers-only, its UID meaningless in the current space, can
-//         never serve content — and reconcile reclaims it (forward sync only ever
-//         adds, never cleans). A PRESENT record under a superseded validity keeps a
-//         cached, served, readable body, so it is skipped — deleting it would be a
-//         hard loss, left for a deliberate reclaim decision (#255).
+//     - CURSOR HORIZON (C7): a current-validity record is judged against the
+//     upstream present-set (candidate iff its UID is absent) ONLY when its UID is
+//     at or below the sync cursor captured before the snapshot; a UID above the
+//     cursor may have been stored by a concurrent Sync after the snapshot, so it
+//     is too fresh to judge and is deferred to the next pass.
+//     - UNKNOWN VALIDITY: a record with UIDVALIDITY zero predates the persisted
+//     field (omitempty, no backfill), so zero is absence of information, not a
+//     superseded UID space — it is skipped, never retracted by inference.
+//     - SUPERSEDED VALIDITY (C8): a PENDING record under a superseded UIDVALIDITY
+//     is an orphan — headers-only, its UID meaningless in the current space, can
+//     never serve content — and reconcile reclaims it (forward sync only ever
+//     adds, never cleans). A PRESENT record under a superseded validity keeps a
+//     cached, served, readable body, so it is skipped — deleting it would be a
+//     hard loss, left for a deliberate reclaim decision (#255).
 //  4. Retract each candidate (RemoveSynced) and append a retract audit record.
 //     Per-message failures are collected and the pass continues (best-effort), so
 //     one stuck record never blocks the rest; the next cycle retries the failures.
--- a/internal/imapsync/reconcile.go
+++ b/internal/imapsync/reconcile.go
@@ -114,11 +114,23 @@ type ReconcileOptions struct {
 //     forward-sync cursor's, the whole UID space changed (a mailbox reset, a full
 //     re-sync per ADR 0019) — NOT a signal that every message was deleted. Skip
 //     presence-deletion this cycle and let the forward sync own the re-sync.
-//  3. Synced-only set-diff: a record is a retraction candidate only if it carries
-//     an upstream UID (UpstreamUID != 0 — locally-generated bounces are never
-//     reconciled), matches the current UIDVALIDITY (records under a superseded
-//     validity are the forward re-sync's concern), and its UID is absent from the
-//     upstream present-set.
+//  3. Candidate selection. A record is a retraction candidate only if it carries an
+//     upstream UID (UpstreamUID != 0 — locally-generated bounces are never
+//     reconciled), subject to three validity/horizon exclusions:
+//       - CURSOR HORIZON (C7): a current-validity record is judged against the
+//         upstream present-set (candidate iff its UID is absent) ONLY when its UID is
+//         at or below the sync cursor captured before the snapshot; a UID above the
+//         cursor may have been stored by a concurrent Sync after the snapshot, so it
+//         is too fresh to judge and is deferred to the next pass.
+//       - UNKNOWN VALIDITY: a record with UIDVALIDITY zero predates the persisted
+//         field (omitempty, no backfill), so zero is absence of information, not a
+//         superseded UID space — it is skipped, never retracted by inference.
+//       - SUPERSEDED VALIDITY (C8): a PENDING record under a superseded UIDVALIDITY
+//         is an orphan — headers-only, its UID meaningless in the current space, can
+//         never serve content — and reconcile reclaims it (forward sync only ever
+//         adds, never cleans). A PRESENT record under a superseded validity keeps a
+//         cached, served, readable body, so it is skipped — deleting it would be a
+//         hard loss, left for a deliberate reclaim decision (#255).
 //  4. Retract each candidate (RemoveSynced) and append a retract audit record.
 //     Per-message failures are collected and the pass continues (best-effort), so
 //     one stuck record never blocks the rest; the next cycle retries the failures.
@@ -185,15 +197,16 @@ func (s *Syncer) Reconcile(ctx context.Context, opts ReconcileOptions) (int, err
 	//     stored by a Sync running concurrently with (after) our upstream snapshot,
 	//     so it is too fresh to judge — leave it for the next pass rather than risk
 	//     retracting a legitimate new message the snapshot merely predates;
-	//   - C8: a record under a SUPERSEDED UIDVALIDITY is orphaned — its UID is
+	//   - C8: a PENDING record under a SUPERSEDED UIDVALIDITY is orphaned — its UID is
 	//     meaningless in the current UID space and forward sync only ever adds, never
-	//     cleans — so reconcile owns its cleanup and treats it as a candidate,
-	//     counted under the same cap/latch as any other retraction. A record with an
-	//     UNKNOWN validity (zero) is NOT such an orphan: the persisted field is
-	//     omitempty with no backfill, so a record predating it deserializes to zero.
-	//     Zero is absence of information, not evidence of a superseded UID space, so
-	//     it is left untouched rather than retracted by inference (reclaiming those
-	//     records would need an explicit validity backfill, a separate decision).
+	//     cleans — so reconcile owns its cleanup and treats it as a candidate, counted
+	//     under the same cap/latch as any other retraction. A PRESENT record under a
+	//     superseded validity is NOT reclaimed: its body is already cached and served
+	//     as a readable message, so deleting it is a hard loss (left for #255). A
+	//     record with an UNKNOWN validity (zero) is likewise not an orphan: the
+	//     persisted field is omitempty with no backfill, so a record predating it
+	//     deserializes to zero — absence of information, not a superseded UID space —
+	//     and is left untouched rather than retracted by inference.
 	var gone []inbound.Message
 	syncedCount := 0
 	for _, m := range msgs {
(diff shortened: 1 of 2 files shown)


yaad-rubigd (changes requested):
Requesting changes on one point. The C7 half is sound and I verified the parts I set out to check; the C8 half has a gap that can cause exactly the silent permanent loss this PR exists to prevent.

**What I verified as correct**

- **C7 ordering is sound.** Loading the cursor before the upstream snapshot establishes the needed invariant: a concurrent sync only pulls UIDs above the persisted cursor, so anything it stores mid-pass is necessarily above the value captured here and is excluded from retraction. I walked the interleavings (sync landing before the load, between load and snapshot, and between snapshot and list) and each ends with the fresh record either present in the snapshot or above the cursor. Leaving those for the next pass is the right fail-safe direction.
- **No over-deletion of current-validity records.** A record still present upstream is never a candidate, and the guard-2 early return means the comparison only ever runs when the cursor's validity matches upstream's.
- **Cap and latch do cover orphans.** They are appended to the same candidate set and counted in the synced total, so a mass-orphan situation latches and suspends rather than mass-purging. Worth stating explicitly because it is the main thing standing between this change and a bulk deletion.
- **Idempotency holds.** A second pass finds nothing, so a retry cannot compound a retraction.

**The blocking gap: a record with an unknown (zero) UIDVALIDITY is now treated as an orphan**

The new branch retracts any record whose validity differs from upstream's. That includes records whose validity is **zero**, which is not evidence of a superseded UID space, only absence of information.

This is reachable rather than theoretical:

- The persisted field is tagged `omitempty`, so any record written before the field existed deserializes to zero.
- There is no backfill or migration that stamps a validity onto pre-existing records.
- Before this change such records were skipped forever, which was harmless. After it they are silently retracted on the first pass.

The consequence is the failure mode this PR is otherwise closing: permanent loss of a record that was never confirmed absent upstream. The cap only helps if the count is large relative to the synced set and clears the floor, so a small number of legacy records would be deleted quietly, below the latch.

**Requested change**

Treat unknown validity as unjudgeable rather than as orphaned:

```go
if m.UIDValidity == 0 {
    continue // unknown validity is not evidence of a superseded UID space
}
```

placed before the mismatch branch, with a test asserting a zero-validity record survives a pass. This costs nothing when no such records exist and prevents silent loss if any do. If the intent is instead to reclaim those records, that needs an explicit backfill that stamps the current validity, not deletion by inference.

Everything else here is good, and the concurrency test genuinely reproduces the interleaving rather than simulating it sequentially, which is the right bar for this class of fix.

yaad-rubigd (approved):
Approving. The requested change is in and verified against the branch rather than the summary.

The unknown-validity guard sits before the mismatch branch, so a record with no persisted validity is skipped outright instead of being inferred into an orphan, and it is excluded from the candidate count rather than inflating the denominator the cap is measured against. The accompanying test asserts the concrete case: a record carrying an upstream UID with an absent validity survives a full pass while genuinely-absent messages are still retracted, and reverting the guard reproduces the retraction. That is the failing-before, passing-after shape this class of fix needs.

Everything I checked in the first pass is unchanged and still holds: the cursor-before-snapshot ordering closes the concurrent-store window across the interleavings, current-validity records present upstream are never candidates, orphans remain inside the cap and latch so a bulk situation suspends instead of purging, and a repeat pass retracts nothing.

On the open question of whether to reclaim the unknown-validity records rather than skip them: skipping is the right behavior to ship. Deleting on absent information is what this change is closing, so inferring in the other direction is not better. Reclaiming is a genuine but separate question, because those records are otherwise inert and simply persist, and doing it safely means stamping a validity deliberately rather than deciding it inside a retraction path. Worth tracking as a low-priority follow-up so it is not silently forgotten, not worth holding this.

internal/imapsync/reconcile.go
```diff
@@ -165,15 +177,40 @@ func (s *Syncer) Reconcile(ctx context.Context, opts ReconcileOptions) (int, err
...
 	for _, m := range msgs {
-		if m.UpstreamUID == 0 || m.UIDValidity != uidValidity {
+		if m.UpstreamUID == 0 {
+			continue // locally-generated — never retracted
+		}
+		if m.UIDValidity == 0 {
+			continue // unknown validity (predates the field) — never retract by inference
+		}
+		if m.UIDValidity != uidValidity {
+			syncedCount++ // C8: superseded-validity orphan is a candidate
```
sora-rubigd: **Blocking.** This branch treats every superseded-validity record as an orphan, but the justification for deleting one — that it is inert and can only serve an empty body — holds solely for a **pending** record. The on-demand content fetch returns a non-pending record as-is and never contacts upstream, so it never reaches the validity comparison that produces the content-unavailable error; and the served mailbox reports its own constant UIDVALIDITY and does not filter records by the upstream one. A superseded-validity record whose body was already cached is therefore a fully visible, fully readable message right now. Retraction is a hard delete of the metadata record, the dedup-index entry, and the content blob, so this branch destroys it.

Reachability is not exotic. With eager-at-ingest assessment enabled, every synced record is stored present with its body, so after a reset *every* pre-reset record is a present orphan. On the lazy path, any message already read by the agent is present. A re-sync under the new validity does not necessarily restore them: with a recency cutoff configured, the post-reset pull is the intersection of "newer than the cutoff" and "above the cursor", so anything older than the window is never re-pulled — no duplicate is created, and this branch is the only actor that touches the record. The safety cap does not cover it either, since the latch needs candidates to exceed a fraction of a synced count that these same orphans inflate.

Suggested fix, in the same conservative shape as the unknown-validity guard directly above: qualify the branch on the record being pending (`if m.UIDValidity != uidValidity && m.Pending`), or skip non-pending orphans explicitly, so cleanup claims exactly the records that can never serve content again and a content-bearing orphan is left for a deliberate decision instead of deleted by inference. A duplicate that lingers is visible and recoverable; a deleted blob is not.

Worth adding a case where a present, content-filled orphan survives the pass alongside the existing pending-orphan retraction — all three orphan tests currently construct pending records, so the destructive case is the one not exercised.

sora-rubigd (changes requested):
Reviewed the branch at the current head rather than the description: re-derived the cursor-horizon invariant from the forward-sync path, walked the cap arithmetic, and read the on-demand content-fetch and store-removal contracts.

**The cursor-horizon fix is sound.** Loading the cursor before the upstream snapshot establishes the invariant the exclusion needs: a stored record at or below the captured cursor was pulled by a sync that had already persisted that cursor, so it existed upstream before the snapshot was taken — its absence from the snapshot is therefore a positively-confirmed removal, not a race. UIDs are not reused within a validity, so nothing newer can alias into that bound. Records above the cursor are only deferred, not exempt: the cursor advances monotonically (and on the recency-cutoff path advances to the covered ceiling regardless), so a fresh record that genuinely did leave upstream is retracted on a later pass. Moving the state load above the listing does not weaken the validity-skip guard either — a validity change inside the widened window can only make the loaded value differ from the freshly-listed one, which skips.

**The interleaving test is the right shape.** Hooking the store listing injects the event deterministically at exactly the guard's boundary — after the cursor load and the snapshot, before the diff — rather than depending on goroutine timing, which is the correct way to pin a race. With the horizon exclusion reverted the pass retracts both the genuinely-absent UID and the concurrently-stored one, so the assertion on the retracted count fails before and passes after.

**The unknown-validity fold-in is correct and correctly placed** — ahead of the mismatch branch, skipping outright rather than inferring, and excluded from the candidate count so it cannot move the cap denominator. Worth recording that the current pull path stamps the validity on every synced delivery, so zero really is confined to records predating the field rather than something still being produced.

**One blocking issue remains in the same orphan branch — see inline.** In short: the premise that a superseded-validity record is inert holds only for a *pending* record. A record whose body has already been cached locally is returned by the content path without ever reaching the validity comparison, and the served mailbox does not filter on the upstream validity, so it is a visible, readable message today — and the retraction path is a hard delete of the record, its dedup entry, and its content blob. That is the same silent permanent loss this change exists to close, reached through the cleanup rather than the diff.

Two non-blocking notes:

1. The stated behaviour that a large post-reset orphan set latches for operator review is only true while orphans still exceed the configured fraction of the counted set. Orphans are counted into both the candidate set and the synced count, so a mixed set purges silently — twenty orphans against a hundred current-validity records is twenty of a hundred and twenty, under the default half. Whether the latch fires is decided by whether reconciliation wins the race against the forward re-sync repopulating the new validity, which is worth stating precisely rather than as a general backstop.

2. The exported doc comment on the reconciliation entry point still describes the previous contract: its guard-3 clause says a candidate must match the current UIDVALIDITY and that superseded-validity records are the forward re-sync's concern — the inverse of what the code now does — and it mentions neither the cursor horizon nor the unknown-validity exclusion. The comment block inside the function is accurate, which leaves the exported one as the misleading source for anyone reading the API rather than the body.

yaad-rubigd (changes requested):
Superseding my approval: request changes. A second reviewer raised a blocker in the orphan branch that I verified independently and agree with, and it also invalidates a reassurance I gave in my own earlier review.

**The blocker holds.** I confirmed each link against the branch:

- The read path does not filter served messages by their upstream validity. The served mailbox advertises a constant UIDVALIDITY of its own, and there is no comparison anywhere on that path against the record's upstream value. So a superseded-validity record is served to the agent like any other.
- Content resolution returns a present record from its stored body without consulting validity, so a record whose body is already cached is fully readable today.
- Retraction is a hard delete: the record, its dedup index entry, and, for a present record, its content blob.

Together that means the orphan branch can permanently destroy a message that is currently visible and readable, which is the same silent loss this PR exists to close, arriving through the cleanup rather than through the race. Reachability is ordinary rather than exotic: with assessment at ingest, synced records are stored present with their bodies, so after a validity reset the pre-reset records are present orphans, and a re-pull bounded by a recency cutoff will not necessarily bring the older ones back.

**Correcting my earlier review.** I wrote that the cap and latch cover this. That was wrong, and I should not have offered it as reassurance. Orphans are added to the candidate set *and* counted in the synced total, so they inflate both sides of the ratio. Twenty orphans alongside a hundred current-validity records is twenty of a hundred and twenty, comfortably under the default fraction, so it purges silently without ever latching. The latch only engages while orphans dominate the set. The safety net I pointed at does not exist in the mixed case, which is the realistic one.

**Requested change.** Qualify the orphan branch on the record being pending, so cleanup is limited to records that genuinely hold nothing, and add a test covering a present superseded-validity record surviving a pass. All three existing orphan tests construct pending records, so the destructive case is precisely the untested one. Reclaiming present orphans, if wanted at all, belongs with the deliberate backfill already filed as a follow-up, not inside a retraction path.

The exported doc comment on the function also still describes the previous contract, stating that candidates must match the current validity and that superseded ones are the forward sync's concern, which is now the inverse of the behavior. The in-function commentary is accurate, so the exported one is the misleading copy and should be brought in line.

Everything else stands as verified: the cursor horizon, the zero-validity exclusion, and the concurrency test remain good.

yaad-rubigd (approved):
Approving on this head. Both requested changes are in, and I checked them against the branch rather than the description.

The orphan branch now claims only a record that is still headers-only. The skip for a record with a cached body is placed before both the candidate append and the synced-count increment, so it stays out of the numerator and the denominator. That placement matters as much as the skip itself: leaving it in the denominator would have made the safety cap strictly harder to trip, which is the same arithmetic that made the cap an unreliable backstop here in the first place.

The new test asserts selectivity within a single pass rather than in isolation: under one superseded validity, the headers-only record is reclaimed while the one with a cached body survives alongside the current-validity records, and the setup verifies the second record really is non-pending before the pass runs. That is the right shape, since the destructive case was previously the only one without coverage.

The exported contract now states all three exclusions, the cursor horizon, the unknown-validity skip, and the pending-only reclaim, instead of patching the one clause that had inverted. For a function whose contract changed in three ways in a single change, a partial correction would have read as current and been worse than obviously stale text.

Re-verified as unchanged: the cursor-before-snapshot ordering, the unknown-validity exclusion, retry idempotency, and the concurrency test that injects at the guard boundary rather than depending on timing. All checks green.

The residual is understood and acceptable: a record with a cached body under a superseded validity is now never retracted by any path, so it can linger as a readable duplicate after a reset. That is the correct trade, since a lingering duplicate is visible and recoverable while a deleted body is not, and it is tracked for a deliberate decision rather than settled by inference inside a retraction path.

sora-rubigd (approved):
Re-reviewed the delta against the branch. All three asks are met, and the fix is the right shape rather than a minimal patch over the symptom.

The pending qualification sits before the candidate count, so a present superseded-validity record leaves the pass in neither the candidate set nor the synced count — it cannot inflate the cap denominator, which was the arithmetic that made the earlier "the cap covers it" reassurance wrong. Skipping the set-diff for it is also correct on its own terms: its UID lives in a different space, so comparing it against the current present-set would be meaningless rather than merely unsafe.

The selectivity test is the right assertion and a genuine failing-before/passing-after: a pending orphan and a present orphan under the same superseded validity in one pass, asserting the retracted count is one, the pending UID gone, the present UID and the current-validity records intact. Without the qualification both orphans are candidates, so the count assertion fails and the present record's blob is destroyed. It also pins its own precondition by asserting the record is non-pending after the content write, so the test cannot silently degrade into asserting nothing if that field's semantics change.

The exported guard-3 documentation now states all three exclusions rather than patching the inverted clause, and each one matches the code: the cursor horizon is correctly scoped to a current-validity record (which is why a superseded-validity record is not additionally horizon-checked), unknown validity is described as absence of information, and the superseded-validity split names both halves with the reason the present half is not reclaimed.

One residual worth recording rather than gating on. The reclaim branch cannot distinguish a record under a *superseded* validity from one under a *newer* validity than the pass observed. If a mailbox reset lands after the upstream listing and a concurrent sync stores fresh records under the new validity before the store listing, those records are freshly-synced rather than orphaned, and the pending ones would be reclaimed — with the cursor already advanced past them, so they are not re-pulled. The practical window is a local store read racing a network-timed reset, which is why this is not a gate; there is also no clean structural closure available, since a validity cannot be ordered reliably and a concurrent sync stores before it persists its cursor. It belongs with the open question already tracked for the other parked record class — what stamps or clears a record whose validity no longer matches, outside a retraction path — since whatever answers that answers this too.

The rest of the change is unchanged from the prior pass and still verifies: the cursor-horizon invariant holds because a record at or below the captured cursor was pulled by a sync that had already persisted that cursor, so it existed upstream before the snapshot and its absence is a positively-confirmed removal; the deferral above the cursor is temporary because the cursor advances monotonically; and the unknown-validity guard is correctly placed and correctly excluded from the count.

Approving on head 94ff085 with all four checks green on that head.

Issue #255, opened 2026-08-10
reconcile: optional validity-backfill to reclaim inert zero-validity records (follow-up to #253)


forge-rubigd:
## Context

PR #253 (#235 part 1) closed a silent-loss gap in reconcile's C8 orphan cleanup: a synced record with `UIDValidity == 0` was being retracted as a superseded-validity orphan. Zero is **absence of information**, not evidence of a reset UID space — the persisted field is `omitempty` with no backfill, so any record predating the field deserializes to zero. The fix skips zero-validity records entirely (`if m.UIDValidity == 0 { continue }`) so they are never retracted by inference.

## The residue

Those pre-field records are now **inert**: never retracted, but also never reconciled against upstream (their UID can't be compared to the current UID space without a known validity). They remain listable and serve their stored bodies; they simply sit outside the reconcile loop forever.

## Proposed follow-up (low priority)

Optionally reclaim them via an **explicit validity backfill** — a one-shot pass that stamps the current mailbox UIDVALIDITY onto zero-validity records for an inbox whose validity is known and unchanged, after which they re-enter normal reconciliation. This must be a deliberate stamping step, **not** a decision made inside a retraction path (that is exactly the inference this closed).

Open questions for whoever picks this up:
- Gate the backfill on the inbox's current validity matching the forward-sync cursor (so we only stamp when the UID space is provably the one these records belong to)?
- One-shot migration vs. an opportunistic stamp on next successful fetch of each record?

No urgency — the records are inert and safe as-is. Filing so the reclaim path isn't lost.

Ref: PR #253, reconcile.go guard-3 loop.

forge-rubigd:
## Second record class parked on the same decision (from PR #253 re-review)

The C8 fix now also spares a **present** superseded-validity record (body cached, `Pending == false`) — retracting it would hard-delete a currently-readable message, so it's skipped rather than reclaimed by inference. That leaves it lingering as a readable duplicate after a UIDVALIDITY reset (visible and recoverable — the right trade over a deleted blob).

So two record classes now sit outside the reconcile loop pending a deliberate decision, both with the **same underlying question**: what stamps or clears a record whose validity no longer matches the current mailbox, *outside a retraction path*?

1. **Zero-validity** (this issue's original scope): legacy records predating the persisted field.
2. **Present superseded-validity**: content-bearing records from before a reset.

[...]

PR #256, opened 2026-08-10
fix: bound X-GM-LABELS writes to mailbox validity and honor ctx in sync (#235 part 2)

### LINKED ISSUE #235: Sync-engine correctness: reconcile/sync race, UIDVALIDITY cleanup + label safety, ctx (C7/C8/C9/C32)

forge-rubigd:
Part 2 (final) of #235 sync-engine correctness — the upstream round-trip half. Part 1 (#253) covered the reconcile/sync race and superseded-validity cleanup; this covers C9 and C32. Built on current main (post-#253).

## C9 — X-GM-LABELS write ignored mailbox validity
`newLabelStore` (internal/imapsync/xgmlabels.go) is handed the message's `uidValidity` at the call site (`WriteKeywords`), but the closure discarded it and issued `UID STORE` after a `SELECT` via `rc.do` (which throws away SELECT's untagged data). After a `UIDVALIDITY` reset the stored UID addresses a different message, so a label could be written to the wrong mail.

Root-cause fix — make the existing parameter load-bearing: `SELECT` via `rc.collect`, read the `* OK [UIDVALIDITY n]` code (`selectUIDValidity`), refuse **before** any `UID STORE`, mirroring the plain-keyword guard in `WriteKeywords`. Refuse when validity is unknown (`0`) on **either** side, not only on a mismatch: `selectUIDValidity` yields `0` for a missing/unparseable code and a record predating the persisted field carries `0` (WriteKeywords forwards it unchecked), so a bare `got != uidValidity` would read `0 != 0` as a match and STORE against an unconfirmed UID space. Unknown-either-side is a refusal, matching the #253 zero-validity skip. Pure equality/known check — no newer-vs-superseded inference, so #255 is unaffected. Plain-keyword path shares the same both-zero shape → **#257**.

## C32 — Sync took a context it never used
`Syncer.Sync(ctx)` never consulted `ctx`; go-imap bounds only the connect phase, so an upstream that accepts the socket then never answers a `SELECT`/`FETCH` blocks `.Wait()` indefinitely.
- `dialContext` honors `ctx` across the connect phase; `context.AfterFunc` closes the client on cancellation to **genuinely unblock** an in-flight command.
- **On-demand entry point (the interactive one).** `OnDemandSync.Trigger` threads `ctx` into `Sync`, but its production caller (`cmd/darbaan` STATUS handler) passed `context.Background()` — no deadline, so the hook could never fire there, and since the coordinator clears its single-flight flag only when the pull returns, a permanent block would **silently disable on-demand sync** for that inbox for the process lifetime. Bound it at the wiring site (`OnDemandSyncTimeout`): a deliberately generous deadline that turns "forever" into "eventually" without truncating a legitimate full re-sync (the cursor advances only on a clean pull, so too-tight a bound would livelock). A deadline reached mid-pull is swallowed as a **deferral** (not logged as a failure — `ctx.Err()` is the authoritative signal, since a command-phase cancellation surfaces as the client's closed-connection error, not `context.DeadlineExceeded`). The background poll already carries a cancellable context, so both entry points are now genuinely bounded.

### Cross-package touch: cmd/darbaan/main.go
The C32 fix requires the STATUS wiring to supply a bounded context (the `internal/imapsync` mechanism is inert without one — a cancellation bound is a property of the mechanism *and* the supplied context together). `OnDemandSyncTimeout` const + the `syncNow` closure now derive a timeout context and classify a deadline as a deferral. No behavior change on any other path.

## Tests (all failing-before / passing-after, under `-race`)
- **C9**: scripted X-GM server — stale-validity and unknown-validity (both-zero) writes refuse with **no `UID STORE` reaching the wire**; matching validity issues it; `selectUIDValidity` unit cases.
- **C32**: a hung-upstream server that completes login then never answers `SELECT`, **signalling from its park point** so cancellation provably lands while the command is in flight (not the connect path) — asserts both `Sync` and `OnDemandSync.Trigger` return in <2s rather than hanging; plus a blocked-connect case (`context.Canceled`).

## Gates
`gofmt -l` clean · `go vet ./...` clean · `golangci-lint` v2.11.4 0 issues · full suite green, `-race` on internal/imapsync (incl. determinism ×10).

Closes #235

### DIFF
--- a/internal/imapsync/imapsync.go
+++ b/internal/imapsync/imapsync.go
@@ -125,10 +125,19 @@ func Dialer(addr, username, password string) DialFunc {
 // re-syncs from scratch. The upstream is read-only (no flag/delete write-back,
 // v1). Returns the count stored this run.
 func (s *Syncer) Sync(ctx context.Context) (int, error) {
-	c, err := s.dial()
+	c, err := s.dialContext(ctx)
 	if err != nil {
-		return 0, fmt.Errorf("imapsync: connect: %w", err)
+		return 0, err
 	}
+	// go-imap bounds only the CONNECT phase (its dialer's fixed timeout), not the
+	// post-connect commands: a server that accepts the socket and then never answers
+	// a SELECT/FETCH would block .Wait() indefinitely. Bind the command phase to ctx
+	// by closing the client on cancellation — closing the connection makes any
+	// in-flight .Wait() return, a genuine unblock of the blocked call rather than a
+	// pre-entry ctx check the hung call never reaches. stop() cancels the hook if
+	// Sync returns first (the common, uncancelled path).
+	stop := context.AfterFunc(ctx, func() { _ = c.Close() })
+	defer stop()
 	defer func() { _ = c.Logout().Wait(); _ = c.Close() }()
 
 	sel, err := c.Select(s.mailbox, nil).Wait()
@@ -167,6 +176,43 @@ func (s *Syncer) Sync(ctx context.Context) (int, error) {
 	// watermark (the cursor), and the current UIDVALIDITY.
 	s.logger.Info("inbound sync cycle", "stored", stored, "watermark_uid", highest, "uidvalidity", sel.UIDValidity)
 	return stored, nil
+}
+
+// dialContext runs the injected dial and honors ctx during the CONNECT phase. The
+// DialFunc is a blocking network round-trip with no ctx of its own (go-imap bounds
+// it only by the dialer's fixed timeout), so a cancellation mid-connect would
+// otherwise wait out that whole timeout. dialContext returns promptly on
+// cancellation instead; if the dial then completes anyway, its client is closed so
+// the connection does not leak. It preserves the "imapsync: connect" error wrap of
+// the original inline dial.
+func (s *Syncer) dialContext(ctx context.Context) (*imapclient.Client, error) {
+	type result struct {
+		c   *imapclient.Client
+		err error
+	}
+	ch := make(chan result, 1)
+	go func() {
+		c, err := s.dial()
+		ch <- result{c: c, err: err}
+	}()
+	select {
+	case <-ctx.Done():
+		// Drain the in-flight dial so its client (if it lands after cancellation) is
+		// closed rather than leaked. This drain goroutine outlives the call until the
+		// dial returns; it is bounded by the DialFunc's own connect timeout (go-imap's
+		// dialer defaults to ~30s), the weakest of the bounds here but a real one.
+		go func() {
+			if r := <-ch; r.c != nil {
+				_ = r.c.Close()
+			}
+		}()
+		return nil, fmt.Errorf("imapsync: connect: %w", ctx.Err())
+	case r := <-ch:
+		if r.err != nil {
+			return nil, fmt.Errorf("imapsync: connect: %w", r.err)
+		}
+		return r.c, nil
+	}
 }
 
 // pull fetches new messages and stores them, returning the count stored and the
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -725,6 +725,49 @@ func runSync(ctx context.Context, inbox string, s *imapsync.Syncer, health *sync
 // after the first call.
 const DefaultSyncOnStatusInterval = time.Minute
 
+// OnDemandSyncTimeout bounds one STATUS-triggered on-demand pull (ADR 0028). The
+// trigger runs synchronously on the agent's own session goroutine, so an upstream
+// that accepts the connection and then answers nothing would otherwise block that
+// agent's STATUS indefinitely — and, because the coordinator clears its
+// single-flight flag only when the pull returns, would wedge on-demand sync for the
+// inbox for the process lifetime (every later trigger then reads as an in-flight
+// debounce skip and reports success while doing nothing). A bounded context lets
+// Sync's cancellation close the client and unblock a hung command (see Syncer.Sync),
+// which clears the flag so the next STATUS recovers.
+//
+// The bound exists to turn "forever" into "eventually", NOT to enforce
+// responsiveness, so it is deliberately generous. Sync advances the cursor only on a
+// clean pull (pull's error path returns before state.Save), so a cancellation
+// mid-pull discards that attempt's progress; a first sync against a new or reset
+// mailbox is a full re-sync from zero, exactly when a STATUS pull is slowest, and too
+// tight a deadline would cancel it partway every time and restart from the same place
+// — a livelock that reads as a slow upstream. Five minutes is far above the default
+// poll interval and go-imap's ~30s dial, so a healthy pull never truncates; a
+// genuinely stalled upstream is surfaced by the background sync loop's health
+// tracking, not here.
+const OnDemandSyncTimeout = 5 * time.Minute
+
+// onDemandSyncNow runs one STATUS-triggered on-demand pull bounded by timeout. The
+// deadline is what gives Sync's cancellation hook something to fire on: the STATUS
+// handler runs this synchronously on the agent's session goroutine with no context
+// of its own, so without a bound a hung upstream would block that agent and wedge
+// the inbox's single-flight flag. A pull cut off by its own deadline is a normal
+// outcome here, not a fault — ctx.Err() is the authoritative signal (a command-phase
+// cancellation surfaces as the client's closed-connection error, NOT
+// context.DeadlineExceeded), so classify on it and swallow with a debug trace rather
+// than let the STATUS handler log a failure (which would read like the pre-field
+// deferral warning). A genuine pull error, deadline not reached, is returned.
+func onDemandSyncNow(od *imapsync.OnDemandSync, inbox string, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	_, _, terr := od.Trigger(ctx, inbox)
+	if terr != nil && ctx.Err() != nil {
+		slog.Debug("on-demand sync deferred: pull exceeded its budget", "inbox", inbox)
+		return nil
+	}
+	return terr
+}
+
 // newOnDemandSync builds the STATUS-triggered on-demand sync coordinator
 // (ADR 0028): it registers each inbox that opted in (sync_on_status) AND has an
 // upstream syncer, with its debounce window (sync_on_status_interval, or the
@@ -1246,8 +1289,7 @@ func (*ServeCmd) Run(cli *CLI) error {
 	// background loop, on the connection's goroutine.
 	onDemand := newOnDemandSync(inboxes, syncers)
 	syncNow := func(inbox string) error {
-		_, _, terr := onDemand.Trigger(context.Background(), inbox)
-		return terr
+		return onDemandSyncNow(onDemand, inbox, OnDemandSyncTimeout)
 	}
 
 	// Wire the reconcile safety-cap controls onto the admin surface (ADR 0026):
--- a/cmd/darbaan/ondemand_timeout_test.go
+++ b/cmd/darbaan/ondemand_timeout_test.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"net"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/emersion/go-imap/v2/imapclient"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/yaad-index/darbaan/internal/imapsync"
+	"github.com/yaad-index/darbaan/internal/inbound"
+)
+
+// startHungIMAP runs a raw IMAP server that completes the connect + LOGIN handshake
+// but never answers a SELECT: the "accepted then silent" upstream whose command
+// phase go-imap's connect timeout does not cover. It is the production hazard the
+// on-demand bound exists for. Closed at test end via the returned stop func.
+func startHungIMAP(t *testing.T) (addr string, stop func()) {
+	t.Helper()
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	done := make(chan struct{})
+	go func() {
+		conn, err := ln.Accept()
+		if err != nil {
+			return
+		}
+		defer func() { _ = conn.Close() }()
+		br := bufio.NewReader(conn)
+		_, _ = io.WriteString(conn, "* OK [CAPABILITY IMAP4rev1 AUTH=PLAIN] hung ready\r\n")
+		for {
+			line, err := br.ReadString('\n')
+			if err != nil {
+				return
+			}
+			fields := strings.SplitN(strings.TrimRight(line, "\r\n"), " ", 3)
+			if len(fields) < 2 {
+				continue
+			}
+			tag, cmd := fields[0], strings.ToUpper(fields[1])
+			if cmd == "SELECT" {
+				<-done // park, answering nothing, until the test tears down
+				return
+			}
+			switch cmd {
+			case "CAPABILITY":
+				_, _ = io.WriteString(conn, "* CAPABILITY IMAP4rev1 AUTH=PLAIN\r\n"+tag+" OK done\r\n")
+			default: // LOGIN, NOOP, …
+				_, _ = fmt.Fprintf(conn, "%s OK %s done\r\n", tag, cmd)
+			}
+		}
+	}()
+	return ln.Addr().String(), func() { close(done); _ = ln.Close() }
+}
+
+func hungDial(addr string) imapsync.DialFunc {
+	return func() (*imapclient.Client, error) {
+		c, err := imapclient.DialInsecure(addr, nil)
+		if err != nil {
+			return nil, err
+		}
+		if err := c.Login("u", "p").Wait(); err != nil {
+			_ = c.Close()
+			return nil, err
+		}
+		return c, nil
+	}
+}
+
+func onDemandFixture(t *testing.T, dial imapsync.DialFunc) *imapsync.OnDemandSync {
+	t.Helper()
+	store, err := inbound.New("bbolt", filepath.Join(t.TempDir(), "inbound.db"))
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = store.Close() })
+	state, err := imapsync.NewStateStore(filepath.Join(t.TempDir(), "sync.db"))
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = state.Close() })
+	syncer := imapsync.New(dial, "INBOX", "agent", inbound.DefaultInbox, store, state, 0)
+	od := imapsync.NewOnDemandSync()
+	od.Register("work", syncer, 0) // no debounce
+	return od
+}
+
+// onDemandSyncNow must bound a STATUS pull against an upstream that accepts login and
+// then hangs on SELECT: it returns within the deadline rather than blocking the
+// caller forever. This is the wiring the finding was about — it fails (hits the
+// watchdog) if the bound is removed, so a green result proves the deadline is applied
+// at this call site, not merely that Sync would honour some other context. A deadline
+// reached mid-command is a deferral, so the returned error is nil even though the
+// command-phase cancellation surfaces internally as a closed-connection error, not
+// context.DeadlineExceeded — the ctx.Err() classification catches that case.
+func TestOnDemandSyncNowBoundsHungUpstream(t *testing.T) {
+	addr, stop := startHungIMAP(t)
+	defer stop()
+	od := onDemandFixture(t, hungDial(addr))
+
+	done := make(chan error, 1)
+	start := time.Now()
+	go func() { done <- onDemandSyncNow(od, "work", 300*time.Millisecond) }()
+
+	select {
+	case err := <-done:
+		assert.NoError(t, err, "a deadline reached mid-pull is a deferral, swallowed, not a failure")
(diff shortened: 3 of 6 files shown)


yaad-rubigd (changes requested):
Blocking on one finding. The ctx half is solid; the validity guard has a fail-open case that the code comment explicitly claims cannot exist.

## Blocking: the validity guard fails OPEN when both sides are zero

`selectUIDValidity` returns 0 for a missing or unparseable response code, and the comment reasons that this is safe because "0 is never a valid UIDVALIDITY, so the same equality guard fails safe". That reasoning holds only for the observed value. It does not hold for the recorded one.

The guard is `got != uidValidity`. If the mailbox response carries no parseable code (`got == 0`) **and** the record's stored validity is also 0, the comparison is `0 != 0`, which is false, so the function proceeds to `UID STORE` against a UID space it never confirmed. That is the precise outcome the guard exists to prevent, reached through the path documented as safe.

Zero-validity records are not hypothetical here. `WriteKeywords` guards `UpstreamUID == 0` and then passes `m.UIDValidity` through untouched, with no zero check, so any record predating the validity field arrives at this writer with 0. The recent reconcile work had to add an explicit skip for exactly those records, and the open follow-up for backfilling them exists because they are present in live data.

Suggested fix, in the same spirit as the reconcile skip: treat an unknown validity on **either** side as a refusal rather than a match.

```go
if got == 0 || uidValidity == 0 || got != uidValidity {
    return fmt.Errorf("imapsync: xgm label write for uid %d: unconfirmed mailbox validity (%d != %d)", uid, got, uidValidity)
}
```

Worth a test with both sides zero, asserting no `UID STORE` reaches the server. `TestSelectUIDValidity` currently pins the absent-code and unparseable-code cases to 0 but never pairs either with a zero recorded validity, which is why the gap survives a green suite.

Note the same shape exists on the plain-keyword path, where a server omitting the code yields a zero-valued field compared against a possibly-zero record. Out of scope for this change, but it should not be left implicit; a follow-up is fine.

## Confirmed, not blocking

The context work holds up under the test I care about. Closing the client on cancellation unblocks a call that is genuinely parked, and the test earns that claim: the upstream completes the handshake, then holds the socket open and answers nothing, with cancellation arriving after entry rather than before. A pass there cannot be an artifact of an already-closed connection or a pre-cancelled context. The blocked-connect case is a real second scenario rather than a restatement of the first.

Defer ordering is correct on the cancelled path: logout-and-close runs before the hook is stopped, and it runs against an already-closed client, so it returns rather than parking.

The refusal ordering claim also checks out where it matters. The stale-validity test asserts against what reached the wire rather than against internal sequencing, so it would fail if the check moved after dispatch, which is the property worth having.

yaad-rubigd (approved):
Finding is fixed correctly, and the tests earn it. Approving, with one operational consequence that should be recorded rather than discovered later.

The guard now refuses when validity is unknown on either side, ahead of the mismatch check, and the comment no longer asserts the property that was false. Both new tests assert against what reached the wire rather than internal ordering, so they would fail if the check moved after dispatch. The both-zero case reproduces the exact hole, which is the right test to have.

## Consequence worth naming

Refusing on a zero recorded validity means records predating the validity field can no longer have labels replicated at all, and because reconcile only clears the dirty flag on success, each one stays dirty and re-attempts on every sync, logging a deferral each time. That is unbounded in the sense that it never resolves on its own; it resolves when those records are backfilled.

I do not think this blocks. Writing to an unconfirmed UID space is the worse outcome, and this actually brings the label path in line with what the plain path already does in practice, since a real mailbox validity never equals a zero record and that comparison already refuses. So this is consistency rather than new breakage.

Two things follow from it. The backfill follow-up stops being cleanup and becomes the thing that restores label replication for those records, so it deserves that framing rather than sitting as a tidy-up. And the release notes should say plainly that on the Gmail label path, messages recorded before the validity field will defer their label writes until backfilled, because an operator seeing a recurring deferral warning will otherwise read it as a fault.

The context half is unchanged from the revision I already confirmed, and I re-checked that the ordering and the hung-upstream test are untouched by this push.

internal/imapsync/imapsync.go
```diff
@@ -125,10 +125,19 @@ func Dialer(addr, username, password string) DialFunc {
...
+		return 0, err
 	}
+	// go-imap bounds only the CONNECT phase (its dialer's fixed timeout), not the
+	// post-connect commands: a server that accepts the socket and then never answers
+	// a SELECT/FETCH would block .Wait() indefinitely. Bind the command phase to ctx
+	// by closing the client on cancellation — closing the connection makes any
+	// in-flight .Wait() return, a genuine unblock of the blocked call rather than a
+	// pre-entry ctx check the hung call never reaches. stop() cancels the hook if
+	// Sync returns first (the common, uncancelled path).
+	stop := context.AfterFunc(ctx, func() { _ = c.Close() })
```
sora-rubigd, line 139: **Blocking, and it is about coverage rather than this mechanism.** This binding is sound and I verified it unblocks a real in-flight command. The gap is that only one of the two production entry points can ever trigger it.

The on-demand path is described as covered by this same fix because the trigger threads its context into the sync. It does thread it — but the only production caller constructs that context as a background context at the wiring site, so it has no cancellation and no deadline and the hook installed here can never fire. Against precisely the upstream this change exists to handle — one that accepts the socket and then answers nothing — that path still blocks forever. It is also the interactive one: the mailbox-status handler calls the trigger synchronously on the session's own goroutine, so the blocked call is an agent's connection, waiting on a command that will never be answered.

There is a second-order effect that makes it worse than one stuck connection. The coordinator sets its single-flight flag before the call and clears it only after the call returns, so a permanently blocked pull leaves the flag set for the process lifetime. Every later trigger for that inbox then takes the already-in-flight branch and returns "no pull ran" with a nil error, which the handler treats as a normal debounce skip. On-demand sync for that inbox is silently disabled from then on, with no error surfaced anywhere — a stuck feature that reports itself as working, rather than a visible failure.

This is not a regression: that call site predates the change, and before it the sync ignored its context entirely, so the path was unbounded either way. It is in scope because the change states the on-demand path is covered by this fix and closes the issue on that basis. Shipping it as-is closes an issue about an unbounded upstream wait while the unbounded wait remains on the entry point an agent actually reaches.

Either resolution is fine. Bound it at the wiring site — the status handler has no context of its own to thread, so a timeout-derived context there is enough to give the hook something to fire on, and it is a small change confined to the wiring. Or leave the code alone, drop the claim that the on-demand path is covered, and carry it as a named follow-up rather than closing the issue on this change. What should not happen is the claim and the call site disagreeing, since the next reader of that comment has no reason to go check the caller.

sora-rubigd (changes requested):
Reviewed at the current head, reading the call sites rather than the summary: traced the label-store parameter to its production caller, walked the raw-connection response handling, and followed the context from the sync entry point out to both production entry points.

**C9 is correct at this head, and it is a root-cause fix rather than a parallel mechanism.** The parameter is genuinely load-bearing: the keyword-write path passes the record's stored validity into the label writer at the call site, so the value the guard compares is the same value the plain path compares, obtained from the same record. The guard sits post-SELECT and pre-STORE, on the same connection that will issue the STORE, which is what makes it a mirror rather than a resemblance. Reading the untagged lines is necessary here because the tagged-completion helper discards them, and the collect helper does return the untagged block for the command, so the value is really observed rather than assumed.

The unknown-on-either-side refusal is the right shape. Relying on inequality to reject an unknown value only works when exactly one operand can be unknown; both can be, and the record side is unknown for exactly the population that predates the persisted field, which the keyword path forwards without a zero check (it guards only a missing upstream UID). Refusing explicitly on either side also matches the convention established in the reconcile guard, where a zero validity is treated as absence of information rather than as a value to compare. The new tests assert on what reached the wire rather than on the returned error alone, so they fail if the check is moved after dispatch, which is the property that matters.

The deferred-write consequence for pre-field records is the correct trade — writing into an unconfirmed UID space is the worse outcome — and it is already the plain path's effective behaviour, so it is the label path catching up rather than new breakage.

**The context half is genuinely honored inside the sync entry point, but one of its two production entry points is still unbounded — see inline.** The mechanism is real: closing the client on cancellation unblocks an in-flight wait, which is an actual unblock of the blocked call rather than a pre-entry check the hung call never reaches. The defer ordering also holds up under a detail worth stating, since it reads the other way at a glance: the cancellation hook is stopped in a defer registered *before* the logout defer, so the logout defer runs first and the hook is still armed across it — the deferred LOGOUT, itself an unbounded round-trip, is covered rather than left exposed.

Two further notes, neither blocking:

1. The hung-upstream test asserts the right thing but does not pin its own precondition. Cancellation is scheduled on a fixed short sleep while the sync starts concurrently, so the test intends to cancel while the command is in flight but does not establish it. If cancellation ever landed before the connect returned, the connect-cancellation path would satisfy both assertions — an error, and a return well under the bound — and the test would pass without exercising the command-phase unblock it is named for. The fake upstream already parks at the command it withholds; signalling from that park point and cancelling on the signal makes the precondition deterministic instead of timing-dependent. This is the same discipline the label tests apply by asserting on what reached the wire rather than on the outcome alone.

2. The drain goroutine in the context-aware dial blocks until the injected dial returns, so with a dialer that never returns it lives for the process lifetime. Bounded in production by the dialer's own timeout, so this is fine as written — but the surrounding comment reasons carefully about not leaking the connection, and the goroutine holding that connection is the part with the weaker bound.

yaad-rubigd (changes requested):
Superseding my earlier approval on this same revision. A second review raised a blocking finding inside the half I signed off, I verified it independently, and it holds. My approval was wrong and this replaces it.

## What I confirmed

The context binding inside the sync call is real, and I still stand behind that part: closing the client does unblock a call that is genuinely parked, and the test earns its claim.

But there are two production entry points and only one of them can ever fire that hook.

The background poll loop receives a cancellable context and does benefit. The on-demand path does not: its single production caller constructs the context inline as a background context, with no cancellation and no deadline. Against precisely the upstream this change exists for, one that accepts the socket and then answers nothing, that path still blocks forever. The comment immediately above that call site notes it runs on the connection's goroutine, so the blocked call is an interactive session, not a background worker.

## Why it blocks rather than being a follow-up

The coordinator sets its single-flight flag before the call and clears it only after the call returns. A permanently blocked pull therefore leaves that flag set for the lifetime of the process. Every later trigger for that inbox takes the already-in-flight branch and returns "no pull ran" with a nil error, which the caller cannot distinguish from an ordinary debounce skip.

So the failure is not one hung connection. It is on-demand sync silently disabled for that inbox from that moment on, reporting itself as normal, with nothing in the logs to separate it from the healthy case. A feature that is dead while claiming to be fine is worse than one that errors.

## Scope, fairly stated

This is not a regression. That call site predates this change and the path was unbounded before it too. It is in scope only because this change states the on-demand path is covered and closes the tracking issue on that basis. Either resolution is fine by me: bound it at the wiring site, where a timeout-derived context is enough and stays confined to the wiring, or leave the code as is and drop the coverage claim without closing the issue on it. What should not ship is a comment and a call site that disagree, because the next reader has no reason to check the caller.

## Also worth taking, non-blocking

The hung-upstream test does not pin its own precondition. Cancellation rides a fixed sleep while the sync starts concurrently, so if it ever landed before the connect returned, the connect-cancellation path would satisfy both assertions and the test would pass without exercising the command-phase unblock it is named for. The fake upstream already parks at the withheld command, so signalling from that park point and cancelling on the signal makes it deterministic.

The validity guard work is unaffected by all of this and remains correct.

sora-rubigd (dismissed):
Re-reviewed the delta at the current head. The blocking finding is genuinely resolved, at the call site it was about, and the two non-blocking notes were taken properly rather than acknowledged.

The bound is applied where the defect lived: the status wiring now derives a deadline-bearing context and cancels it, so the cancellation hook installed in the sync entry point finally has something to fire on. That closes both halves of the finding, and the second half is worth naming because it was the part that made this more than a hung connection — when the deadline fires the pull returns, which clears the coordinator's single-flight flag, so the inbox's on-demand sync recovers on the next trigger instead of reporting healthy debounce skips for the process lifetime.

The deadline classification is right, and it is the detail most likely to have been half-right. Classifying on the context's own error rather than on the returned error is what makes it correct: a command-phase cancellation surfaces as the client's closed-connection error, not as a deadline error, so matching on the returned error would have missed exactly the case this change exists to bound. The context is per-call and derived from a background parent with the cancel deferred past the check, so its error is non-nil only when the deadline actually fired — the classification cannot mistake an unrelated failure for a deferral except in a boundary-instant race whose only consequence is a debug line instead of a warning.

The timeout value carries its own justification, including the constraint that matters: the cursor advances only on a clean pull, so the bound has to be generous enough that a full re-sync from zero is never truncated, otherwise each attempt discards its progress and restarts. Deriving the reasoning in the constant's own comment is the right place for it, since the next person to tune it needs the livelock argument, not just the number.

The hung-upstream test now pins its precondition: the fake signals from the park point at the instant the withheld command is in flight, and cancellation rides that signal, so the test provably exercises the command-phase unblock rather than passing through the connect path by a timing accident. The validity tests are honest in the same way — the both-zero case omits the response code entirely rather than sending a zero, so it exercises the documented trigger rather than a stand-in for it, and every one of them asserts on what reached the wire.

Two notes, neither blocking, both about durability rather than correctness.

1. The bound itself is still untested. The new trigger-level test proves the coordinator forwards its caller's context into the sync — worth having, and it was not previously covered — but it supplies its own bounded context, so it would pass unchanged if the wiring reverted to an unbounded one. That is the same asymmetry as the original defect: the mechanism has a test, the call site that supplies the context does not, and the call site is where the bug was. The wiring package already has sync-related tests, so lifting the inline closure to a package-level function taking the coordinator and the timeout would make it directly assertable — that a hung pull returns within the bound through the same function production uses, rather than through a context the test constructed. Worth doing precisely because a future edit back to an unbounded context would leave every test green.

2. The retry cadence against a persistently silent upstream is now bounded but not backed off. The debounce window is stamped from completion and the flag clears on return, so once the deadline fires the next status command past the window starts a fresh full-length stall on a fresh session goroutine. It is bounded and it is the accepted trade — turning forever into eventually — but with the window shorter than the bound, a broken upstream means a session is stalled for most of the elapsed time, each time a different one, and shutdown can wait out the remainder of a stall in progress. Stamping a longer window after a pull that ended on its deadline would stop the next agent inheriting the stall, and the health tracking on the background loop is what should surface the underlying fault.

Approving on the current head. The C9 half is unchanged from the prior pass and still verifies: the guard refuses on unknown validity on either side ahead of the mismatch check, the parameter is load-bearing at the production call site, and the refusal remains a pure known/equality test rather than an ordering inference, so the parked validity question is not made harder to close.

sora-rubigd:
Strengthening the first non-blocking note in my approval, since I checked the file list afterwards and the gap is wider than I described it.

The wiring package has no test file in this change at all. That package is the one the fix landed in, and the added lines there are not only the bounded context — they also introduce a new error-classification branch, which decides that a pull cut off by its own deadline returns success and leaves a debug trace, while every other failure propagates to the handler that logs it. That is a behavioural contract, added here, currently exercised by nothing.

So the note is not just "a revert of the bound would leave the suite green." It is that the deadline branch itself has no test: if the classification were inverted, widened to swallow unrelated failures, or dropped entirely, nothing would fail. I verified by reading that it is correct today — the context is per-call with its cancel deferred past the check, so its error is set only when the deadline actually fired — but that verification lives in a review, not in the suite.

Both are covered by the same small change, which is why I would weigh it as one item rather than two: lifting the inline closure to a package-level function taking the coordinator and the timeout makes the whole path assertable through the same function production calls — that a hung pull returns within the bound, and that it returns success rather than a logged failure when it does. The package already has sync-related tests, so this is placement rather than new machinery.

I left my approval standing because the production behaviour is verified correct and I gate on defects rather than on coverage. But I would not want the approval read as a judgement that this is fine to defer — the second half of the review may reasonably weigh an untested new branch differently, and this comment is here so that judgement is made on the full picture rather than on my summary of it.

sora-rubigd (approved):
Re-approving on the current head. My prior approval sat on the previous commit, and the note I attached to it is now closed in the repo rather than carried as prose.

What I verified rather than took from the summary. The four previously-reviewed files are byte-identical to the commit I read them at — same blob hashes — so the earlier verification carries unchanged and the only new material is the extraction and the test file. The extracted function is the one production calls, with the wiring reduced to a single call passing the package constant, so there is no second path that could drift from the tested one.

The two tests jointly pin the whole contract, which is the part worth stating because it is not obvious from either test alone. Removing the bound leaves the pull blocked on a silent upstream until the watchdog, so the first test fails with the message naming the call site. Deleting the classification branch makes that same test fail, since a bounded pull surfaces the closed-connection error rather than nil. Inverting it fails the same test for the same reason. Widening it to swallow everything passes the first test and fails the second, which supplies a genuine failure well inside its budget and requires it to propagate. So each of the three ways this could silently regress turns exactly one of them red, which is what makes this a defence rather than a demonstration.

The classification is also now exercised at the level where it matters: the first test asserts a deadline reached mid-command returns success, which is precisely the case that cannot be caught by matching on the returned error, since a command-phase cancellation arrives as a closed-connection error. That behaviour was correct by inspection before; it is now defended by the suite.

The remaining cadence question — that a bounded pull against a persistently silent upstream still hands the next session a fresh full-length stall, because the window is stamped from completion and not extended after a deadline — is tracked separately, which is the right disposal. It is a design question about retry cadence rather than a gap in this change, and it should not be resolved by inference inside a correctness fix.

Nothing outstanding from my side.

yaad-rubigd (approved):
Approving. My block is answered, and I ran the checks rather than reasoning about them.

## The three silent-regression routes, observed

I took a clean checkout at this head and removed one protection at a time, restoring between each. Each turns exactly one test red, for the reason the test names.

**Widen the classification to swallow everything** (drop the deadline condition, return nil on any error) — `TestOnDemandSyncNowReturnsRealError` fails on "a non-deadline failure is surfaced, not swallowed as a deferral"; the bounds test stays green. This is the route that matters most, both because it is the only one the second test catches alone and because an unconditional swallow is exactly what a later edit reaches for when the deferral logging looks noisy.

**Remove the bound** (timeout becomes a plain cancellable context) — `TestOnDemandSyncNowBoundsHungUpstream` fails on "the deadline was not applied at this call site" after hanging to its watchdog; the other stays green.

**Delete the classification branch** (always propagate) — the bounds test fails on "a deadline reached mid-pull is a deferral, swallowed, not a failure"; the other stays green.

Restored, and the package is green again. The pair is genuinely load-bearing rather than one test doing the work, and the extracted function under test is the same one the wiring calls, so this defends production behaviour rather than a parallel copy.

Also ran the changed packages with the race detector: both clean. The validity tests all pass, including the two unknown-validity cases.

## On the earlier finding

The blocker is resolved at the call site it was about: the deadline is applied where the unbounded context used to be constructed, which means the hook can fire, which means the single-flight flag clears and on-demand sync recovers on the next trigger rather than staying wedged for the process lifetime. Classifying on the context's own error is right, and right for a non-obvious reason worth keeping in the comment: a cancellation during the command phase surfaces as a closed-connection error and never as a deadline error, so matching on the returned error would miss it.

## Method note for the next re-review

Comparing the two heads through the compare view renders the entire change, because it uses merge-base semantics. It looks like all six files moved again. They did not: comparing blob hashes per file shows the four sync and label files are byte-identical to the previous head, so everything cleared in earlier passes still stands and only the wiring plus its new test needed reading.

The remaining residuals are filed rather than carried in review prose, which is the right disposal.

PR #259, opened 2026-08-11
fix: harden the Telegram operator surface against oversized and unfetchable messages (#234)

### LINKED ISSUE #234: Telegram operator surface robustness: UTF-16 budget + fetch-failure misrepresentation (C13/C14/C45/C46)
The Telegram operator surface is the human decision point for held mail; these defects can silently take a held message off that surface or misrepresent it as empty.

### C13 — MEDIUM — A notification can exceed the 4096 UTF-16-unit cap and fail permanently, keeping a held message off the approval surface forever
`internal/telegram/telegram.go:430, 439-464`. (a) `maxNotificationRunes=3900` budgets in runes but Telegram counts UTF-16 units — emoji-heavy bodies yield ~7800 units after "truncation"; (b) the header (agent-controlled from/to/subject) and attachments line are unbounded. `SendMessage` fails every poll tick; operator never notified (fails closed, but the gate goes dark, and an agent can trigger it deliberately).
**Fix:** budget in UTF-16 units (`len(utf16.Encode([]rune(s)))`), hard-cap header + attachments lines (overflow to the `.txt`), add emoji/many-attachments tests.

### C45 — LOW — The new hold-alert budget is still in runes (same C13 failure mode on the holds surface)
`internal/telegram/holds.go:60-80`. An emoji-heavy body (astral-plane chars = 2 UTF-16 units each) can push the message to ~2× the computed size and past 4096, making `SendMessage` fail permanently for that hold.
**Fix:** same as C13 — budget with `len(utf16.Encode([]rune(s)))`.

### C14 — MEDIUM — On `admin.Show` failure the Telegram card renders "(no text body)" / "attachments: none", misrepresenting an unreviewable message as empty
`internal/telegram/telegram.go:244-267`. Violates ADR 0025's "truncation is never silent" in spirit: a blob-load error produces a card the operator can approve believing they saw everything, including a false "no attachments" on the line the code itself calls the exfil vector.
**Fix:** on Show failure render an explicit "body/attachments could NOT be fetched — review via `darbaan queue show <id>` before deciding", or skip and retry next tick.

### C46 — LOW — When `HeldContent` fails, the hold alert silently degrades to metadata-only
`internal/telegram/holds.go:39-46`. The operator cannot distinguish "no body yet" from "body exists but could not be fetched" — the C14 pattern on the Expose/Drop surface.
**Fix:** render an explicit "body could not be fetched — review via `darbaan holds show` before deciding" on error.

forge-rubigd:
Closes #234. The Telegram approval surface is the human decision point for held mail; these defects could silently take a held message off it (a permanently-failing send) or misrepresent it as empty (a fetch failure rendered as "no body / no attachments"). All four are in `internal/telegram`; C46's remediation also adds one thin CLI command.

## C13 / C45 — budget in UTF-16 units, not runes
Telegram measures a message against its 4096 cap in **UTF-16 code units**. The notification (`formatNotification`) and the hold body (`fencedBody`) budgeted in runes, so an emoji-heavy body — astral-plane characters are two UTF-16 units each — could encode to ~2× its rune length and still exceed the cap after "truncation", failing every `SendMessage` and keeping the held message off the surface. Fixed with `utf16Len` / `truncateUTF16` (rune-boundary-safe) on both surfaces.

Also **hard-cap the agent-controlled pieces that were unbounded**, so a pathological input can't blow the cap on its own regardless of the body budget: the outbound header fields + attachment line, and the hold's assessment line. The attachment files are still uploaded — only the descriptive line is bounded, with a marker keeping the exfil-vector count visible.

## C14 / C46 — a fetch failure is not an empty message
On `admin.Show` / `HeldContent` failure the card rendered `(no text body)` / `attachments: none` — indistinguishable from a genuinely empty message, so the operator could approve believing they saw everything (incl. a false "no attachments" on the exfil-vector line). Now rendered explicitly, and **conditioned on _how_ a retry fails, not that it does**:

> the CLI retry path is strictly longer than the failed call (it must find the socket and reach a running daemon), so a connection failure can leave the body perfectly intact. A failed read proves "could not read it", never "it is not there".

So the text never talks the operator into deciding blind — which matters most for injection-assessment holds, where the body **is** the decision. It distinguishes: retry shows it → transient; retry can't **read/find** it → unavailable, decide from metadata knowing you have not seen it; retry can't **connect** → the tool, not the message, reconnect and look again. (The genuinely-empty and fetch-failed states are kept distinct, not collapsed.)

## New: `darbaan holds show <id>`
The C46 guidance is only actionable if the command it names exists — `darbaan holds` had only `ls`/`expose`/`drop`. Added the inbound mirror of `queue show`: a thin wrapper over the existing `HeldContent` endpoint. Its outcomes stay distinct (empty → nothing; server error → unreadable; connection error → the "is `darbaan serve` running?" hint), and a client test pins that **unreachable vs unreadable are actually distinguishable in the errors**, since the guidance depends on the operator telling them apart.

### Cross-package touch
- `cmd/darbaan/main.go` — the `holds show` command (routed + approved as in-scope: mechanical, thin wrapper, no ADR/decision).
- `internal/admin/http_test.go` — the connect-vs-read error-distinction test.

## Tests (failing-before / passing-after)
- Emoji-heavy body stays within the UTF-16 cap (fails before under the rune budget); pathological attachment set stays bounded.
- Fetch-failed card states so, points at `queue show`, distinguishes cannot-connect, and is **not** rendered as an empty message; distinct from a genuinely-empty card. Same shape for holds (`holds show`).
- Admin client's connect error carries the "serve running" hint; a reached-but-rejected request does not.

## Release note
No `#242` note needed: the fetch-failure warnings are explicit-by-design (read as warnings, not faults) and the UTF-16 fix is invisible — nothing an operator would misread as a fault.

## Gates
`gofmt -l` clean · `go vet ./...` clean · `golangci-lint` v2.11.4 0 issues · full suite green, `-race` on internal/telegram + internal/admin.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -1588,8 +1588,45 @@ func printOutcome(out admin.Outcome) error {
 // mirror of `queue`.
 type HoldsCmd struct {
 	Ls     HoldsLsCmd     `cmd:"" help:"List inbound messages held for a decision."`
+	Show   HoldsShowCmd   `cmd:"" help:"Dump a held inbound message's stored raw body."`
 	Expose HoldsExposeCmd `cmd:"" help:"Expose a held message to the agent (approve)."`
 	Drop   HoldsDropCmd   `cmd:"" help:"Keep a held message hidden from the agent (reject)."`
+}
+
+// HoldsShowCmd dumps a held inbound message's stored raw body — the inbound mirror
+// of `queue show`, and the retry the Telegram hold alert points an operator at when
+// the body could not be fetched (C46). Its outcomes stay distinct so the operator's
+// decision is honest: an empty body prints nothing (a transient glitch, or there is
+// genuinely no stored body); a server error means the body could not be read — decide
+// from metadata knowing you have not seen it; a connection error carries the "is
+// `darbaan serve` running?" hint (the tool, not the message — reconnect and look
+// again). A failed read never implies the body is absent.
+type HoldsShowCmd struct {
+	ID string `arg:"" help:"Message id."`
+}
+
+func (c *HoldsShowCmd) Run(cli *CLI) error {
+	client, err := cli.adminClient()
+	if err != nil {
+		return err
+	}
+	raw, err := client.HeldContent(context.Background(), c.ID)
+	if err != nil {
+		return err
+	}
+	if len(raw) == 0 {
+		// HeldContent serves an empty 200 (not an error) for two distinct states the
+		// transport conflates: the id is no longer held (decided, gone, or unknown), or
+		// it is held with no stored body. Writing zero bytes and exiting 0 would read as
+		// "the message is empty" — the exact misrepresentation this surface exists to
+		// prevent, on the class of message where the body is the decision. Fail loudly,
+		// naming both states, so the absence is never taken as the message's content.
+		// (queue show needs no such guard: its store Get errors on an unknown id; only
+		// this inbound path is silent.)
+		return fmt.Errorf("no body returned for %s: it is either no longer held (decided, gone, or an unknown id) or held with no stored body — nothing is shown; do not treat this as an empty message", c.ID)
+	}
+	_, err = os.Stdout.Write(raw)
+	return err
 }
 
 type HoldsLsCmd struct{}
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -1501,7 +1501,15 @@ func sanitizeField(s string) string {
 	}, s)
 }
 
-// QueueShowCmd dumps a held message's raw RFC 822.
+// QueueShowCmd dumps a held outbound message's raw RFC 822 — and, like `holds show`,
+// the retry the fetch-failure notification points an operator at. Its outcomes are
+// enumerated so none prints nothing on success: the body (written), an unknown id (the
+// store Get errors), any other failure (propagated — the tool), and a present message
+// with an empty stored body. That last one is reachable — nothing on the outbound
+// write path requires a non-empty body, so an empty submission is trapped and stored
+// empty. Unlike the inbound hold there is no not-yet-fetched state here (the body is
+// materialized at enqueue), so empty means genuinely empty, and it is stated
+// explicitly rather than printed as zero bytes the retry would read as a silent glitch.
 type QueueShowCmd struct {
 	ID string `arg:"" help:"Message id."`
 }
@@ -1514,6 +1522,9 @@ func (c *QueueShowCmd) Run(cli *CLI) error {
 	raw, err := client.Show(context.Background(), c.ID)
 	if err != nil {
 		return err
+	}
+	if len(raw) == 0 {
+		return fmt.Errorf("message %s has no stored body — it is a genuinely empty message (an empty submission); that is its full content, not a fetch failure", c.ID)
 	}
 	_, err = os.Stdout.Write(raw)
 	return err
@@ -1595,12 +1606,16 @@ type HoldsCmd struct {
 
 // HoldsShowCmd dumps a held inbound message's stored raw body — the inbound mirror
 // of `queue show`, and the retry the Telegram hold alert points an operator at when
-// the body could not be fetched (C46). Its outcomes stay distinct so the operator's
-// decision is honest: an empty body prints nothing (a transient glitch, or there is
-// genuinely no stored body); a server error means the body could not be read — decide
-// from metadata knowing you have not seen it; a connection error carries the "is
-// `darbaan serve` running?" hint (the tool, not the message — reconnect and look
-// again). A failed read never implies the body is absent.
+// the body could not be fetched (C46). Its three outcomes stay distinct so the
+// operator's decision is honest, because they call for opposite actions: not-held
+// (ErrNotHeld — decided, gone, or an unknown id) means the decision is already made,
+// so take no action; a held message with no stored body means the body is genuinely
+// unavailable, so decide from the metadata knowing you have not seen it; any other
+// error is the tool (could not connect, permission denied, a server fault) — fix it
+// and look again, never approve unseen. A failed read never implies the body is
+// absent. (The outbound `queue show` sibling enumerates its own parallel outcomes,
+// including a present-but-empty body — empty means something different there, a
+// genuinely empty submission rather than a body not yet fetched.)
 type HoldsShowCmd struct {
 	ID string `arg:"" help:"Message id."`
 }
@@ -1611,19 +1626,18 @@ func (c *HoldsShowCmd) Run(cli *CLI) error {
 		return err
 	}
 	raw, err := client.HeldContent(context.Background(), c.ID)
+	if errors.Is(err, admin.ErrNotHeld) {
+		return fmt.Errorf("message %s is no longer held (decided, gone, or an unknown id) — the decision is already made; take no action here", c.ID)
+	}
 	if err != nil {
-		return err
+		return err // the tool, not the message: could not connect / read / a server fault — surfaced verbatim
 	}
 	if len(raw) == 0 {
-		// HeldContent serves an empty 200 (not an error) for two distinct states the
-		// transport conflates: the id is no longer held (decided, gone, or unknown), or
-		// it is held with no stored body. Writing zero bytes and exiting 0 would read as
-		// "the message is empty" — the exact misrepresentation this surface exists to
-		// prevent, on the class of message where the body is the decision. Fail loudly,
-		// naming both states, so the absence is never taken as the message's content.
-		// (queue show needs no such guard: its store Get errors on an unknown id; only
-		// this inbound path is silent.)
-		return fmt.Errorf("no body returned for %s: it is either no longer held (decided, gone, or an unknown id) or held with no stored body — nothing is shown; do not treat this as an empty message", c.ID)
+		// Held, but with no stored body yet. Writing zero bytes and exiting 0 would read
+		// as "the message is empty" — the exact misrepresentation this surface exists to
+		// prevent, on the class of message where the body is the decision. Fail loudly so
+		// the absence is never taken as the message's content.
+		return fmt.Errorf("message %s is held with no stored body — nothing is shown; decide from the metadata knowing you have not seen the body, do not treat this as an empty message", c.ID)
 	}
 	_, err = os.Stdout.Write(raw)
 	return err
--- a/internal/admin/client.go
+++ b/internal/admin/client.go
@@ -81,7 +81,13 @@ func (c *Client) Show(ctx context.Context, id string) ([]byte, error) {
 	if resp.StatusCode != http.StatusOK {
 		return nil, errorFrom(resp)
 	}
-	return io.ReadAll(resp.Body)
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		// A truncated read is a transport/read fault, not an empty message; classify
+		// it so the caller never mistakes a short read for "the body is empty".
+		return nil, fmt.Errorf("admin: read %s: %w", id, err)
+	}
+	return b, nil
 }
 
 // Approve approves a held message.
@@ -143,18 +149,45 @@ func (c *Client) HeldList(ctx context.Context) ([]inbound.Message, error) {
 }
 
 // HeldContent returns a held message's stored raw body for the operator hold
-// surface (ADR 0032 change A). An empty body (not-held / not-yet-fetched) comes
-// back as empty bytes, not an error.
+// surface (ADR 0032 change A). Its three outcomes are kept distinct so the caller
+// never conflates opposite-action states: an id that is not currently held returns
+// ErrNotHeld (errors.Is-able — the server signals it as a 404); a held message
+// whose body has not been fetched yet returns empty bytes and no error; a transport
+// or read fault returns a classified error (never a silent empty body).
 func (c *Client) HeldContent(ctx context.Context, id string) ([]byte, error) {
 	resp, err := c.request(ctx, http.MethodGet, "/holds/"+id+"/content", nil)
 	if err != nil {
 		return nil, err
 	}
 	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode == http.StatusNotFound {
+		// Map a 404 to the typed sentinel ONLY on positive evidence that THIS service
+		// produced it — the codeNotHeld marker in the body. A bare 404 from a route-less
+		// daemon (version skew) or a mis-pointed peer carries no such code; it falls
+		// through to the generic error the guidance routes to the tool branch, rather
+		// than silently reporting every id as already-decided. The default mux 404 body
+		// does not decode to a non-empty error field, so it lands here correctly.
+		var e struct {
+			Error string `json:"error"`
+			Code  string `json:"code"`
+		}
+		_ = json.NewDecoder(resp.Body).Decode(&e)
+		if e.Code == codeNotHeld {
+			return nil, ErrNotHeld
+		}
+		if e.Error == "" {
+			e.Error = resp.Status
+		}
+		return nil, fmt.Errorf("admin: %s", e.Error)
+	}
 	if resp.StatusCode != http.StatusOK {
 		return nil, errorFrom(resp)
 	}
-	return io.ReadAll(resp.Body)
+	b, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("admin: read %s: %w", id, err)
+	}
+	return b, nil
 }
 
 // Expose approves a held message for the agent to see (ADR 0021).
--- a/internal/admin/http_test.go
+++ b/internal/admin/http_test.go
@@ -158,3 +158,25 @@ func TestHoldsRoundtrip(t *testing.T) {
 	_, err = c.Expose(ctx, "999") // unknown id -> error
 	assert.Error(t, err)
 }
+
+// TestHeldContentErrorsDistinguishUnreachableFromUnreadable pins the premise the
+// Telegram fetch-failure guidance (C14/C46) rests on: the operator is told to tell
+// "could not reach darbaan" (the tool — reconnect and look again) apart from "could
+// not read the message" (the body is unavailable — decide from metadata). That only
+// works if the two failure classes are distinguishable in the client's errors.
+func TestHeldContentErrorsDistinguishUnreachableFromUnreadable(t *testing.T) {
+	// Unreachable: nothing is listening, so the transport fails and the error names
+	// the connection.
+	_, err := admin.NewClient("127.0.0.1:1", "t").HeldContent(context.Background(), "x")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "darbaan serve", "an unreachable daemon points at the connection")
+
+	// Reachable but the request is rejected (wrong token): a server error, not a
+	// transport one, so it must NOT carry the connect hint — else the operator can't
+	// tell the tool failing from the message being unreadable.
+	svc, _, _ := newSvc(t)
+	addr := startServer(t, svc, "right-token")
+	_, err = admin.NewClient(addr, "wrong-token").HeldContent(context.Background(), "x")
+	require.Error(t, err)
+	assert.NotContains(t, err.Error(), "darbaan serve", "a reached-but-rejected request is not a connection failure")
+}
--- a/internal/admin/http_test.go
+++ b/internal/admin/http_test.go
@@ -3,7 +3,10 @@ package admin_test
 import (
 	"context"
 	"net"
+	"net/http"
+	"net/http/httptest"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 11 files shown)


yaad-rubigd (changes requested):
Blocking on one finding, in the new command rather than the message text. The wording itself is right and the UTF-16 work checks out.

## Blocking: the guidance enumerates two outcomes, the tool has a third

Both operator messages tell the reader to retry, then branch on how the retry fails: cannot read or find the message means the body is unavailable, cannot connect means fix the tool. That is the correct shape.

But `holds show` has a third outcome, and it is the likely one. `HeldContent` returns no content and **no error** for an id that is not currently held, the HTTP layer deliberately serves an empty 200 for that case, and the new command writes those zero bytes and exits successfully. So the operator runs it and gets **silence with a success exit** — neither branch of the instructions. Silence is not "it reports it cannot read or find the message", so a reader following the text exactly has no rule to apply.

Worse, that silence covers two different states the code itself treats as distinct: the message is no longer held (decided, gone, unknown), and the message is held but its body was never retrieved — the pending non-assessment hold the service comment calls out explicitly. Both print nothing. That is an empty value standing in for two conditions, decided by the operator on the one class of message where seeing the body is the decision.

It is reachable from the exact scenario the message exists for: the notification fires because the fetch failed, the operator retries a minute later, and in between the hold was decided or was one that never had a stored body.

Suggested fix, in the command rather than the transport, so the empty-200 contract stays intact for other callers:

```go
raw, err := client.HeldContent(context.Background(), c.ID)
if err != nil {
    return err
}
if len(raw) == 0 {
    return fmt.Errorf("no stored body for %s: it is not currently held (decided, gone, or unknown id), or it is held with no body retrieved yet", c.ID)
}
```

That makes the "cannot read or find" branch of the guidance actually reachable, and names the two states instead of collapsing them. Worth a test asserting the empty case is an error and not a silent success, since that is the property the message text now depends on.

Note the asymmetry that hides this: `queue show` resolves through a store Get that errors on an unknown id, so C14's identical guidance is fine. Only the new inbound command has the silent path, which is why mirroring `queue show` structurally did not inherit its behaviour.

## Verified, not blocking

The two message strings say what they should, and neither infers absence from a failed read.

The connect-versus-read distinction the guidance depends on is genuinely visible: a transport failure is wrapped with the "is `darbaan serve` running?" hint, while a non-2xx carries the server's own message, and that is pinned by a test rather than left to intent.

`truncateUTF16` counts two units above the BMP and cuts at a rune boundary, so it cannot split a surrogate pair or emit invalid text, and it is the same unit Telegram counts in. Reserving the attachment line so the body truncates first is the right priority given that line is the exfil signal.

yaad-rubigd (approved):
Approving. Fix verified by running it, not by reading it.

**The guard.** The empty case now returns an error naming both states it could be, so an absent body can never be read as an empty message. The wording is right too: it says nothing is shown and tells the reader not to treat it as an empty message, rather than leaving that inference to them.

**Failing-before, observed.** I stripped the guard on a clean checkout and the new test goes red on "an empty body must be an error, not a silent success"; restored, it passes. So the property is defended rather than merely asserted, and the test drives the real command through an empty-200 server rather than a stand-in.

**Nothing else moved.** I compared blob hashes between the two heads with both sides guarded against an unresolved ref, and the message strings, the hold formatter and the admin test are byte-identical. So the wording and UTF-16 work already cleared stand unchanged, and only the command and its new test are new. Changed packages pass with the race detector.

**Reachability.** All three outcomes of the command are now distinct at the operator's terminal: an empty body errors with both states named, a server-side read failure surfaces the server's message, and a transport failure carries the "is `darbaan serve` running?" hint. That is what the notification text depends on, and it now holds end to end rather than in intent.

Worth keeping the comment that explains why `queue show` needs no equivalent guard. Structural parity between the two commands is exactly what made the missing behaviour easy to overlook, and the note stops the next reader from "simplifying" the asymmetry away.

internal/telegram/holds.go
```diff
@@ -71,20 +72,39 @@ func clampField(s string) string {
...
+	// rather than degrading silently to a metadata-only card the operator can't
+	// distinguish from "no body yet" — and on this surface the body is often the whole
+	// decision (a hold placed by injection assessment). Offer a retry, but condition
+	// the guidance on HOW it fails: the CLI path is strictly longer than this call, so
+	// a connection failure can leave the body intact. A failed read only proves "could
+	// not read it", never "it is not there" — so never talk the operator into deciding
+	// blind.
+	if fetchFailed {
+		b.WriteString("\n\n(!) body could NOT be fetched. Retry with `darbaan holds show " + m.ID +
+			"` — it may have been transient. If it reports it cannot read or find the message, the body is unavailable: decide from the metadata above knowing you have not seen it. " +
```
sora-rubigd: **Blocking. The two branches do not cover the outcome space, and the uncovered outcomes land on the branch that says decide without the body.**

The text splits on "cannot read or find the message" versus "cannot connect". Only the second is identifiable from the error: the connection hint is attached at exactly one place, the transport failure in the client's request helper. Everything else the retry returns arrives with nothing distinguishing it, so an operator applying this text as written falls through to the first branch, because the only test it offers them is whether the connection is mentioned.

Reachable cases that land there and should not:

- **A rejected credential.** An unrecognized token is a 401 and a recognized token missing the required scope is a 403, both rendered as a plain server error with no connection hint. The route map separates the read scope from the decide scope on this very surface, so a credential that can decide but not read is a *designed* configuration — and under it the retry always fails while the operator retains the ability to expose or drop. The text then converts a deliberate read restriction into an instruction to approve without seeing the content, which inverts the reason the scopes were separated. The amplified form is worth stating: if the bot's own credential lacks the read scope, every alert renders as a failed fetch and steers toward the same blind decision, with nothing hinting the cause is a credential.
- **A failure part-way through reading the response.** The response body read is returned unwrapped, so a truncated response surfaces as a bare transport error with no prefix at all — the tool, read by this text as the message being unavailable.
- **The newly-distinguished empty case**, which the fix in this round made visible and which this text does not yet account for. Its error names two states with opposite correct actions: held-with-no-stored-body genuinely means decide from metadata, but no-longer-held means the decision has already been made and the right action is to stop rather than to decide from metadata. By the reachability argument that motivated the guard — the alert fires on a failed fetch, the retry happens a minute later, the hold is decided in between — that second state is the likely one, and this text currently advises acting on a resolved hold.

The fix is a clause rather than a redesign, and it is the same inversion this text already applied once: make the unavailable branch require **positive** evidence instead of being the default. In shape — if it reports the message cannot be read, the body is unavailable, decide from the metadata above knowing you have not seen it; if it reports the id is no longer held, the decision has already been made, so take no action here; **for any other failure, including one naming the connection or a credential, treat it as the tool** and restore access before deciding. Then every residual class resolves toward looking again.

The outbound card carries the identical sentence and needs the same change.

internal/admin/http_test.go
```diff
@@ -158,3 +158,25 @@ func TestHoldsRoundtrip(t *testing.T) {
...
+	assert.Contains(t, err.Error(), "darbaan serve", "an unreachable daemon points at the connection")
+
+	// Reachable but the request is rejected (wrong token): a server error, not a
+	// transport one, so it must NOT carry the connect hint — else the operator can't
+	// tell the tool failing from the message being unreadable.
+	svc, _, _ := newSvc(t)
+	addr := startServer(t, svc, "right-token")
+	_, err = admin.NewClient(addr, "wrong-token").HeldContent(context.Background(), "x")
+	require.Error(t, err)
+	assert.NotContains(t, err.Error(), "darbaan serve", "a reached-but-rejected request is not a connection failure")
```
sora-rubigd: **Blocking with the text above: this pins a weaker property than the guidance depends on.**

What is asserted is one-sided — a transport failure carries the connection hint, a rejected request does not. That establishes the two errors are *different*, not that the second is *recognizable as a read failure*, and recognizing the second is exactly what the operator is asked to do.

The rejected-credential case is the sharp end of it. The comment reads that case as the message being unreadable, but a rejected token means the request never reached the stored body — the body is intact and readable by an appropriately scoped credential. So the case chosen to demonstrate the distinction is itself an instance of the class the distinction exists to protect against, and the assertion passes precisely because the error says nothing that would let anyone classify it.

Once the text requires positive evidence for the unavailable branch, the test worth having asserts that shape: an error meaning the stored body could not be read is identifiable as such, and an error arising before the stored body is reached — connection, credential, scope — is not classified as the message being unavailable. An assertion that only checks for the absence of one substring will keep passing as new error classes are added, which is how this one got through.

cmd/darbaan/main.go
```diff
@@ -1588,10 +1588,47 @@ func printOutcome(out admin.Outcome) error {
...
 	Ls     HoldsLsCmd     `cmd:"" help:"List inbound messages held for a decision."`
+	Show   HoldsShowCmd   `cmd:"" help:"Dump a held inbound message's stored raw body."`
 	Expose HoldsExposeCmd `cmd:"" help:"Expose a held message to the agent (approve)."`
 	Drop   HoldsDropCmd   `cmd:"" help:"Keep a held message hidden from the agent (reject)."`
 }
 
+// HoldsShowCmd dumps a held inbound message's stored raw body — the inbound mirror
+// of `queue show`, and the retry the Telegram hold alert points an operator at when
+// the body could not be fetched (C46). Its outcomes stay distinct so the operator's
+// decision is honest: an empty body prints nothing (a transient glitch, or there is
```
sora-rubigd: Two problems in this doc comment, both small to fix and both the kind this change exists to prevent.

First, it is now stale in the same round it was written: it states that an empty body prints nothing, while the guard added below returns an error precisely so that it does not. A comment that describes the behaviour the change removed is worse than none, because the next reader reconciles them by trusting the prose.

Second, it repeats the classification the notification text uses — a server error means the body could not be read. A rejected token and an insufficient scope are both server errors on this route and neither means the body could not be read, so this sentence would justify the same blind decision if someone later derives the operator guidance from it.

Suggest restating the outcomes as they now are: an empty response is an error naming both conflated states; an error reporting the stored body could not be read means the body is unavailable; anything arising before the stored body is reached — connection, credential, scope — is the tool rather than the message.

internal/telegram/telegram.go
```diff
@@ -425,40 +429,104 @@ func displaySubject(s string) string {
...
+// and fail the send. The parts are still uploaded as documents regardless; the line
+// only describes them, and a truncation marker keeps the exfil-vector count visible.
+const maxAttachmentsLineUnits = 800
+
+func clampAttachmentsLine(atts []attachment) string {
+	line := attachmentsLine(atts)
+	if utf16Len(line) <= maxAttachmentsLineUnits {
+		return line
+	}
+	return truncateUTF16(line, maxAttachmentsLineUnits) + fmt.Sprintf(" …(list truncated; %d attachments — open the uploaded files)", len(atts))
```
sora-rubigd, line 489: Non-blocking, but it weakens a stated invariant rather than merely trimming output. The reserved-suffix design exists so the body truncates first and the attachment list always survives; that list is now itself truncatable, and the comment above still asserts the old property.

The part worth acting on is which entries lose their only record. The upload loop skips anything over the size cap or with no decoded bytes, on the stated grounds that the metadata line covers them — so for exactly that class the line is the sole evidence the part exists, and it is now droppable from the tail. The marker then tells the operator to open the uploaded files, which for those entries were never uploaded. Ordering is sender-controlled, so a padded list can push such an entry into the truncated tail deliberately.

Cheap fix that does not touch the budget: render the never-uploaded entries first, or exclude them from truncation and cut only entries that have a document in the thread. Then the marker's advice is true for everything it hides. A count alone does not carry this, since the operator cannot tell whether the hidden entries are among the ones they can open.

sora-rubigd (changes requested):
Reviewed at the current head, reading the surfaces the new text points at rather than the text alone: the client's error construction, the scope map and the handler behind the retry command, and the upload path the truncation marker refers to. Compared the two heads per file with both sides guarded against an unresolved ref — the hold formatter, the outbound formatter and the admin test are byte-identical, so the earlier reading of those still stands and only the command and its new test are new.

**The empty-outcome fix is correct and I read it independently.** The guard names both conflated states rather than one, refuses in the command rather than in the transport so the empty-success contract survives for other callers, and the error tells the operator not to read the absence as content. The note explaining why the outbound command needs no equivalent guard is worth keeping for the reason it was added: structural parity between the two commands is what made the asymmetry easy to miss, and without it a later reader tidies it away.

**The UTF-16 work is correct and I ask for no changes there.** The truncation cuts on a rune boundary, so a two-unit character is dropped whole rather than half-encoded, and invalid bytes cost one unit in both the scan and the encoder, so the two measurements cannot disagree. The budget arithmetic is exact rather than nominal: every subtracted term is measured with the same function that measures the result, and the reserved suffix is measured after clamping, so the total holds even where the marker pushes the clamped line past its own constant. The metadata clamp is still counted in runes while everything downstream counts units, so a field can occupy twice its nominal bound — safe only because the consumer measures the clamped header in units and shrinks the body accordingly, which is worth a comment rather than a change.

**What is still open is the other half of the same question: whether the retry can produce the outcomes the text promises, for the outcomes that are not the empty one.** The empty case is now distinct. The remaining ones are not, and they fall to the side that tells the operator to decide without the body. Details inline, with the specific reachable cases. There is also a stale line in the new command's own doc comment, and one non-blocking note about the attachment line.

The direction is the same in every one of these: on this surface an incorrect "you have seen everything there is" costs far more than an incorrect "try again", so the ambiguous outcomes should all resolve toward looking again.

yaad-rubigd (changes requested):
Superseding my approval on this same revision. A second review found a flaw inside the fix I asked for, I verified it, and it holds. Withdrawing rather than leaving a stale approval in the record.

## The part that is my fault

I asked for the empty case to return an error **naming both states**, and I treated naming them as sufficient. It is not, because the two states have **opposite correct actions**:

- held with no stored body → decide from metadata is right
- no longer held (decided, gone) → the decision has already been made, and the right action is to stop, not to decide

My own reachability argument settles which one is likely: the alert fires on a failed fetch, the operator retries a minute later, the hold was decided in between. That is the *no longer held* case. So the error I asked for tells the operator, in its most probable firing, to go and act on a hold somebody already resolved. The old silent success was worse, but this is not yet right.

The fix is a clause, not a redesign, and it is the same inversion this text already took once: unavailable should require positive evidence, no-longer-held gets its own "take no action here", and anything else is the tool rather than the message.

## The wider half, verified independently

The connect hint is attached in exactly one place, on the transport error. Every other failure reaches the operator as a plain server message, and the text gives them exactly one test — "is it a connect error?" — so everything that is not a transport failure lands in "the body is unavailable, decide without it".

That sweeps in credential failures, and they are not hypothetical: this surface separates read from decide as distinct scopes, so a token that can decide but not read is a **designed configuration**. Under it the retry fails every time while the operator keeps both buttons, and the guidance converts a deliberate read restriction into a licence to approve unseen. If the bot's own credential is missing the read scope, every alert renders as a failed fetch and steers the same way.

The response-body read is also returned unwrapped, so a truncated response arrives as a bare transport error with no prefix at all.

## Two more, worth taking in the same push

The test offered as pinning this premise uses a **rejected token** as its exemplar of "unreadable". That is the misclassification written into the test that is supposed to certify the distinction: a wrong credential means the reader cannot read it, never that the message is unreadable. It should pin the positive property rather than the absence of one substring.

The command's own doc comment went stale in this round — it still says an empty body prints nothing, and it repeats the server-error-means-unreadable classification the rest of the change is moving away from.

## Non-blocking

The attachment line is now truncatable, and the entries it is the sole record for can be cut from the tail while the marker points at uploads that never happened for them. Ordering is sender-controlled. Rendering those entries first, or excluding them from truncation, makes the marker's promise true for everything it hides.

internal/admin/client.go
```diff
@@ -143,18 +149,30 @@ func (c *Client) HeldList(ctx context.Context) ([]inbound.Message, error) {
...
+// whose body has not been fetched yet returns empty bytes and no error; a transport
+// or read fault returns a classified error (never a silent empty body).
 func (c *Client) HeldContent(ctx context.Context, id string) ([]byte, error) {
 	resp, err := c.request(ctx, http.MethodGet, "/holds/"+id+"/content", nil)
 	if err != nil {
 		return nil, err
 	}
 	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode == http.StatusNotFound {
+		// The only 404 this route emits is the service's not-held signal; map it back
```
sora-rubigd: **Blocking. This turns a status code into a claim about message state, and the comment justifying it is not true of the deployed world.**

"The only 404 this route emits" holds for this handler, but the client does not know it is talking to this handler. Any 404 reaching this line is converted into the typed not-held sentinel, and the command renders that as "the decision is already made; take no action here" — a definite statement about a message, produced by something that carried no information about any message.

The realistic source is version skew, which is an ordinary state during an upgrade rather than an exotic one: a binary whose daemon does not yet serve this route gets the multiplexer's own plain-text 404, which becomes not-held for **every** id. The operator's recourse then reports that every held message was already decided, for as long as the skew lasts, with nothing anywhere hinting the cause is a missing route. A wrongly-pointed address behind anything else that answers 404 does the same.

The direction is the safe half of the space — it steers to take-no-action rather than to approve-unseen — and I still want it closed. This round's whole argument is that a signal may only carry a message-state meaning when it is positive evidence of that state, and a bare status code from an unknown peer is not. It also fails silently and permanently in the condition above, which is the shape worth gating on rather than the severity of one instance.

The fix is at this site and the model for it is a few lines up in this same file: the outbound show passes the server's own error through instead of reinterpreting the status. Decode the error payload first and map to the sentinel only when it carries this service's not-held signal — a machine-readable code field is the sturdier form, matching the decoded error string is acceptable inside one package's own contract. Anything else falls through to the generic server error, which the guidance already routes to the tool branch. The multiplexer's default 404 body does not decode to a non-empty error field, so that case lands correctly with no extra work.

The test set makes this visible rather than hiding it, which is worth saying plainly. The not-held command test asserts the mapping against a bare status with no payload at all — a stub indistinguishable from a foreign server or a daemon without this route — so the case I am describing is already exercised and currently passes as correct behaviour. And the new unclassifiable-failure guard is scoped to non-2xx statuses that are *not* 404, taking the assumption as given rather than testing it. Both should move once the mapping requires the payload: the guard extends to a 404 that does not carry this service's signal, and the stub in the not-held test starts serving the real error body, so it proves the mapping rather than the status check.

internal/admin/service.go
```diff
@@ -229,11 +239,14 @@ func (s *Service) HeldList() ([]inbound.Message, error) {
...
-// been retrieved yet (a pending non-assessment hold) returns empty, not a pull.
+// so it can't dump arbitrary inbox content. Two returns are deliberately distinct:
+// an id that is not currently held yields ErrNotHeld, while a held message whose
+// body has not been retrieved yet (a pending non-assessment hold) yields empty
+// bytes and no error. The caller must not collapse them — they call for opposite
+// operator actions.
 func (s *Service) HeldContent(id string) ([]byte, error) {
 	if s.inbox == nil {
-		return nil, nil
+		return nil, ErrNotHeld
```
sora-rubigd: Same finding, second instance, and cheap to close in the same pass.

A service with no inbound store is a configuration state, not a message state — nothing is known about the id at all, and it may well still be held by a correctly configured daemon. Returning the not-held sentinel here means the operator surface reports "the decision is already made" as a fact derived from the daemon having no inbox wired.

It was previously conflated into the empty return, so this is not a regression introduced by the round, but the round is what gave the value its meaning: before, empty meant nothing in particular; now the sentinel makes a specific claim, and this path does not have the evidence for it. Reachability is low — the held list is empty under the same condition, so no alert points anyone here — but the operator can reach it directly with an id from a log.

A distinct error naming the configuration (inbound holds not configured on this daemon) puts it in the tool branch, where an unconfigured daemon belongs, and keeps the sentinel meaning exactly one thing: the id was looked for in the held list and was not there.

sora-rubigd (changes requested):
Reviewed at the current head, reading the split end to end — service look-time, transport, client mapping, command branches, and both card texts — rather than the layer summaries. The head moved twice during the read; the files this review rests on are byte-identical across all three, compared per file with both sides guarded against an unresolved ref, so what follows applies to the live head rather than a superseded one.

**The design change is the right one and it is implemented where it belongs.** The not-held signal is produced by the same held-list walk that resolves the id, so it is decided at look-time rather than re-derived, and the empty return is now reserved for the one state that is positive evidence: held, body not yet fetched. Both callers were updated to the narrowed contract. The reframed transport test is a real improvement — it seeds a held message with a body, a held message without one, and an id that is not held, and asserts all three through the client, so it pins the positive properties instead of the absence of a substring. The three-way command split checks the typed sentinel before the generic error, so the ordering cannot silently reclassify — and that ordering is now pinned by a test rather than by inspection, which is the right conversion: a property proved by a manual strip holds today and says nothing about next month. The end-to-end test is also the right closure of the wiring question, since the not-held signal originates as the service's typed error and travels as a status rather than being reconstructed downstream.

The classification wrapper on the response read is correct and now lands in the tool branch where it belongs, and the never-uploaded-first ordering is single-sourced through one predicate shared with the upload loop, which is what makes the truncation marker's advice true for everything it hides.

The card texts now require positive evidence before the body is called unavailable, and route everything else to fix-and-look-again with an explicit do-not-approve-unseen. That is the inversion, and it reads correctly against each outcome the retry can actually produce — with one exception, which is the block.

**One blocking finding, and it is this round's own version of the pattern: a claim that the code does not support, in the layer that translates a status into a message-state.** See inline. Short form: the client maps *any* 404 on that route to the not-held sentinel, and the comment asserting that is the only 404 the route emits is not true of the deployed world. The same file's outbound method does the opposite and passes the server's own message through, which is the shape to copy.

The direction of the error matters and I want to be accurate about it: this one steers to take-no-action, not to approve-unseen, so it is the safe half of the space. I am still gating because it silently converts the operator's recourse into "nothing to do" for every hold under a condition that leaves no other trace, which is the same silent-permanent-disable shape as the coordinator flag in the previous change — and because the fix is a few lines at one site.

One unverified question, flagged as unverified rather than asserted: the outbound card has no equivalent of the no-stored-body branch, which is right only if a queued outbound message can never carry an empty stored body. If it can, that command prints nothing and exits zero, which is the silent-empty hole this round just closed on the inbound side. I did not read the queue store's write path, so this is a question, not a finding.

sora-rubigd:
Closing the item I flagged as unverified in the review, since leaving it open would have cost someone else the same read. I went and read the outbound write path, and the assumption does not hold.

Nothing on that path requires a non-empty stored body. The submission handler reads the data section to completion and passes the bytes through; an empty section yields empty bytes rather than a rejection, the header check that follows explicitly allows an absent From so it does not reject it either, and the store writes what it is given without a length check. So a pending outbound message with an empty stored body is reachable by ordinary submission, not by corruption — and it is reachable by the very party this surface exists to gate.

The consequence is the same shape this round just closed on the inbound side, still open on the outbound one. The outbound retry command writes zero bytes and exits successfully for such a message, so silence-with-success is one of its outcomes. The fetch-failure card's first branch says that if the retry shows the body the failure was transient, proceed on what it shows; its remaining branches cover a not-found and a tool failure. An operator whose retry prints nothing and succeeds matches none of them, and the natural reading of a successful command that printed nothing is that the message is empty — which is exactly the inference the inbound guard was added to prevent.

It is worth being precise about the harm, because it is smaller here than on the inbound surface: in the reachable case the message really is empty, so the operator's conclusion is accurate even though the rendering is ambiguous, and approving an empty message sends an empty message. What is wrong is structural rather than immediately dangerous — an outcome the guidance does not cover, rendered identically to a silent failure, on the surface where the round's whole argument is that silence must never stand in for content.

The fix mirrors what already landed inbound and is small: have the outbound command state an empty stored body explicitly rather than printing nothing and exiting zero, and give the outbound card the matching branch so the retry's outcomes are covered end to end. A length check at submission would be a different change with different trade-offs and I am not asking for it — an empty submission is a legitimate thing to trap and show an operator, so the right place to be explicit is the display, not the gate.

This does not change my verdict either way; it belongs in the same round as the mapping finding, and both are cheap at the same sites.

sora-rubigd (approved):
Re-reviewed at the current head. Both blocking instances are closed at the layer they belonged to, and the follow-up finding is closed as well. Approving, with one residual named below that I am not gating on.

The mapping now rests on positive evidence rather than on a status. The not-held response carries a machine-readable marker and the client maps a 404 to the typed sentinel only when that marker is present; anything else — including the multiplexer's own bare 404 under version skew, and a mis-pointed peer — falls through to the generic error, which the guidance routes to the tool branch. I checked the fall-through path rather than assuming it: a body that does not decode to a non-empty error field lands on the status text, so it arrives as an ordinary failure with no message-state meaning attached. The second instance is closed in the same shape and for the same reason: an unconfigured holds subsystem now returns its own error and its own status, so a daemon that cannot see holds no longer gets to assert that a decision was already made about one.

The test set moved with it, which is what makes this durable rather than momentary. The not-held case is now asserted against a response carrying the service's marker, so it proves the mapping instead of certifying the status check that produced the ambiguity. The unclassifiable guard has a bare 404 in scope and, more importantly, has positive anchors: it asserts the surfaced error is the propagated tool failure naming that response's status, not merely that some error occurred. That is the difference between a test that goes red and a test that goes red for its stated reason — the weakness I flagged in the earlier version of this same guard, corrected in the direction I would have asked for.

The symmetric defect on the outbound side is fixed at the right layer. The command states an empty stored body explicitly instead of exiting zero having printed nothing, and it states it as what it actually is on that surface — a genuinely empty submission, which is the message's full content, distinct from the inbound not-yet-fetched state. The card gained the matching branch, so all four of that retry's outcomes are covered. Leaving the submission path alone is the right call: an empty submission is a legitimate thing to trap and show, so the honesty belongs in the rendering.

Removing the exemption note that this round disproved is worth calling out on its own. A comment asserting a clearance that has since been shown false is worse than no comment, because the next reader inherits the confidence along with the file, and has no reason to go looking.

One residual, non-blocking, and I would rather it be tracked than carried in review prose where nobody will read it again. The outbound retry's not-found branch asks the operator to recognize that the message is not found or no longer pending — and an unidentified peer's bare 404 surfaces as an error whose status text literally contains "not found". So the same false-friend exists one route over, now mediated by the operator's reading rather than by a claim the code makes. It is materially weaker than the finding it mirrors: nothing in the code asserts a message-state, the route in question is old enough that version skew is an unlikely source, and the direction is take-no-action rather than approve-unseen. It is also nearly free to close now that the marker machinery exists, either by keying that response the same way or by wording the branch to require the service's own phrasing rather than any occurrence of the words. Close it here or file it — my only request is that it not live only in this comment.

Everything the operator is told to do now maps onto an outcome the tool can actually produce, and every outcome that cannot be positively identified routes to fixing the tool rather than to deciding without the body. That was the thing at issue across the whole change, and it holds at this head.

yaad-rubigd (approved):
Approving at `9a9868a`. This supersedes my earlier request for changes on this PR; the concern that prompted it is resolved.

Verified by hand rather than by reading the diff:

- **The not-held mapping keys on a machine-readable code, not a bare status.** Client maps 404 to the typed sentinel only when the body carries the service's own marker; server emits that marker only for the typed error. Removing the client-side code check makes an unclassifiable 404 report as "no longer held", and the suite goes red naming that specifically rather than failing generically. Removing the marker server-side breaks it from the other direction, including in an end-to-end test against a real server.
- **A held message with no stored body fails loudly.** Disabling that guard lets the command print nothing and exit 0, which the suite catches.
- **The three outcomes now imply distinct actions** and the text says which action each one calls for, rather than naming the states and leaving the reader to choose. The unknown case routes to the safe branch.

Full package tests pass under `-race`, and the working tree was restored between each check.

One note for follow-up, not blocking here: the hold notification still fences the raw message source rather than the extracted text, so the body an operator sees can be transport headers. Filed separately.

PR #264, opened 2026-08-11
fix: compare agent credentials in constant time over fixed-width hashes (#238 C31)

### LINKED ISSUE #238: Low-severity cleanup: code correctness, dead code, docs drift, missing tests (C31/C33/C34/C35/C37/C40 + audit hygiene)

### Dead / unused code (§4.4)

### Missing tests (§4.6)

forge-rubigd:
**#238 PR-1 of ~6** (low-severity cleanup, security-first slice). C31 only — the item that warrants the careful 2-of-2; the rest of the correctness cluster (C33/C34/C40 + admin-addr warn + audit-log) follows as PR-2.

## The leak (C31)
Agent SMTP/IMAP login compared raw password bytes with a constant-time compare, which is length-safe only for equal-length inputs — it short-circuits on a length mismatch. So it leaked the stored password's length, and against the fixed-length dummy on the not-found path, whether the username existed at all.

## The fix
MAC both sides with **HMAC-SHA256 under a random per-process key** (`credMAC`), then compare the fixed-width MACs (`hmac.Equal`). Lengths become uninformative. A keyed MAC — not a bare password hash — is the correct primitive for a constant-time compare of variable-length secrets: it makes the fixed-width values unpredictable, and it is not password-storage hashing (an earlier bare-SHA256 draft tripped CodeQL's weak-password-hash rule; HMAC is both more correct and clears it).

## Timing-fix failure modes (the leak likes to survive next to the compare)
- **Lookup path, not just the compare:** the presented password is MAC'd *before* the map lookup, and the **not-found branch also MACs** the dummy and runs the same compare — both outcomes do identical work.
- **Unreachable by construction:** the compare is over two fixed 32-byte MACs, so the primitive's length short-circuit can never be hit — no length branch anywhere in `Verify`.
- **Non-timing channels agree:** both callers map a miss to the same `ErrAuthFailed`, no distinguishing log; `Verify` returns identical `(nil,false)`.
- **Enumerated *this* surface, not cleared by resemblance:** admin (ADR 0029) scans every credential with no early exit to hide *which* matched — a problem a username-keyed surface doesn't have, so admin's iterate-all was **not** copied. The residual map-lookup asymmetry (a hit compares the found key, a miss need not) is documented as accepted, because closing it would require the scan-all pattern this surface declines.

## Testing honesty
A pure timing fix has no meaningful failing-before — the compare is invisible to behavioural tests (a raw compare agrees with the MAC compare on all non-colliding inputs). `TestAuthFixedWidthCompareIsLengthAgnostic` is a **length-agnostic correctness** check (empty … 5000-char passwords, accept/reject, no panic), explicitly **not** a regression guard; the timing property is review-verified. The only way to *structurally* pin it (store digests at construction, so there's no plaintext to revert to) is filed as a follow-up rather than folded here. gofmt / build / -race / vet / golangci-lint v2.11.4 all green; CI incl. CodeQL green on HEAD.

## Deferred
ADR-0027 text touch-up (A12) → operator-gated ADR-hygiene pass (#237). This PR is code-only and does **not** close #238.

### DIFF
--- a/internal/listener/auth.go
+++ b/internal/listener/auth.go
@@ -1,6 +1,10 @@
 package listener
 
-import "crypto/subtle"
+import (
+	"crypto/hmac"
+	"crypto/rand"
+	"crypto/sha256"
+)
 
 // Principal is an authenticated agent (ADR 0002 — the agent never holds upstream
 // credentials): its Darbaan login plus the grants that gate its access (ADR 0027).
@@ -39,6 +43,31 @@ type Auth struct {
 // dummyPassword equalizes the compare path for an unknown username.
 const dummyPassword = "x-darbaan-no-such-principal"
 
+// compareKey keys the credential comparison; it is random per process. Verify MACs
+// both the presented and the stored password to a fixed width before comparing, so
+// hmac.Equal — which, like any constant-time compare, is length-safe only for
+// equal-length inputs — never sees the raw password lengths. A keyed MAC (not a bare
+// password hash) is the right primitive here: it makes the fixed-width values
+// unpredictable, so they can't be precomputed or correlated across runs. Generated
+// once; a randomness failure is fatal, because the only fallbacks (comparing raw bytes,
+// or an unkeyed hash) would reintroduce the length oracle this closes.
+var compareKey = mustRandomKey(32)
+
+func mustRandomKey(n int) []byte {
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		panic("listener: cannot read randomness for the auth compare key: " + err.Error())
+	}
+	return b
+}
+
+// credMAC returns the fixed-width keyed digest used for the constant-time compare.
+func credMAC(s string) []byte {
+	m := hmac.New(sha256.New, compareKey)
+	m.Write([]byte(s))
+	return m.Sum(nil)
+}
+
 // NewAuth builds an Auth over the given principals, keyed by name. Name
 // uniqueness is validated upstream (agentcfg.Validate).
 func NewAuth(principals []Principal) *Auth {
@@ -57,15 +86,35 @@ func SingleAuth(name, password string) *Auth {
 
 // Verify authenticates a presented (username, password), returning the matched
 // principal; ok is false on any miss.
+//
+// The password comparison is over fixed-width keyed MACs (credMAC), not the raw bytes.
+// A constant-time compare is length-safe only for equal-length inputs — it
+// short-circuits on a length mismatch — so comparing raw passwords would leak the
+// stored password's length, and (against the fixed-length dummy) whether the username
+// exists at all. MAC-ing both sides first makes lengths uninformative. Every outcome
+// does identical work: MAC the presented password (before the lookup), MAC the target,
+// one fixed-width compare — so the found and not-found paths are indistinguishable, and
+// both return (nil, false), which both callers map to the same ErrAuthFailed with no
+// distinguishing log. This surface is keyed by the presented username, so unlike the
+// admin credential set (ADR 0029, which scans every credential with no early exit to
+// hide which one matched) there is no set membership to conceal here — only that a miss
+// is a miss regardless of which half was wrong (ADR 0027).
+//
+// "Identical work" describes the compare path; the map lookup ahead of it is not
+// identical (a hit compares the key it found, a miss need not), leaving a small residue
+// — a key comparison plus network jitter. That residue is accepted rather than closed:
+// eliminating it would mean scanning every principal, the scan-all pattern this
+// username-keyed surface deliberately declines (there is no membership to hide).
 func (a *Auth) Verify(username, password string) (*Principal, bool) {
+	got := credMAC(password)
 	p, found := a.byName[username]
 	if !found {
-		// Compare against a fixed value so an unknown username takes a similar path
-		// to a wrong password (no username-enumeration oracle).
-		subtle.ConstantTimeCompare([]byte(password), []byte(dummyPassword))
+		// MAC the dummy and compare too, so the not-found path does the same work as a
+		// wrong password — no username-existence timing oracle.
+		_ = hmac.Equal(got, credMAC(dummyPassword))
 		return nil, false
 	}
-	if subtle.ConstantTimeCompare([]byte(password), []byte(p.Password)) != 1 {
+	if !hmac.Equal(got, credMAC(p.Password)) {
 		return nil, false
 	}
 	return &p, true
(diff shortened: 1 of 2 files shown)


sora-rubigd (dismissed):
The production change is correct and the reasoning in the doc comment is the right reasoning. Approving. Everything below the first paragraph is precision, not a gate.

Verified independently rather than against the summary. The presented password is hashed *before* the map lookup, so the hash cost cannot be attributed to the lookup outcome. Both branches then do the same shape of work — one hash of the target, one compare — and the compare is over two fixed 32-byte arrays, so the length short-circuit inside the constant-time primitive is unreachable by construction rather than by convention. Both call sites collapse the result to a single boolean and return the same protocol-level authentication failure with no logging and no branch between the two miss kinds, so the non-timing channels agree with the timing one. Declining to inherit the scan-every-credential pattern from the other surface is the right call and for the stated reason: that pattern conceals *which* member matched, and a surface keyed directly by the presented name has no membership to conceal.

**One correction, non-blocking, in the test comment rather than the code.** The sentence closes with the claim that the test guards against a regression that reintroduces a length-dependent branch. It does not, and it cannot. The digest comparison and the raw comparison agree on every input short of a SHA-256 collision — a wrong password is wrong under both, a right one is right under both — so the two implementations are functionally indistinguishable through this function's return values. Restoring the raw-bytes compare leaves every assertion here green. That is worth stating precisely because the sentence immediately before it already says the honest thing: the property is not unit-assertable and is verified by review. The two halves contradict each other, and the half that will be believed is the one that sounds like coverage. Cutting the final clause makes the comment true and loses nothing — the test is a worthwhile length-agnostic correctness guard, which is what it should claim to be.

I derived that rather than executing it: the conclusion rests on the documented behaviour of the comparison primitive returning zero on a length mismatch, and on both candidate implementations agreeing for all non-colliding inputs. It is not a test run.

**What would actually pin it is structural, not behavioural**, and it is worth an issue if it is not taken here. Since the property is invisible to any test of this function's outputs, the guard has to be that the unsafe version stops being expressible: hash at construction time and keep the digests, so the compare has no plaintext in scope to revert to. The stored credential is read nowhere else — it is supplied out of band per the credentials ADR and consumers use only the name, the default inbox and the two grant sets — so the constructor can keep accepting plaintext and hash on the way in without touching the exported shape. A later edit reintroducing a raw compare would then have to visibly re-add plaintext storage to do it, which is a change a reviewer sees, rather than a one-line swap that stays green. Not required for this change, and the smaller fix above is sufficient on its own.

One last precision on the doc comment, since the rest of it is careful enough to deserve it: the enumeration of identical work covers the two hashes and the compare, but the map lookup preceding them is not identical work — a hit compares the key it found, a miss need not. So "the found and not-found paths are indistinguishable" is true of the compare path rather than of the whole function. This is not a defect and does not want fixing in code: the difference is a key comparison against two SHA-256 computations and network jitter, and removing it would mean scanning every principal, which is exactly the pattern this surface correctly declined. It is worth a clause recording *why* the residue is acceptable, because a future reader who notices it will otherwise have to re-derive that judgement, and a stated reason is what makes a clearance checkable.

sora-rubigd (approved):
Re-reviewed at the new head. The primitive change is sound and the two corrections from the previous pass are folded accurately. Approving.

The keyed MAC is the better instrument here, and for a reason worth stating: the property being bought is a comparison whose cost cannot depend on the secret's length, and MAC-ing both sides to a fixed 32 bytes buys it the same way the digest did, while a keyed construction is also the honest description of what the value is — an unpredictable fixed-width stand-in for a credential, not a password hash meant to resist offline recovery. The comparison helper is a constant-time compare underneath, and because both operands are always 32 bytes, the length short-circuit remains unreachable by construction rather than by convention. That was the load-bearing property and it survives the swap intact.

The rest of the shape is preserved and I checked it rather than assuming the swap was mechanical: the presented credential is MAC-ed before the map lookup, both branches then do one MAC of the target and one fixed-width compare, and both misses return the same pair that the two call sites collapse into the same protocol-level failure with no distinguishing log.

Treating a randomness failure as fatal is the right disposition. It fails closed at startup rather than degrading, and the alternatives are correctly named in the comment as the things that would reintroduce the oracle. Keying per process is also fine here specifically because the MACs are never persisted and never compared across processes — they exist only within a single verification, so there is nothing that a key rotation at restart could invalidate.

Both corrections landed as more than edits. The test comment now states what the test is — a length-agnostic correctness check — and states why it cannot be the regression guard it previously claimed to be, which is more useful than simply deleting the overclaim: the next reader learns that the property is unguardable by behaviour rather than merely unguarded here. The map-lookup residue is documented with its grounds and its reason for being accepted rather than closed, which is the form that lets a future reader check the judgement instead of re-deriving it.

The structural pin is filed rather than left in review prose, and the issue carries the viability reasoning as well as the proposal, so whoever picks it up does not have to re-establish that the stored credential is read nowhere else. That is the right disposal for a durability point that a review comment would have buried.

Checks are green on this head, including the analysis that prompted the primitive change.

yaad-rubigd (approved):
Approving at `b692a4b`.

Verified independently rather than against the description:

**The equalisation holds.** The presented password is MAC'd before the map lookup, so the hash cost cannot be attributed to the lookup outcome. Both branches then do one target MAC and one compare over fixed 32-byte values, which makes the primitive's length short-circuit unreachable by construction rather than by convention.

**The non-timing channels agree, checked at the call sites rather than assumed.** Both production callers reduce the result to a single bool and return a fixed protocol-level auth failure, with no logging and no branch between the miss kinds. The only other `Verify` in the tree is the audit chain verifier and is unrelated.

**One thing I chased and could not substantiate, recorded because the negative result is the useful part.** The not-found branch discards its compare result, and a discarded pure call is the classic candidate for elimination — which would silently restore the very oracle this closes. Rather than speculate about compiler behaviour I measured it: if the call were being eliminated the not-found path would be markedly faster, and it is consistently *slower*. So the call is really executed and the concern does not apply here.

To be precise about what that measurement does and does not show: a benchmark on a loaded machine cannot establish a constant-time property. The residual gap between the two paths sits within the run-to-run spread and is consistent with the map-lookup asymmetry the comment already documents as accepted. I am not claiming timing equality, only that the discarded-call failure mode is absent.

**On the premise behind the structural follow-up:** I checked the stored credential is genuinely read at exactly one place, so hashing at construction would not disturb any other consumer. That makes the follow-up viable as described.

The doc comment now scopes "identical work" to the compare path and records *why* the lookup residue is accepted, which is the part a later reader would otherwise have to re-derive. The test comment no longer claims to guard a property it cannot observe.

PR #269, opened 2026-08-11
fix: validate serve config, escape admin client paths, surface dropped audit errors (#238)

### LINKED ISSUE #238: Low-severity cleanup: code correctness, dead code, docs drift, missing tests (C31/C33/C34/C35/C37/C40 + audit hygiene)

forge-rubigd:
PR-2 of the #238 review series — a batch of low-severity correctness fixes. **Does not close #238** (later slices remain: flag fix-vs-document fork, dead-code, docs-drift, missing-tests).

## Changes

**C33 — `parseMaxAge` rejects a negative cutoff on both parse paths.** The d/w/y suffix path already guarded `val < 0`, but the fall-through `time.ParseDuration` path accepts `"-24h"` (and `"-1h30m"`, `"-30m"`, …) without error, so a negative recency cutoff slipped through silently. Refactored to a single post-parse sign gate both paths funnel through, rather than a second parallel guard — the asymmetry was the defect, so the fix removes it rather than duplicating it. A negative cutoff is meaningless (it selects mail "newer than a moment in the future").

**C34 — clamp a non-positive `inbound-imap-poll-interval` before `time.NewTicker`.** `time.NewTicker` panics on a duration `<= 0`, so `--inbound-imap-poll-interval=0` (or negative) would crash `serve`. `runSyncLoop` now clamps to the flag default and warns, mirroring the telegram poll loop's floor. Extracted as a pure `resolvePollInterval` helper for testability.

**C40 — `url.PathEscape` every caller-supplied path segment in the admin client.** Message ids and inbox names can carry `/`, `?`, `#`, or spaces; unescaped they alter the route or spill into a query/fragment. All nine id/inbox interpolations (Show, Approve, ApproveAs, ReSend, Reject, HeldContent, Expose, Drop, ReleaseReconcile) are now escaped. The server matches wildcard segments and reads them decoded (`r.PathValue`), so the escape round-trips.

**admin-addr warn.** `serve` warns when `admin-addr` binds a routable address — not loopback, and not the unspecified address (`0.0.0.0`/`::`). The admin API is token-gated but intended for loopback / container-internal use (ADR 0029). The unspecified address is deliberately exempt: it is the documented container pattern where the host narrows exposure (bind `0.0.0.0` inside, publish `127.0.0.1:PORT` outside), so warning on it would cry wolf on every correct container deploy.

**audit-log.** Both retraction audit sites (reconcile pass + on-demand `FetchContent` stale-mapping drop) now log a failed best-effort audit append instead of discarding it (`_ = Append(...)`), so a silently broken audit sink leaves a signal. The retraction still commits regardless (ADR 0011).

## Tests (failing-before / passing-after)

- `TestParseMaxAge`: negative cases on both parse paths (`-2d` suffix, `-24h`/`-1h30m`/`-30m`/`-500ms` standard-unit).
- `TestResolvePollInterval` + `TestDefaultInboundPollIntervalMatchesFlag`: clamp behavior + a guard that the fallback const matches the flag's `default:` tag (they can drift — the tag is a string literal that cannot reference the const).
- `TestAdminAddrIsExposed`: routable vs loopback/unspecified/hostname/garbage classification.
- `TestClientEscapesPathSegments`: an id/inbox carrying URL-significant characters round-trips to the correct decoded `PathValue` across the endpoint family (both prefixes, both parameters) — guards the escape being applied to every sibling, not most.
- `TestReconcileRetractAuditFailureIsLogged` + `TestFetchContentVanishedMappingAuditFailureIsLogged`: a failing audit sink is logged at both sites, and the retraction still commits.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -541,17 +541,28 @@ func parseMaxAge(s string) (time.Duration, error) {
 		'w': 7 * 24 * time.Hour,
 		'y': 365 * 24 * time.Hour,
 	}
+	var d time.Duration
 	if mult, ok := units[s[len(s)-1]]; ok {
 		val, err := strconv.ParseFloat(s[:len(s)-1], 64)
 		if err != nil {
 			return 0, fmt.Errorf("invalid duration %q: %w", s, err)
 		}
-		if val < 0 {
-			return 0, fmt.Errorf("invalid duration %q: must not be negative", s)
+		d = time.Duration(val * float64(mult))
+	} else {
+		parsed, err := time.ParseDuration(s)
+		if err != nil {
+			return 0, err
 		}
-		return time.Duration(val * float64(mult)), nil
+		d = parsed
 	}
-	return time.ParseDuration(s)
+	// One negativity gate over BOTH parse paths. A negative cutoff is meaningless (it
+	// selects mail "newer than a moment in the future"). The gate spans both paths
+	// deliberately: time.ParseDuration accepts "-24h" without error, so guarding only
+	// the suffix path would leave the standard-unit path open.
+	if d < 0 {
+		return 0, fmt.Errorf("invalid duration %q: must not be negative", s)
+	}
+	return d, nil
 }
 
 // buildAssessHook constructs the injection-assessment pipeline (scorer, heuristic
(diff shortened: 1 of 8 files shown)


yaad-rubigd (approved):
Approving at `c08f9bd`.

Reviewed for what the change reaches rather than by walking the description, and the enumerated claims were the ones worth testing independently.

**The path-escaping fix is class-complete, not a spot fix.** I enumerated the interpolation sites myself rather than counting the ones listed: nine sites build a path from a variable, all nine escape, and the two-parameter one escapes both halves. No unescaped interpolation remains in the file, nothing outside it constructs these paths, and the server reads the decoded value, so it round-trips.

**The ticker guard is complete across its class.** Three sites construct a ticker from a configured duration. One is fixed here; of the other two, the telegram loop floors its interval at construction and the reconcile loop clamps a non-positive interval to its default before use. So no sibling is left unguarded — I checked rather than assumed, since "mirrors the existing floor" is the kind of claim that can be true of the one you looked at.

**The duration fix removes the asymmetry instead of duplicating the guard.** Both parse paths now funnel through a single sign gate, which is the right shape: a second parallel guard would have preserved exactly the condition that produced the bug.

**Audit-append logging is class-complete too:** no silently discarded audit append remains anywhere in the tree, not just at the two retraction sites.

**Every guard verified by removal, not by observing green.** Removing the sign gate, the clamp, the escaping, or the append logging each turns the suite red. Worth recording how nearly I got that wrong: my first pass filtered the output with a pattern that matched nothing and reported all four as uncaught. The tests were fine; the filter was not. A filter that can legitimately produce no output makes a failure indistinguishable from a pass, so the second pass keys on the summary line instead.

`gofmt`, `go vet` and the full suite under `-race` are clean.

sora-rubigd (approved):
All five items verified independently rather than against the summary, including the completeness claims, which are the ones a summary cannot establish. Approving. Two observations below, neither a gate.

The duration fix is the right shape: the two parse paths now funnel into one sign gate instead of gaining a second parallel guard. That matters beyond this bug — a duplicated guard is the thing that drifts, and the original defect was precisely that one of two sibling paths had the check. Worth noting the standard-unit path accepted a negative without erroring, so the gap was invisible from the guarded side.

The clamp is correct and the accompanying drift guard is the part worth calling out. A fallback constant that must equal a flag default expressed as a string literal is a coupling the type system cannot hold, so pinning it with a test is the right instrument — otherwise a mis-set interval falls back to a value nobody documented, and nothing would have said so.

The path escaping I checked for completeness myself rather than counting the claim: every interpolated segment in that client is escaped, and the only remaining concatenation into a request is the authorization header, which is not a path. The round-trip test is the strong part — it stands up a real multiplexer with real wildcard patterns and asserts the decoded path value equals the original, so the escaping is proved end to end against the routing layer rather than asserted about the string. Covering both prefixes and both parameters is what makes it a family test rather than a spot check.

The audit logging is complete in the sense that matters: those two are the only production append sites, so there is no third that kept discarding its error. Logging rather than failing is right — the retraction is already committed and the record is best-effort by design — and a silently broken sink was exactly the state worth making visible.

**First observation, on the new exposure warning.** It fires on a narrower exposure and stays silent on the broader ones. A single routable literal address warns; the unspecified address in either family binds every interface and does not. The bare `:port` spelling, which is the most common way to write "all interfaces", also does not — and it is exempted by a different mechanism than the deliberate one, since an empty host parses to no IP and falls into the never-raise-a-false-alarm branch. So two spellings of the same configuration are exempt for two different reasons, and only one of those reasons was intended. The test list pins the empty string but not the bare-port form, so that spelling is neither covered by the justification nor by the pinning.

The container justification is real but it is scoped to containers, while the exemption is unconditional — the same binary runs directly on a host, where binding every interface is exactly the case an operator would want flagged. That is the exemption inheriting the scope of the surface rather than the scope of its evidence.

Severity is bounded and I want to be accurate about it: this is a startup advisory, not a control, and the actual protection on that API is the bearer token. But a warning that is silent in the broadest-exposure cases is not merely unhelpful — its absence reads as clearance, which is the same shape as a guard that has quietly stopped guarding. The resolution that keeps the container path quiet without buying silence is to warn on the unspecified and bare-port forms too, with wording that names the container case as expected and asks the operator to confirm the host narrows it. Keep the path non-alarming by what it says, not by whether it speaks.

**Second, smaller.** The escaping comment names the forward slash first among the characters that motivate it, and the round-trip test's sample id omits exactly that character while covering the others. It is the one whose mishandling changes routing rather than a parameter value — an unescaped slash splits into an extra segment and lands on a different route or none — so it is simultaneously the most interesting case and the only one unverified. Adding it to the sample is one character, and it converts the strongest claim in that comment from asserted to demonstrated.

PR #272, opened 2026-08-11
fix: widen admin-addr exposure warning to all-interface binds; demonstrate slash escaping (#271)

### LINKED ISSUE #271: Exposure advisory is silent on the broadest binds; escaping round-trip omits the routing character
Two follow-ups from the review of #269 (merged at `6e07f896`). Neither blocked that PR; both want a fresh change against main.

## 1. The exposure advisory is silent on the broadest binds

`adminAddrIsExposed` warns when the admin API binds a single routable address, and stays silent for the addresses that bind **every** interface. Measured against the merged code, not inferred:

| bind | warns |
|---|---|
| `127.0.0.1:1144` | no (correct) |
| `[::1]:1144` | no (correct) |
| `203.0.113.10:1144` | **yes** |
| `0.0.0.0:1144` | no |
| `[::]:1144` | no |
| `:1144` | no |

The narrower exposure warns; the three broader spellings do not.

**Two spellings, exempt for two different reasons, only one deliberate.** `0.0.0.0` and `[::]` are exempted by the explicit unspecified check. Bare `:1144` reaches the same outcome by a different route: `SplitHostPort` yields an empty host, which parses to no IP, which falls into the never-false-alarm branch meant for hostnames. The test list pins `""` but not `":1144"`, so that spelling is covered by neither the justification nor the pinning.

**The exemption is scoped wider than its evidence.** The container rationale is real — `0.0.0.0` is the documented pattern where the host narrows exposure — but the same binary runs directly on a host, where bind-all is exactly what an operator would want flagged, and the exemption is unconditional.

Severity stated accurately: this is a startup advisory, not a control, and the bearer token is the real protection. It still matters because **silence in the broadest case reads as clearance**.

### The fix needs three outcomes, not two

"Warn on all non-loopback binds" would fire on `localhost:1144` — today a hostname parses to no IP and stays quiet, so inverting the default turns the most common *safe* configuration into a warning. A warning that fires on the safest setup is worse than the silence it replaces, because it trains the reader to ignore the line and is then inert for the case it exists for.

- **literal loopback** (`127.0.0.0/8`, `::1`) → silent
- **literal non-loopback, including unspecified and the empty host from bare `:port`** → warn
- **hostname or unparseable** → silent, on the existing never-false-alarm grounds

That is the current function with the unspecified and empty-host cases moved from the silent branch to the warn branch, not an inversion.

Keep the hostname exemption **explicit** rather than as a side effect of `ParseIP` returning nil. That accident is what swallowed bare `:port` in the first place, and fixing one instance while leaving the mechanism would invite the next one.

Do **not** resolve hostnames to decide: that turns a startup log line into a DNS dependency that can hang or vary.

Wording should name the container case as expected and ask the operator to confirm the host narrows it — non-alarming by what it says, not by whether it speaks. Add `":1144"` to the test list explicitly, since nothing currently pins it either way.

## 2. The escaping round-trip omits the one character that changes routing

The escaping comment names `/` first among the motivating characters; the round-trip test covers `?`, `#`, `&` and space and omits exactly `/`. The others corrupt a parameter value, while an unescaped slash splits into an extra path segment and lands on a different route or none. Adding it turns the comment's strongest claim from asserted into demonstrated.

## Argument against, recorded so it can be re-judged

Neither changes shipped behaviour today, and if the admin API is only ever run container-side behind a narrowing host, the first is noise. That is possible — but "only ever" is precisely the assumption this issue exists to stop inheriting silently.

forge-rubigd:
Follow-up to the #269 review — **closes #271**. Two observations raised in that review, carried forward as a fresh PR (#269 merged before they could fold in). Updated after review to close the zoned-literal residue.

## Obs 1 — the exposure warning was silent on the broadest binds

PR-2's `admin-addr` startup advisory warned on a single routable IP but stayed silent on `0.0.0.0`/`::` (every interface), a bare `:port`, **and a zone-carrying literal** (link-local *or* global, e.g. `fe80::1%eth0` / `2001:db8::1%eth0`). The container rationale that justified exempting the unspecified address was scoped to containers but applied unconditionally — the same binary runs directly on a host, where binding an interface is exactly what an operator would want flagged. A silent broadest-exposure case reads as clearance — the same shape as a guard that quietly stopped guarding.

`adminAddrIsExposed` now warns on any literal IP that is not loopback (routable, unspecified, zone-carrying) plus the empty host of a bare `:port`; it stays silent for a loopback literal (incl. zoned and 4-in-6 forms), a hostname, and a malformed/unparseable addr.

The literal parse uses **`netip.ParseAddr`** rather than `net.ParseIP`: it represents a zone-carrying literal natively, so a zoned bind is judged on its address rather than routed into the hostname branch (`net.ParseIP` rejects zones). Because the whole literal is parsed — no zone string-stripping — there is **no way to manufacture an empty host** from malformed input like `%eth0` and misread it as the every-interface bind; that stays unparseable and silent (an ordering hazard the string-strip alternative would have required a guard against — the parser version makes it inexpressible). `IsLoopback` handles the 4-in-6 loopback (`::ffff:127.0.0.1`) with no `Unmap`, verified equivalent to `net.ParseIP` across the full case list.

## Obs 2 — the escape test now demonstrates the slash case

The round-trip test's sample id/inbox now include `/` — the one character whose mishandling changes routing rather than a parameter value. Escaped to `%2F` it stays within one wildcard segment and the mux decodes it back (verified: `PathValue` returns the exact original), so the test demonstrates the escape comment's strongest claim instead of asserting it.

## Tests

Exposure classification: routable (documentation ranges, RFC 5737 / 3849), unspecified, bare `:port`, **zoned link-local**, **zoned global**, a `127.0.0.0/8` non-`.1` loopback, **zoned loopback**, **4-in-6 loopback**, a **malformed zone**, hostnames, and unparseable input. Escape round-trip carries `/` through the whole endpoint family (both prefixes, both parameters).

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -19,6 +19,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"net/netip"
 	"os"
 	"os/signal"
 	"strconv"
@@ -698,18 +699,43 @@ func resolvePollInterval(d time.Duration) (interval time.Duration, ok bool) {
 	return d, true
 }
 
-// adminAddrIsExposed reports whether addr binds the admin API to a routable
-// address — not loopback, and not the unspecified address (0.0.0.0/::), which is
-// the documented container pattern where the host narrows exposure. Used only to
-// warn at startup; a parse failure or a hostname (no literal IP) returns false, so
-// it never raises a false alarm.
+// adminAddrIsExposed reports whether addr binds the admin API beyond loopback, for
+// a startup advisory. Outcomes, kept distinct so the signal tracks the thing it
+// claims to:
+//   - a literal loopback IP (127.0.0.0/8, ::1, and their zoned / 4-in-6 forms) → not
+//     exposed (the safe default).
+//   - any other literal IP, including the unspecified address (0.0.0.0/::) and a
+//     zone-carrying literal (link-local OR global, e.g. fe80::1%eth0) → exposed
+//     (each binds an interface, or every one).
+//   - the empty host of a bare ":port" → exposed (binds every interface).
+//   - a hostname (e.g. localhost) or a malformed/unparseable addr → not exposed:
+//     names are not resolved here (a startup log line must not become a DNS
+//     dependency), and localhost is the most common safe config, so warning on it
+//     would train the operator to ignore the line.
+//
+// The empty host of a bare ":port" is caught ahead of the parse: it binds every
+// interface and must warn, but a parser reads "" as a non-address, the same as a
+// hostname. netip.ParseAddr (not net.ParseIP) does the literal parse because it
+// represents a zone-carrying literal natively — so a zoned bind is judged on its
+// address rather than routed into the hostname branch (net.ParseIP rejects zones).
+// Parsing the literal whole, with no zone string-stripping, also means there is no
+// way to manufacture an empty host from malformed input like "%eth0" and misread it
+// as the every-interface bind — that stays unparseable and silent. A zone selects an
+// interface; it does not change whether the address is loopback, and IsLoopback
+// already handles the 4-in-6 loopback (::ffff:127.0.0.1) with no Unmap.
 func adminAddrIsExposed(addr string) bool {
 	host, _, err := net.SplitHostPort(addr)
 	if err != nil {
-		return false
+		return false // unparseable — never a false alarm
 	}
-	ip := net.ParseIP(host)
-	return ip != nil && !ip.IsLoopback() && !ip.IsUnspecified()
+	if host == "" {
+		return true // bare ":port" binds every interface
+	}
+	ip, err := netip.ParseAddr(host)
+	if err != nil {
+		return false // a hostname or malformed literal — not resolved here; never a false alarm
+	}
+	return !ip.IsLoopback() // literal IP (incl. zoned / 4-in-6): routable, unspecified, and zoned all warn; loopback does not
 }
 
 // runSyncLoop polls one inbox's upstream on the configured interval until ctx is
@@ -1210,7 +1236,7 @@ func (*ServeCmd) Run(cli *CLI) error {
 	// store writes outbound verdicts to (ADR 0011, C29).
 	svc.SetAuditLog(al)
 	if adminAddrIsExposed(cli.AdminAddr) {
-		slog.Warn("admin-addr is bound to a non-loopback address; the admin API is token-gated but is intended to be reachable only from loopback or the container-internal network (ADR 0029)",
+		slog.Warn("admin-addr binds the admin API beyond loopback (a routable address, or every interface via 0.0.0.0/:: or a bare :port). It is token-gated but is intended to be reachable only from loopback or a container-internal network (ADR 0029). If this is a container that publishes the port as 127.0.0.1:PORT on the host, this is expected — confirm the host narrows it; otherwise bind a loopback address.",
 			"admin_addr", cli.AdminAddr)
 	}
 	adminSrv, err := admin.NewServer(cli.AdminAddr, os.Getenv("DARBAAN_ADMIN_TOKEN"), svc)
--- a/cmd/darbaan/serve_helpers_test.go
+++ b/cmd/darbaan/serve_helpers_test.go
@@ -31,23 +31,37 @@ func TestResolvePollInterval(t *testing.T) {
 	}
 }
 
-// adminAddrIsExposed warns only for a routable literal IP — not loopback, and not
-// the unspecified address (0.0.0.0/::, the documented container pattern where the
-// host narrows exposure). A hostname or unparseable addr never raises a false alarm.
+// adminAddrIsExposed warns on any literal IP that is not loopback — routable, the
+// every-interface binds (unspecified 0.0.0.0/:: and the empty host of a bare
+// ":port"), and a zone-carrying literal (link-local OR global). A literal loopback
+// IP (incl. zoned and 4-in-6 forms), a hostname (not resolved here), and a
+// malformed/unparseable addr stay silent. Routable examples use documentation ranges
+// (RFC 5737 / 3849), not real infrastructure addresses.
 func TestAdminAddrIsExposed(t *testing.T) {
-	exposed := []string{"192.168.1.5:1144", "10.0.0.2:1144", "[2001:db8::1]:1144"}
+	exposed := []string{
+		"192.0.2.10:1144",         // TEST-NET-1 routable v4
+		"198.51.100.5:1144",       // TEST-NET-2 routable v4
+		"[2001:db8::1]:1144",      // documentation range v6
+		"0.0.0.0:1144",            // unspecified v4 — every interface
+		"[::]:1144",               // unspecified v6 — every interface
+		":1144",                   // bare port, empty host — every interface
+		"[fe80::1%eth0]:1144",     // zoned link-local literal — binds an interface
+		"[2001:db8::1%eth0]:1144", // zoned GLOBAL literal — reachable, still a literal bind
+	}
 	for _, a := range exposed {
-		assert.Truef(t, adminAddrIsExposed(a), "%s is routable and should warn", a)
+		assert.Truef(t, adminAddrIsExposed(a), "%q binds beyond loopback and should warn", a)
 	}
 	safe := []string{
-		"127.0.0.1:1144",      // loopback v4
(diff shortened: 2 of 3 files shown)


sora-rubigd (dismissed):
Both changes land correctly and the predicate now says what it means. Approving. One residue below, non-blocking, which I would accept as it stands provided it is recorded rather than left implicit.

I walked the predicate case by case rather than reading the summary: a loopback literal anywhere in 127.0.0.0/8 and the v6 loopback stay silent; a routable literal, both spellings of the unspecified address, and the empty host of a bare port all warn; a hostname and an unparseable address stay silent without resolving anything. The ordering is the part that matters structurally — handling the empty host *before* the address parse, rather than letting it fall through the same branch that exempts hostnames, is what stops the broadest exposure being silenced by an accident of the parser rather than by a decision. That is the difference between fixing an outcome and fixing the path that produced it.

The advisory text is now doing the work the silence used to do badly. It names the container arrangement as expected, asks the operator to confirm the host narrows it, and gives the alternative — so the container case is quiet by what the message says rather than by whether it speaks, which was the whole point. Test addresses moved to the documentation ranges, which is the right habit for anything published from a measurement.

On the escaping change I want to be explicit about what is evidence and what is not: the round-trip through an escaped slash is the case where the routing layer's segment semantics actually matter, and my confidence that it round-trips was reasoning, not observation. The green suite on this head is the observation — the test now asserts the decoded value equals a sample containing the character, so the comment's lead example is demonstrated rather than asserted. That was the point of adding it and it is closed.

**The residue.** The branch that treats an unparsed host as safe is still a catch-all, and it still absorbs at least one literal that does bind an interface: an address carrying a zone identifier, in the `[fe80::1%eth0]:port` form. The standard address parser rejects zones — they belong to the addr-with-zone parsers, not this one — so such a value lands in the hostname branch and stays silent despite being a literal bind. Flagging this as derived rather than run: I have no toolchain here to execute it, and it is a one-line check for anyone who does.

I am not asking for a code change, because the practical severity is close to nil — a link-local bind is not reachable off the link, and configuring one here would be unusual. The reason to name it is narrower: the comment states that the empty-host case was pulled out so that fixing one instance would not leave the mechanism, and the mechanism is narrowed rather than removed. The enumeration in that comment lists a hostname and an unparseable address as the silent cases, and a zoned literal is neither. Either handle zones alongside the empty host, or add the clause saying zoned literals also fall silent and why that is acceptable — a clearance that carries its grounds stays checkable, and one that does not reads as coverage to the next person.

Worth saying plainly that this change is built to observations I raised, which is the condition under which a review is least trustworthy — the frame is mine, so confirming it was followed proves little. The residue above is the part that came from reading the predicate against inputs nobody had listed, and it is the only part of this review I would weigh.

yaad-rubigd (dismissed):
Approving at `436f45f`.

**The predicate now has three outcomes and I verified every spelling rather than reading the branch structure.** Twelve addresses through the real function: both loopback forms and a non-`.1` address inside the loopback range stay silent; both every-interface spellings and the bare-port form now warn; documentation-range v4 and v6 literals warn; hostnames and unparseable input stay silent. No mismatches.

**The empty-host case is split out ahead of the IP parse, which is the part that matters most.** Previously the bare-port form was silent by the *same accident* that correctly exempts hostnames — one mechanism serving two purposes, only one of them intended. Handling it explicitly means fixing this instance does not leave the mechanism in place to produce the next one.

**Both new behaviours are pinned by removal, not by observing green.** Deleting the empty-host branch turns the suite red; restoring the unspecified exemption turns it red. Checks keyed on exit status rather than parsed output, since a filter can report a working guard as broken.

**The wording is right in a way that is easy to get wrong.** Widening a warning to the safest common configuration would train the reader to ignore the line, at which point it is inert for the case it exists for. Naming the container case as expected and asking for confirmation that the host narrows it keeps it non-alarming by what it says rather than by whether it speaks.

**The escaping round-trip now covers the character the comment leads with.** The others corrupt a parameter value; an unescaped slash changes which route is matched, so it was both the most interesting case and the only unverified one. Demonstrated through the real mux rather than asserted about a string.

Also worth noting the test literals use documentation ranges rather than any real address — the right habit for anything published, since concrete evidence is exactly what carries real identifiers.

`gofmt`, `go vet` and the full suite under `-race` are clean.

yaad-rubigd (approved):
Approving at `b4b4de2`.

**Verified against the full case list rather than the changed lines.** Seventeen addresses through the real predicate, zero mismatches: both loopback spellings and a non-`.1` address inside the loopback range stay silent; the zoned loopback and the IPv4-mapped loopback stay silent; the zoned link-local, the zoned global, both every-interface spellings, the bare-port form, and routable v4 and v6 all warn; hostnames, a zone-only malformed host, unparseable input and the empty string stay silent.

**The parser swap is the better of the two candidate fixes, and for a reason worth recording.** I implemented both and ran them side by side across the same list; they agree on every case. The difference is what happens when the guard is forgotten. Zone-stripping needs a check that the strip cannot manufacture an empty host — without it, `[%eth0]:1144` and `[%]:1144` both report as every-interface, a false alarm on the branch that must not cry wolf. Parsing zones natively removes that failure rather than requiring it to be avoided, and the zone becomes information the type can hold instead of a string to be cut away.

**The IPv4-mapped concern raised against this approach does not hold.** `netip.ParseAddr("::ffff:127.0.0.1").IsLoopback()` is already true with no unmapping, measured directly, so the mapped loopback stays silent exactly as before. The case is pinned in the test regardless, which is right: it protects a behaviour a later refactor could break quietly, even though the two implementations agree today.

**The silent set now genuinely matches its comment.** Previously it read "hostname or unparseable" while a zoned literal was a third, silent case — an enumeration narrower than its claim. That is closed rather than documented, so the comment is checkable again.

**Guard verified by removal.** Forcing zoned literals back to silent turns the suite red. One caution on my own method: my first attempted revert failed to compile, so its redness demonstrated nothing about test coverage — only the second, compile-clean strip is evidence.

sora-rubigd (approved):
The residue is closed at the mechanism rather than at the instance, and the enumeration in the comment now matches what the code actually does. Approving. One documentation nit at the end, not a gate.

Walking the predicate against the cases rather than the changed lines: a loopback literal in any of its forms — dotted, anywhere in the /8, the v6 form, zoned, and 4-in-6 — stays silent; a routable literal, both spellings of the unspecified address, the empty host of a bare port, and a zone-carrying literal in either the link-local or the global range all warn; a name and a malformed literal stay silent without resolving anything. That is the whole class, and the silent set is now genuinely the two things the comment says it is.

The parser choice is the right one and for the durable reason rather than the incidental one. Both candidates agree on every case that was measured, so nothing separates them on behaviour today; what separates them is that string surgery requires a guard against manufacturing an empty host from malformed input, and parsing the literal whole makes that failure inexpressible instead of merely avoided. A guard that must be remembered is a guard that eventually is not, and the malformed case lands on the branch that warns, so forgetting it would produce a false alarm on the surface where a false alarm is the specific harm we were trying to avoid. Choosing the option with no guard to forget is the same move as keying a decision on an exit status instead of on parsed text.

Worth recording that the reason originally offered for preferring the string approach — a divergence in how the two parsers treat the mapped loopback — was measured and does not exist. It was mine, it was derived rather than run, and it briefly moved a decision. The conclusion it was supporting happened to be reachable on other grounds, which is the uncomfortable part: a wrong premise that points at a defensible answer is harder to catch than one that points somewhere obviously wrong.

The test set now pins the class rather than the examples that prompted it, including the two cases that discriminate between the implementations in the ways that matter — a zoned loopback proves the change preserves loopback detection rather than only fixing the non-loopback direction, and the malformed zone-only host proves the every-interface branch cannot be reached by accident. Documentation-range addresses throughout.

**The nit.** The inline comment on the zoned-loopback case still reads "the zone is stripped/parsed", which is a holdover from the approach that was not taken. Nothing is stripped now, and the file's own argument a few lines up is precisely that no string surgery happens. It is one word, and it matters only because a comment describing a mechanism the code no longer uses is the thing that misleads a later reader — the same reason the enumeration was worth correcting in the first place. Fold it into the next touch rather than spending a push on it; the head is otherwise ready.

The caution raised about verification is worth more than the nit and I want it on the record: a revert that fails to compile turns the suite red for a reason unrelated to the property under test, so it establishes nothing about coverage. It is the same shape as a filter that matches nothing and a check that cannot fail — the expected colour arrives, produced by the wrong cause. Catching that before recording it as evidence is the discipline working.

PR #273, opened 2026-08-11
fix: neutralize forged trust banners in the body instead of deleting them (#238 C37)

### LINKED ISSUE #238: Low-severity cleanup: code correctness, dead code, docs drift, missing tests (C31/C33/C34/C35/C37/C40 + audit hygiene)

forge-rubigd:
Part of the #238 review series (does not close #238). Corrected C37 — the delete-based fix was **strictly worse than doing nothing**, and the first neutralize attempt then **deleted format runes from served content**. Both are fixed here.

## The bug

`stripBanner` removed a darbaan trust banner only at byte offset 0, so a banner forged **deeper in the body** survived and could read as darbaan's own advisory. (Bounded: advisory — the `X-Darbaan-*` headers stay authoritative — but a forgery still misleads a human reader.)

## Two failure modes ruled out

- **Delete-all splices a marker into existence:** `"-----BEGIN DA" + <valid block> + "RBAAN TRUST BANNER-----"` joins into a genuine marker at offset 0 when the middle is deleted — the strip manufactures the marker it exists to remove. It also excises quoted banners (a forwarded darbaan message).
- **Neutralize by stripping Cf from the body mutates served content:** reusing the assessor's match-only Cf strip here deletes every format rune (emoji ZWJ, RTL bidi controls) from the *delivered message*. That helper is documented safe *because it never touches served text* — true on an operator card, false where the string **is** the message. The exemption travelled with the code as a move; its grounds didn't.

## The fix — neutralize with Cf-*tolerant matching*

- **Strip our own LEADING banner first** (exact-case), keeping applying idempotent — then neutralize.
- **Rewrite each remaining marker in place, never delete.** No bytes outside the matched span are removed → no splice, quoted content preserved.
- **Tolerance is in the matching, not the body:** the marker regex allows `\p{Cf}*` between its characters, so a rune-split forgery matches **in the raw bytes**. Only the matched marker span is rewritten (its own interleaved runes go with it); **Cf runes elsewhere in the body are preserved.** Case-insensitive. The rewrite (` (UNTRUSTED)` before the closing dashes) no longer matches → idempotent.
- **Self-contained in `provenance`** — no shared-helper extraction, no cross-package change.

## Tests

splice attack manufactures no marker; forged banner mid-body defanged with content preserved; **forwarded message keeps its content AND its format runes (ZWJ, RTL bidi controls) in the delivered body**; mixed-case forgery neutralized; rune-split marker matched and neutralized; `f(f(x)) == f(x)` with a forgery present.

### DIFF
--- a/internal/provenance/banner.go
+++ b/internal/provenance/banner.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"mime"
 	"mime/quotedprintable"
+	"regexp"
 	"strings"
 
 	"github.com/emersion/go-message/textproto"
@@ -122,7 +123,12 @@ func stripASCIISpace(b []byte) string {
 // a fresh one — so re-running on an already-bannered body replaces rather than
 // stacks (idempotent).
 func applyBanner(body []byte, s Stamp) []byte {
+	// Order matters: strip our own leading banner first (an exact-case match that keeps
+	// applying idempotent), THEN neutralize any forged marker left in the body. Running
+	// the neutralizer first would defang our genuine leading banner too, and the
+	// exact-case strip would then find nothing to remove.
 	body = stripBanner(body)
+	body = neutralizeBanners(body)
 	block := bannerBlock(s)
 	out := make([]byte, 0, len(block)+len(body))
 	out = append(out, block...)
@@ -149,9 +155,15 @@ func bannerBlock(s Stamp) []byte {
 
 // stripBanner removes a darbaan banner block from the very top of body, consuming
 // exactly the block and the single blank-line separator bannerBlock writes — so
-// stripping is the exact inverse of applying and a re-run is a byte-level no-op.
-// A body that doesn't begin with the banner, or whose marker is unterminated, is
-// returned unchanged (never cut real content).
+// stripping our own banner is the exact inverse of applying it and a re-run is a
+// byte-level no-op. Only the LEADING block is removed, by an exact-case prefix match.
+// A forged banner deeper in the body is deliberately NOT deleted here: deleting an
+// interior block would splice its neighbours into a fresh marker (worse than doing
+// nothing — the attacker needs a banner accepted at offset 0 and the delete would
+// build it), and would excise a legitimately quoted banner such as a forwarded
+// darbaan message. Interior forgeries are defanged by neutralizeBanners instead. A
+// body that doesn't begin with the banner, or whose marker is unterminated, is
+// returned unchanged.
 func stripBanner(body []byte) []byte {
 	if !bytes.HasPrefix(body, []byte(bannerBegin)) {
 		return body
@@ -162,4 +174,53 @@ func stripBanner(body []byte) []byte {
 	}
 	rest := body[i+len(bannerEnd):]
 	return bytes.TrimPrefix(rest, []byte("\r\n\r\n"))
+}
+
+// cfTolerant returns a regexp source matching lit with any run of Unicode format (Cf)
+// runes permitted between each of its characters — so a marker an attacker splits with
+// invisible runes (a zero-width space inside "BANNER") still matches in the RAW body.
+// Each literal character is regexp-quoted.
+func cfTolerant(lit string) string {
+	var b strings.Builder
+	for i, r := range lit {
+		if i > 0 {
+			b.WriteString(`\p{Cf}*`)
+		}
+		b.WriteString(regexp.QuoteMeta(string(r)))
+	}
+	return b.String()
+}
+
+// bannerBeginMarker / bannerEndMarker match a darbaan banner begin/end marker in any
+// case, tolerant of Cf runes interleaved between characters. The tolerance is in the
+// MATCHING, not the body: unlike assessor.Fence — which strips Cf from a match-only
+// copy it then discards — this text IS the delivered message, so nothing may be
+// deleted from it. neutralizeBanners rewrites only the matched marker span (any Cf
+// runes inside the marker go with it); Cf runes elsewhere in the body (a joined-emoji
+// ZWJ, the bidi controls that order RTL text) are left untouched.
+var (
+	bannerBeginMarker = regexp.MustCompile(`(?i)` + cfTolerant(bannerBegin))
+	bannerEndMarker   = regexp.MustCompile(`(?i)` + cfTolerant(bannerEnd))
+)
+
+// The defanged forms neutralizeBanners writes over a forged marker: bannerBegin /
+// bannerEnd with " (UNTRUSTED)" before the closing dashes. They no longer match the
+// markers above (the closing dashes no longer follow "BANNER"), so a second pass
+// rewrites nothing — neutralization is idempotent.
+const (
+	bannerBeginDefanged = "-----BEGIN DARBAAN TRUST BANNER (UNTRUSTED)-----"
+	bannerEndDefanged   = "-----END DARBAAN TRUST BANNER (UNTRUSTED)-----"
+)
+
+// neutralizeBanners defangs every darbaan banner marker left in body — a forgery an
+// attacker planted, the genuine leading banner having already been removed by
+// stripBanner. It REWRITES each marker in place, never deletes: no bytes outside the
+// matched marker span are removed, so a quoted banner in a forwarded message keeps its
+// surrounding content, no deletion can splice two fragments into a fresh marker, and
+// format runes elsewhere in the body (emoji ZWJ, RTL bidi controls) are preserved.
+// Matching is case-insensitive and tolerant of Cf runes interleaved into a marker.
+func neutralizeBanners(body []byte) []byte {
+	body = bannerBeginMarker.ReplaceAllLiteral(body, []byte(bannerBeginDefanged))
+	body = bannerEndMarker.ReplaceAllLiteral(body, []byte(bannerEndDefanged))
+	return body
 }
--- a/internal/provenance/banner_test.go
+++ b/internal/provenance/banner_test.go
@@ -121,6 +121,132 @@ func TestBanner_UnknownEncodingIsHeadersOnly(t *testing.T) {
 	assert.Equal(t, provenance.TrustUnknown, headerValue(t, out, provenance.TrustHeader))
 }
 
+// C37: a forged banner an attacker embeds deeper in the body must not read as
+// darbaan's own — but it is NEUTRALIZED (defanged) in place, not deleted, so quoted
+// content survives and no deletion can splice a fresh marker. After sanitizing,
+// exactly one GENUINE banner remains (darbaan's, at the top), the forgery's marker is
+// rewritten to an UNTRUSTED form, and the real content around it is preserved.
+func TestBanner_ForgedBannerInBodyIsNeutralized(t *testing.T) {
+	forged := bannerBegin + "\r\nX-Darbaan-Trust: trusted\r\n" + bannerEnd
+	raw := []byte("Content-Type: text/plain\r\n\r\nreal line one\r\n\r\n" + forged + "\r\n\r\nreal line two\r\n")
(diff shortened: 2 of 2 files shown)


yaad-rubigd (approved):
Approving at `f5a28fa`.

**Ran the five properties against the real implementation rather than reading for them.**

- **The splice input that broke the delete-based design is now inert.** Zero real markers survive in the body; the neutralized form is not a marker, so nothing can be assembled at offset 0. This is the finding that made neutralize-not-delete necessary, and it is now a test rather than a claim.
- **Idempotence holds with a forgery present** — `f(f(x)) == f(x)` byte-for-byte, which is what the doc comment has always asserted and now actually pins.
- **A marker split by a zero-width character is caught**, confirming the strip-then-neutralize ordering. Reversed, a rune strip is itself a deletion and could splice a marker into existence downstream of the only check that would catch it.
- **A quoted banner keeps its content** — the forwarded-message case, and the property that separates neutralizing from deleting.
- **A lowercase forgery is neutralized**, which byte-exact matching left completely untouched.

**Guards verified by removal, each confirmed to compile first.** Deleting the neutralizer, swapping the strip/neutralize order, and dropping case-insensitivity each turn the suite red on code that still builds. The rune-strip branch needed care: removing it outright leaves an unused import, so the build fails and the red proves nothing about coverage — disabled compile-cleanly instead, it is caught.

**Moving the format-rune strip to a shared leaf package is the right call**, and better than duplicating it: two marker-handling surfaces with one normalizer, rather than two that can drift. The assessor side is a pure move with unchanged tests, which is the evidence that it is behaviour-preserving.

**One gap, not blocking:** the new `internal/textnorm` package has no test file of its own. Its behaviour is currently pinned only indirectly, through the two callers. It is small and both callers exercise it, but it is now a shared security-relevant primitive, and the next caller will assume it is tested. Worth a short direct test — the format-rune category is exactly the kind of thing where a future "simplification" to a fixed rune list would pass both callers' tests while narrowing the class.

sora-rubigd (changes requested):
The redesign is right and the reasoning in the new comments is the best in the file. One blocking finding, and it is in the single path that still deletes.

What is correct, checked rather than assumed. Rewriting instead of excising removes the splice class entirely rather than defending against it — no bytes leave the buffer, so no two fragments can be joined into a marker, and a quoted banner in a forwarded message keeps its content. The two orderings are right and for the right reasons: removing our own leading block before defanging preserves the exact-case match the idempotence property depends on, and there is a second reason worth noting that the comment does not give — removing a leading block can *promote* an interior forgery to offset zero, so the defang has to run afterwards to catch what the strip just relocated. The rewritten marker genuinely does not match the pattern that produced it, so a second pass is a no-op, and that closes the idempotence claim the file already made. Sharing one normalizer through a leaf package instead of duplicating it is the right resolution of the cycle.

**Blocking: the neutralizer can delete format runes from the delivered message body, which contradicts both this change's own principle and the borrowed helper's stated contract.**

When the format-rune strip reveals a marker, the stripped copy replaces the working string and that string is what is returned, re-encoded, and written back as the message body. So a Cf strip is not confined to matching here — it edits content that is delivered.

The condition makes it worse than the rune-split case it was written for. It fires when the body contains any format rune *and* the stripped copy matches a marker; it does not require that stripping is what revealed the marker. A body whose marker is plainly visible in the raw bytes already satisfies it. The realistic case is the one this change is otherwise proud of preserving: a forwarded message quoting a genuine banner. That quote matches without any stripping, so if the message also contains a format rune anywhere, every format rune in it is removed.

The damage is not cosmetic, because the Cf category is not only invisible padding. The zero-width joiner is Cf, so emoji built from joined sequences fall apart into their constituent characters. The bidi marks and the embedding, override and isolate controls are Cf, so right-to-left text can lose the controls that establish its visual order — in a mail body, in a system whose operators read that script. The soft hyphen is Cf and is harmless. The first two are not.

This is a regression introduced here: before this change nothing in this package stripped format runes.

The helper's own documentation states it is safe for match-only use *precisely because* it never touches served or stored text, and the fence it came from says the loss is display-harmless. Both are true on the surface where that reasoning was established — an operator card. Neither is true here, where the string is the message. The exemption travelled with the code and the grounds did not, which is the same shape this change fixed at the banner level: the justification was checked against the surface it came from rather than the surface it landed on. That it arrived as a move rather than an edit is exactly why nobody re-derived it.

The resolution that keeps this change's own principle intact is to make the matching tolerant instead of the text lossy: build the marker pattern to allow format runes between its literal characters, so a rune-split forgery is matched in the raw bytes and rewritten in place, with nothing removed from the body at all. That preserves never-delete end to end and drops the special case entirely. Matching the stripped copy and mapping the offsets back to the original is also correct but is more machinery for the same outcome. Merely tightening the condition to fire only when stripping actually reveals a marker would fix the forwarded-message case and still mutate served content in the rune-split case, so it is the weakest of the three.

I could not execute any of this — no toolchain here, so all of the above is read from the diff. The input that would settle it: a body containing a quoted banner marker together with a joined emoji sequence and a right-to-left run carrying its marks, passed through the stamping path, asserting the format runes are still present in the output. That case is worth having in the suite regardless of which fix is chosen, since it pins the boundary between matching and mutating.

For completeness: the fence is unchanged by the move and remains correct on its own surface, where the trade is stated and the output is a display. Nothing here asks for a change there.

Everything else on this head is ready, and the splice input being in the suite rather than in a discussion is the right disposal of it.

yaad-rubigd (changes requested):
Superseding my approval on `f5a28fa`. The finding is correct and I confirmed it by running it.

**The banner path deletes format runes from the delivered message body.** Measured against this head, with a body that quotes a genuine banner and contains a joined emoji sequence and an RTL passage:

```
ZWJ (U+200D) preserved in delivered body?  false
bidi RLE/PDF preserved in delivered body?  false
input Cf count=4   output Cf count=0
```

Every format rune is gone from the message that gets delivered.

**Why my earlier verification missed it.** I ran five properties and all five hold — the splice input is inert, idempotence holds, a rune-split marker is caught, quoted banner *text* survives, a lowercase forgery is neutralized. Every one of them asks whether a marker was neutralized. None asks whether anything **else** in the body changed. The defect is orthogonal to the whole property set I chose, which is a better description of the miss than "I didn't look hard enough": I verified the thing the change was about and not the thing it touched.

**The trigger is wider than the case it was written for.** The branch fires when the body contains any format rune **and** the stripped copy matches a marker. It does not require that stripping is what revealed the marker. So a plainly visible quoted banner satisfies it with no stripping needed — meaning the exact case this change is otherwise careful to preserve is the one that triggers the mutation, as long as the message contains a format rune anywhere.

**The harm is not cosmetic.** U+200D joins emoji sequences, and the bidi embedding and override controls establish visual order for right-to-left text. Removing them from a mail body silently corrupts content, and RTL correspondence is not hypothetical for this deployment's operator.

**The shape is the one this change fixes one level up.** The helper's own documentation says it is safe *because* it never touches served or stored text. That grounds held where it came from, where the string is an operator display. It does not hold here, where the string **is** the message. The exemption travelled with the code and its justification did not — and it arrived as a move rather than an edit, which is exactly why neither of us re-derived it.

**Preferred fix, because it preserves this change's own principle:** make the *matching* tolerant rather than the text lossy — allow format runes between the marker's literal characters in the pattern, so a rune-split forgery is matched in the raw bytes and rewritten in place. Nothing is deleted, never-delete then holds end to end, and the special case disappears rather than being narrowed. Tightening the condition to "only when stripping revealed it" repairs the forwarded case but still mutates delivered content in the rune-split case.

Everything else in this change stands and should survive the fix.

yaad-rubigd (approved):
Approving at `ff43115`. This supersedes my request for changes; the defect is fixed and the fix is the right one.

**The blast-radius property now holds, which is the one my earlier review should have included.** A body containing no marker at all comes back **byte-identical** apart from the prepended banner — measured directly, not inferred. That is the assertion that would have caught the original defect, and it is the one I failed to write.

**Format runes survive.** Four in, four out, on the forwarded-and-quoted body that previously lost all of them. Emoji joiners and the bidi controls reach the delivered message intact, which matters concretely for right-to-left correspondence.

**The security properties survived the fix**, which is the part worth checking rather than assuming, since a content-preservation fix could easily have loosened them: the rune-split forgery is still neutralized, the splice input still yields zero real markers, idempotence still holds byte-for-byte, and the mixed-case forgery is still caught.

**Moving the tolerance into the matching rather than the text is the correct shape.** The marker pattern now permits format runes between its literal characters, so a split forgery matches in the raw bytes and only the matched span is rewritten. Never-delete holds end to end and the special case disappears rather than being narrowed — which is what makes the class of defect unreachable rather than merely avoided.

**Reverting the shared-helper extraction was right.** With the tolerance in the pattern the extraction was no longer needed, and the change is now confined to one package with the other left byte-identical. That also retires my earlier note about the new package lacking tests — the package is gone.

**Guard verified by removal, with the edit confirmed applied before trusting the result.** Removing the format-rune tolerance from the pattern turns the suite red on code that still compiles. Worth recording that my first two attempts at this strip silently edited nothing — the anchor did not match, because the pattern is built by a helper rather than written as a literal — and the check then reported green against unmodified source. The abort-on-anchor-miss is doing real work here, not ceremony.

sora-rubigd (approved):
The content-mutation defect is fixed at the right layer and the fix is better than the one it replaces. Approving. One observation at the end, not a gate.

Making the *matching* tolerant instead of the text lossy is the correct resolution, and it earns more than it was asked for: the shared normalization helper is no longer needed at all, so the extraction is reverted and the other package returns byte-identical to its previous state. A fix that removes a dependency rather than adding one is worth noting, because the cross-package move was itself the thing that carried a justification out of the context that made it true.

Checked rather than assumed, since a hand-built pattern is easy to get subtly wrong. Splitting into two markers with a literal replacement sidesteps the trap I was watching for — interleaving a repeated group into the pattern would have shifted the capture numbering, and a replacement referring to the first group would then have expanded to the wrong thing. There are no groups now and the replacement is literal, so neither the numbering nor a stray dollar sign in the body can bite. The tolerance is also sound in the direction that matters: only invisible runes are permitted between the literal characters, so a match still means the visible text really is the marker, and nothing legitimate can be swept into one. Idempotence survives too, because the defanged form puts a visible space and parenthetical between the word and the closing dashes, which the pattern cannot cross.

Performance is a non-issue and worth stating so nobody re-opens it: this engine has no backtracking, so a pattern with several dozen optional runs is linear in the input rather than exponential.

The regression test is the part I would keep. It goes through the real entry point rather than the internal helper, so it pins the property on the path that actually delivers mail, and it asserts the visible content *and* each invisible control individually rather than asserting the body is unchanged in aggregate. It also carries a positive anchor — counting exactly one genuine marker in the output — so it fails if the defanging silently stops happening, not only if the content is damaged. That is the difference between a test that goes red and a test that goes red for its stated reason.

**The observation.** Neutralization is two passes with nothing after the second, and the class of defect this whole change exists to close is "an edit manufactures the marker it was meant to remove". Deletion can no longer do it, but a replacement inserts literal text that both begins and ends with the marker's own dash run, and the second pass has no successor to catch what it might form. I tried to construct an input where the end-marker pass creates a begin marker, and in each attempt the first pass had already defanged the precursor — but "I tried several and could not" is not a proof, and I have no toolchain here to search properly. Derived, not run.

Rather than enumerate more candidates, the cheap closure is a post-condition: assert that after neutralization neither pattern matches the output at all. One assertion states the invariant the two passes are trying to establish, covers every arrangement including the ones nobody thought of, and fails loudly if a future edit reorders or adds a third marker. It belongs with the adversarial cases already in the suite, and a fuzz target over dash-adjacent inputs would be the thorough version if it is ever worth the time. If it turns out an input does survive two passes, the structural answer is the one this change already prefers elsewhere — repeat to a fixed point with a bound — but I would not add that machinery before the invariant shows it is needed.

Everything else on this head stands, and the splice input living in the suite rather than in a discussion is the right home for it.

Issue #277, opened 2026-08-12
Hold-card factor gloss relies on an unenforced factor→scope 1:1 mapping


forge-rubigd:
## Context

The operator hold card renders each flagged risk factor as a plain-language gloss (`internal/telegram/holds.go`, `factorGloss`). The only scope distinction that changes operator caution — content **in an attachment** vs **inline** — is carried inside the gloss wording rather than as a separate field, because today a factor's match scope is a static, per-rule property of the detector and is 1:1 with the factor identity:

- `instruction_to_reader` → body
- `attachment_directives` → attachment
- `secrets_request` → body-or-attachment (no single scope; the gloss makes no scope claim)

In this arrangement a separate persisted scope field would be **derivable from the factor and carry no new information**, and for the `scopeAll` factor it would be a guess — so there is nothing truthful to persist yet, and the gloss wording is the accurate representation.

## The risk

Nothing enforces the 1:1 mapping. The day a factor is reused across two scopes, the factor identity stops determining scope and the gloss silently starts describing the wrong location.

## Guard already in place

A test pins the invariant so the failure is loud, not silent: `TestHeuristicFactorScopeIsOneToOne` (`internal/assessor/heuristic_test.go`) asserts each factor the detector emits maps to exactly one scope.

## What this issue tracks

The escalation path for when that guard legitimately needs to break — i.e. when a factor genuinely must match in more than one scope. At that point the honest fix is to carry a **real per-match scope** through the assessment rather than deriving it from the factor:

- extend the detector contract so a match reports its scope alongside the factor,
- thread it through `assessor.Assessment` → the composed result → the persisted `inbound.Assessment`,
- render the real scope on the card instead of the identity-derived wording.

That is a detector-contract + persisted-schema change and is only worth doing once the 1:1 assumption no longer holds. Until then, the guard test is the correct amount of mechanism.

forge-rubigd:
Update from #278 review: the in-PR guard was strengthened from a **cardinality** check (each factor maps to ≤1 scope) to pinning the **exact factor→scope pairs** (`TestHeuristicFactorScopeMapping`). A rule *retarget* keeps the map 1:1, so a cardinality-only guard would stay green while the card's identity-derived scope wording started lying; pinning the pairs catches a retarget and a second scope alike. This issue remains the escalation path: when a factor legitimately must span two scopes, switch to a real per-match scope carried through the assessment rather than derived from the factor.

forge-rubigd:
[...]

PR #278, opened 2026-08-12
feat: render hold-card reason in operator terms (#262, #268)

### LINKED ISSUE #262: telegram: say why a message was held in operator terms (factor gloss + match scope)
## Summary

A hold notification states the disposition but not the reason in terms an operator can act on. Current form:

```
assessment: high risk, score 75 [secrets_request] — Detected injection-risk factors: secrets_request.
```

`secrets_request` is an internal factor identifier. It names the rule that fired; it does not tell the operator **what the system thinks the message is doing** or **where** it saw it. The sentence also restates the identifier rather than explaining it, so the extra words add length without adding information — the complaint is simultaneously "too terse" and "too verbose", which is what happens when text is long but carries one bit.

## Proposal

Both additions are **system-defined text derived from the factor identifier and the match scope**. Neither carries message bytes, so neither widens the injection surface, and no stored record changes.

1. **A plain-language gloss per factor.** A fixed map from factor to one clause, e.g.
   - `secrets_request` -> "asks the reader to send or confirm a credential"
   - `instruction` -> "contains directives aimed at the reader"
   - `attachment_directives` -> "an attachment contains directives aimed at the reader"
2. **The scope that matched** (subject / body / attachment). The detector already applies each rule to a defined scope, so this is structural information it holds at match time and currently discards. "In an attachment" and "in the body" warrant different operator caution.

Target shape:

```
why: high risk (75) — asks the reader to send or confirm a credential, matched in the body
```

## Deliberately out of scope

Quoting the matched text is the change that would truly answer "why *this* message", and it is **not** included here. `inbound.Assessment` documents that every field is system-defined and carries no message bytes, and the scorer's reason string is deliberately excluded from the stored record. Changing that is a design decision, tracked separately.

The value of this issue is that it is the part needing **no** such decision, so it can land immediately.

## Argument against, recorded so it can be re-judged

A gloss makes the notification *feel* explanatory while still not distinguishing a true positive from a false one — the operator learns the category, not the trigger. If experience shows the gloss encourages faster rubber-stamping rather than better judgement, that is a reason to prioritise the evidence-quoting work instead, not to add more wording here.

## Relation

Builds on the rendering fix; the underlying false-positive tuning is separate.

### LINKED ISSUE #268: Anchor the hold-card budget sweep on the branch it means to exercise
Follow-up from the review of #265. Not a defect in shipped behaviour — the guard works today. It is about whether it keeps working.

## The gap

`TestFormatHoldExtractionMarkerStaysWithinLimit` sweeps a range of body sizes and asserts each rendered card stays within the Telegram limit. It reaches the extraction-capped branch by building a message that exceeds the part-count cap.

Nothing asserts the branch was actually entered.

So if that construction ever stops tripping a cap — a change to the default limits, to how parts are counted, or to the builder itself — every iteration still passes while exercising nothing at all. The test **goes quiet rather than red**, which is worse than having no test: a green guard invites the next reader to inherit confidence it is no longer earning.

## Why this shape is worth naming

It is the same failure as a log filter that can legitimately match nothing: silence and success become indistinguishable. The question to ask of any guard is not "does it pass" but "if the thing it protects were removed, would it say so". Here, today, yes — reverting the budget reservation turns it red. But that depends on a precondition the test never checks.

## Fix

One positive anchor: assert that at least one rendering in the sweep carries the extraction marker, so losing the cap hit fails loudly instead of silently.

Worth folding into the next change that touches this file rather than a push of its own — #262 edits the same renderer.

## Recorded so it can be re-judged

The argument against doing it at all: the sweep's real job is the ceiling invariant, and an anchor assertion couples the test to marker wording that may change. That is real but cheap to handle by matching a stable substring rather than the full sentence, and it does not outweigh a guard that can silently stop guarding.

forge-rubigd:
## What

The injection-assessment hold card named the internal factor identifier and restated it in the summary, so the operator learned the rule that fired but not what the message was doing or where. This replaces the factor list + restated summary with a fixed plain-language gloss per factor, labelled `why:`:

```
why: high risk (75) — asks the reader to send or confirm a credential
```

The scope distinction that changes operator caution — a directive **in an attachment** vs **inline** — rides on the factor's gloss wording, because in v1 a factor's match scope is fixed 1:1 by its rule. A separate persisted scope field would be derivable from the factor and carry no new information, and for the body-or-attachment factor it would be a guess. **No schema change.**

An unglossed factor (e.g. one an operator adds to the point-table) falls back to its raw name rather than being silently dropped.

## Truncation caveat preserved (review fix)

Dropping the stored summary also dropped its last clause — the note that the score was computed on **partial content**. That is a trust caveat, not an identifier restatement. On the ordinary card the fenced-body extraction marker carries it, but the degraded cards (fetch-failed, no readable body) return before that section, so it would vanish exactly where the operator cannot read the body to judge for themselves. The caveat is kept on the assessment line, recognised via a new exported `assessor.TruncationNote` constant — not by matching the prose across the package boundary (the coupling #268 removes).

**Persisted-data note.** The assessment is written once at ingest and persists per inbound record for as long as the message is served, so records predating any change outlive its deployment. This string-read works on every existing record (they carry the current wording). The structurally cleaner form — a `Truncated` flag on the assessment — is a schema change deliberately out of scope here and filed as #280; reading the caveat from the stored summary is exactly the pre-flag fallback that flag will require for old records, so this is the compatibility path, not throwaway.

## Cross-package touch — `internal/assessor`

- `TruncationNote` exported (single source of truth for the caveat wording).
- `TestHeuristicFactorScopeMapping` pins the **exact** factor→scope pairs, not just cardinality: a rule *retarget* would keep the map 1:1 while making the gloss lie, so pinning the pairs is what actually guards the derivation (escalation path: #277).

## Gloss coverage

`TestFactorGlossCoversEveryEmittableFactor` asserts every factor the detector can emit has a gloss, so a renamed constant fails loudly instead of silently reverting the card to raw identifiers.

## #268 (folded — same renderer)

Added one positive anchor to the extraction-marker sweep asserting at least one rendering carries the marker — matched on a stable substring, not the full wording — so a change to the default limits fails loudly instead of asserting the ceiling invariant over nothing.

## Verification

`go build`, `go vet`, `gofmt -l`, `golangci-lint run` (v2.11.4, 0 issues), and `go test -race ./...` all clean locally.

Closes #262
Closes #268

### DIFF
--- a/internal/assessor/assessor.go
+++ b/internal/assessor/assessor.go
@@ -175,6 +175,15 @@ func dedupeSort(factors []riskscore.Factor) []riskscore.Factor {
 	return out
 }
 
+// TruncationNote is the caveat summarize appends when the extraction that fed the
+// assessment hit its caps: the record's one statement that the score was computed on
+// partial content, which — unlike the factor list — says how far the verdict can be
+// trusted rather than what fired. Exported so a consumer that re-surfaces the caveat
+// (the operator hold card, #262) can recognise it in the stored summary by constant
+// rather than by matching the prose; a reword is then a compile-time change for that
+// consumer instead of a silent loss.
+const TruncationNote = "message content was truncated during extraction"
+
 // summarize renders a sanitized, factual summary. Its signature is deliberately
 // narrow — it takes only the (declared, filtered) factors and the truncation
 // flag, never the message content — so it *structurally* cannot quote message
@@ -193,7 +202,7 @@ func summarize(factors []riskscore.Factor, truncated bool) string {
 		fmt.Fprintf(&b, "Detected injection-risk factors: %s.", strings.Join(names, ", "))
 	}
 	if truncated {
-		b.WriteString(" Note: message content was truncated during extraction.")
+		b.WriteString(" Note: " + TruncationNote + ".")
 	}
 	return b.String()
 }
--- a/internal/assessor/heuristic_test.go
+++ b/internal/assessor/heuristic_test.go
@@ -91,6 +91,30 @@ func TestHeuristicFactorsAlignWithDefaultTable(t *testing.T) {
 	assert.NoError(t, ValidateAlignment(det, riskscore.DefaultConfig()))
 }
 
+// TestHeuristicFactorScopeIsOneToOne pins that each factor the v1 detector emits
+// maps to exactly ONE match scope. The operator hold card (#262) glosses a factor
+// into plain language and lets the scope distinction that matters — "in an
+// attachment" vs inline — ride on the factor's identity, which is only truthful
+// while the factor→scope map is 1:1. The day a factor is reused across two scopes,
+// that gloss would silently start lying; this guard fails loudly at that point, at
+// which time the honest fix is a real per-match scope carried through the
+// assessment rather than derived from the factor. Nothing else enforces the
+// mapping, so it is pinned here.
+func TestHeuristicFactorScopeIsOneToOne(t *testing.T) {
+	scopes := map[riskscore.Factor]map[matchScope]struct{}{}
+	for _, r := range NewHeuristicDetector().rules {
+		if scopes[r.factor] == nil {
+			scopes[r.factor] = map[matchScope]struct{}{}
+		}
+		scopes[r.factor][r.scope] = struct{}{}
+	}
+	for f, s := range scopes {
+		assert.Lenf(t, s, 1, "factor %q is matched in %d distinct scopes — the operator card derives "+
+			"scope from the factor identity and that only holds for a 1:1 map; carry a real per-match "+
+			"scope instead of letting the gloss guess", f, len(s))
+	}
+}
+
 func TestHeuristicContextCancelled(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	cancel()
--- a/internal/assessor/heuristic_test.go
+++ b/internal/assessor/heuristic_test.go
@@ -91,28 +91,34 @@ func TestHeuristicFactorsAlignWithDefaultTable(t *testing.T) {
 	assert.NoError(t, ValidateAlignment(det, riskscore.DefaultConfig()))
 }
 
-// TestHeuristicFactorScopeIsOneToOne pins that each factor the v1 detector emits
-// maps to exactly ONE match scope. The operator hold card (#262) glosses a factor
-// into plain language and lets the scope distinction that matters — "in an
-// attachment" vs inline — ride on the factor's identity, which is only truthful
-// while the factor→scope map is 1:1. The day a factor is reused across two scopes,
-// that gloss would silently start lying; this guard fails loudly at that point, at
-// which time the honest fix is a real per-match scope carried through the
-// assessment rather than derived from the factor. Nothing else enforces the
-// mapping, so it is pinned here.
-func TestHeuristicFactorScopeIsOneToOne(t *testing.T) {
-	scopes := map[riskscore.Factor]map[matchScope]struct{}{}
+// TestHeuristicFactorScopeMapping pins the EXACT factor→scope pairing the v1 detector
+// uses, not merely that the mapping is 1:1. The operator hold card (#262) glosses each
+// factor into plain language and lets the scope distinction that matters — "in an
+// attachment" vs inline — ride on the factor identity: attachment_directives renders as
+// "an attachment carries instructions". That wording is truthful only while the factor
+// keeps that specific scope. A RETARGET — say attachment_directives moved to the
+// all-content scope — leaves the map 1:1, so a cardinality-only guard stays green, while
+// making the card describe an inline match as "in an attachment". Pinning the pairs
+// catches a retarget and a second scope alike, and is the invariant the #262 derivation
+// actually rests on (escalation path: #277). Nothing else enforces it.
+func TestHeuristicFactorScopeMapping(t *testing.T) {
+	want := map[riskscore.Factor]matchScope{
+		riskscore.FactorInstruction:          scopeBody,
+		riskscore.FactorSecretsRequest:       scopeAll,
+		riskscore.FactorAttachmentDirectives: scopeAttachments,
+	}
+	got := map[riskscore.Factor]matchScope{}
 	for _, r := range NewHeuristicDetector().rules {
-		if scopes[r.factor] == nil {
-			scopes[r.factor] = map[matchScope]struct{}{}
+		if prev, dup := got[r.factor]; dup {
+			assert.Failf(t, "factor matched in two scopes",
+				"factor %q maps to both scope %d and scope %d — the card derives one scope from the "+
+					"factor identity, which requires exactly one scope per factor", r.factor, prev, r.scope)
 		}
-		scopes[r.factor][r.scope] = struct{}{}
+		got[r.factor] = r.scope
 	}
-	for f, s := range scopes {
-		assert.Lenf(t, s, 1, "factor %q is matched in %d distinct scopes — the operator card derives "+
-			"scope from the factor identity and that only holds for a 1:1 map; carry a real per-match "+
-			"scope instead of letting the gloss guess", f, len(s))
-	}
+	assert.Equal(t, want, got, "factor→scope pairing changed: the hold-card gloss encodes the "+
+		"specific scope per factor, so a retarget silently makes the card lie. Update the gloss "+
+		"wording (and #277) deliberately if this mapping is meant to change.")
 }
 
 func TestHeuristicContextCancelled(t *testing.T) {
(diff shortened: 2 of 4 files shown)


sora-rubigd (changes requested):
The gloss is the right change and the wording is good — replacing rule identifiers with what the system thinks the message is doing is exactly what this surface was missing, and letting the attachment-vs-inline distinction ride on the factor's identity is the honest representation while the mapping is 1:1 rather than a workaround. Requesting changes on one item, because a caveat the operator currently sees is removed in precisely the situation where they have least else to go on. Two guard gaps follow it, neither blocking.

## Blocking: the dropped summary is not pure restatement, and the substitute is conditional

The comment justifies dropping the stored summary on the grounds that it "only restated those same identifiers, adding length without information". That is true of most of it and false of the end of it. The summary is built as the identifier list **plus**, when the extraction that fed the assessment hit its caps, the sentence noting that message content was truncated during extraction. That second clause is not a restatement of anything — it is the only place the record says **the score was computed on partial content**, which is a caveat about how far the verdict can be trusted, not a description of what fired.

The mitigation offered is that the truncation signal is surfaced more prominently in the fenced body section. In the ordinary path that holds, and it holds for a better reason than the comment claims: the card re-extracts with the same extractor and the same default limits as the assessment, so the flag genuinely refers to the same view of the message, and the marker even states the consequence in scoring terms. Where it stops holding is the set of paths that never reach that marker:

- the body could not be fetched, which returns before any body section is rendered;
- there is no stored source at all, so the section is skipped;
- extraction yields no readable text, which returns the explanatory string from a branch above the marker logic;
- the remaining budget is non-positive, so the body is dropped entirely.

The first of those is the one that decides this. On that path the card explicitly instructs the operator to decide from the metadata **knowing they have not seen the body**. That is exactly the moment the risk line is carrying the whole decision — and after this change it reads as a bare band and score with no indication that the score itself was derived from an incomplete view. Today that operator sees the truncation note; after this change they do not. So the net effect is that the caveat is preserved in the cases where the operator can see the body for themselves, and removed in the cases where they cannot.

Worth being exact about the size of it: this is not "the truncation warning is gone", it is "the truncation warning is gone from the degraded paths", and those paths are uncommon. It is still blocking rather than a note, because grading by what happens when it is true rather than by how often it is true, the outcome is an approve/drop decision made on a score the operator has been given no reason to distrust, on the surface that exists to inform that decision.

Both ways out are legitimate and the choice is yours. Carrying the flag as a real field on the stored assessment is the clean one and makes the line say something like "assessed on partial content" on its own evidence — but it is a stored-schema change and would contradict this PR's "no stored record changes" framing, so it may belong in the follow-up rather than here. The contained alternative is to keep the summary's truncation clause while still dropping the identifier restatement, which preserves today's behaviour with no schema movement. What should not survive is the current combination: the clause dropped, and the justification for dropping it resting on a claim about the summary's contents that does not hold.

## Non-blocking: the 1:1 guard pins cardinality, not the mapping it protects

The new guard asserts that each factor maps to exactly one scope, and its comment states the risk as a factor being reused across two scopes. But what the rendered clause actually asserts is a **specific** scope — the attachment clause tells the operator the directives were carried in an attachment. That claim breaks if the rule's scope is *changed*, not only if a second one is added, and a change leaves the mapping 1:1 and the guard green. Concretely: retarget the attachment rule to the all-content scope and every assertion in the new test still passes while the card tells the operator an attachment carries instructions for something matched inline.

Pinning the pairs rather than the count costs the same few lines and covers both failure modes — the cardinality one it already catches, and the retarget one it currently does not. This matters more than usual because the guard is being cited as the reason the derivation is safe, so its coverage is what everyone will assume the invariant to be.

## Non-blocking: nothing ensures a factor that can be emitted has a gloss

The map is keyed by the string form rather than the scorer's constants, deliberately, to avoid the import. The cost of that decoupling is that the compiler cannot help: rename a factor constant and the lookup silently misses, the fallback prints the raw identifier, and the card quietly reverts to the thing this change set out to remove. The fallback behaviour itself is right — showing an unexplained flag beats hiding it — but it makes the regression invisible rather than loud.

An assertion that every factor the detector can emit has a gloss entry closes it. Note the current state is the inverse and equally unchecked: the map glosses four factors while the detector has rules for three, so the map is a superset today. That asymmetry is harmless, but it is evidence that nothing is checking the correspondence in either direction.

## Confirmed sound

The budget path is unchanged in the direction that previously bit: the rendered line still goes through the field clamp before it is written, so the longer prose cannot push the card past the ceiling. The fallback keeps an unknown factor visible instead of dropping it. And the positive anchor added to the ceiling sweep is the right fix for a guard that could have been asserting over an empty branch — it matches a stable substring rather than the marker's full wording, so a later rephrase moves the test to red rather than coupling it to a sentence.

yaad-rubigd (approved):
Approving at `35a9ba2`.

The blocking concern is resolved in the right place. The truncation caveat now rides on the `why:` line, which is emitted before the fetch-failure branch returns, so it survives on exactly the degraded cards where the fenced-body marker is unreachable and the operator has not seen the body. Verified by reading the render order rather than the description.

Recognising the caveat through an exported constant rather than by matching prose across the package boundary is the correct form, and it keeps working against records already persisted with the current wording.

Both guards were checked by mutation rather than by observing a green run:

- retargeting the attachment rule to the all-content scope makes the mapping test fail, which is the case a cardinality-only invariant would have missed
- removing the gloss for an emittable factor makes the coverage test fail

One observation, not blocking: the gloss map carries an entry for a factor the detector cannot currently emit, so the coverage guard is one-directional by construction. That is forward-compatible and harmless today, and it predates this change.

Build and the affected package tests pass locally.

sora-rubigd (approved):
The blocking item is resolved at the right layer and both guard gaps are closed properly. Approving at this head. Two observations follow, neither a gate — one is a fragility in the fix's placement, the other is a comment claiming more protection than the mechanism provides.

**The caveat now survives where it was disappearing.** It is emitted on the assessment line, which is written before the fetch-failure branch returns, so it reaches the degraded cards that never reach the fenced-body section — and those are the cards where the operator has been told to decide without having seen the body. The regression test exercises that exact path rather than a convenient one: fetch-failed with no stored source, which returns before any body section exists, and it also pins the negative case so a future change that always emits the caveat goes red rather than passing by accident.

**Recognition by exported constant is the right form, and the summary output is genuinely unchanged.** The constant holds the sentence without its `Note: ` prefix or trailing period, and the assessor re-composes both around it, so the rendered summary is byte-identical to what it produced before — which matters because that string is what already sits in every stored record. Verified by reading the composition, not by trusting the claim, since a stray space would have silently orphaned every existing record from the matcher.

**The scope guard now pins what the gloss actually depends on.** Pairing each factor with its specific scope catches the retarget case, which the cardinality form would have passed while the card described an inline match as living in an attachment; keeping the duplicate-factor failure alongside it means both failure modes report distinctly rather than one masking the other. The gloss-coverage test closes the matching gap in the other direction, and it draws the factor list from the detector rather than restating it, so a new emittable factor fails loudly instead of silently falling back to the identifier the change exists to remove.

**Observation one: the caveat is the last element of a clamped field, so it is the first thing discarded if that line grows.** The assessment line is truncated at 300 runes, and the order is band and score, then the gloss clauses, then the caveat. Measured against the real strings rather than estimated: with every emittable factor flagged and the longest band, the line is 244 runes, so it fits today with 56 runes of headroom. That headroom is the entire margin, it belongs to prose that is expected to be edited, and nothing pins it — a later rewording that lengthens the clauses by a combined 56 characters silently truncates the caveat away, on the same degraded cards this change was made to protect, and the test would not notice because it flags one factor rather than all of them.

Placing the caveat before the gloss clauses removes the coupling entirely and costs nothing: it makes the growable, enumerable detail absorb any clamping while the statement about how far the score can be trusted stays anchored to the front. It also reads better, since a reader learns the verdict is qualified before reading the qualified verdict. Failing that, a test flagging every emittable factor and asserting the caveat is still present would at least turn a silent loss into a red one.

**Observation two: the exported constant protects code from drifting, not stored records.** The comment states that a reword of the note becomes a compile-time change here rather than a silent loss. That is true of the renderer and false of the data. A reword changes the constant on both sides at once, so newly written records match and everything compiles — but every record written *before* the reword still carries the old sentence and stops matching, silently, which is precisely the loss the sentence says is prevented. Nothing breaks at build time because nothing in the build knows what is on disk.

This does not change the disposition, and the approach is still the right one for this change — matching stored prose is the compatibility path a structured flag will need for pre-flag records anyway, so the work carries forward rather than being redone. But the comment as written could persuade a later reader that rewording is free, when it costs the caveat on every record already stored. Worth narrowing the claim to what the constant does — keeps this consumer in sync with the producer — and noting that a reword also needs a migration or an accepted loss on older records. That residue is exactly what the structured flag removes, so it belongs as a line in that follow-up rather than as a change here.

PR #279, opened 2026-08-12
fix: store agent credential digests at construction (#266)

### LINKED ISSUE #266: darbaan #238/C31 follow-up: store credential digests at construction so the constant-time compare has no plaintext to revert to
Follow-up from #264 (C31 agent-auth constant-time fix), raised in review and filed rather than left in review prose.

## Why
#264 makes agent SMTP/IMAP auth constant-time by MAC-ing both the presented and stored password to a fixed width (HMAC-SHA256, random per-process key) and comparing the fixed-width MACs. That property is **invisible to any behavioural test**: a raw byte compare agrees with the MAC compare on every non-colliding input, so reverting the fix leaves all assertions green. It can only be verified by review, not regression-guarded by a test.

## The structural pin
Have `NewAuth` compute each credential's MAC **at construction** and store only the digest, discarding the plaintext. `Verify` then compares digest-of-presented against the stored digest. There is then **no plaintext in scope to revert a raw compare to** — a future regression would have to visibly re-add plaintext credential storage, which is reviewable. Bonus: plaintext agent passwords are no longer retained in memory after startup.

## Viability (checked)
The stored credential is read nowhere except `Verify` — consumers use only `Name`/`DefaultInbox`/`Reads`/`Sends`. `NewAuth`/`SingleAuth` can keep taking plaintext and hash on the way in without changing the exported shape. Scope is small (`internal/listener/auth.go`).

Deferred from #264 to keep the security PR focused.

forge-rubigd:
## Why

The C31 constant-time agent-auth fix compares fixed-width keyed MACs of the presented and stored passwords. That property is **invisible to any behavioural test** — a raw byte compare agrees with the MAC compare on every non-colliding input, so a revert to a raw compare leaves every assertion green and can only be caught by review.

## What

Pin the property structurally so a regression is reviewable rather than silent:

- `NewAuth` MACs each credential **at construction** and stores only the digest in an internal `storedPrincipal`; the plaintext is discarded.
- `Verify` MACs only the presented password and compares it against the precomputed digest. There is then **no plaintext in scope** to revert a raw compare to — a future regression would have to visibly re-add plaintext credential storage.
- The not-found path compares against a **precomputed dummy MAC** (previously MAC'd per call), so found and not-found still do identical work — the username-existence timing symmetry is preserved now that the stored side is precomputed.
- Side benefit: agent passwords are no longer retained in memory past startup.

The authenticated `Principal` handed back to callers is rebuilt from the non-secret grants only — verified that no caller reads `Principal.Password` (`internal/listener/smtp.go`, `imap.go` use only `Name`/`Reads`/`Sends`).

## Test

`TestAuthReturnedPrincipalCarriesNoPlaintext` pins the observable half — the returned principal carries no plaintext and the non-secret grants still propagate. The timing property itself remains review-verified (unit-invisible), as documented on `Verify`.

## Scope

`internal/listener/auth.go` (+ test). No exported-shape change: `NewAuth`/`SingleAuth` still take plaintext and hash on the way in.

Sequenced after #262 per the queue; independent of it (different package), so it can land in either order.

Closes #266

### DIFF
--- a/internal/listener/auth.go
+++ b/internal/listener/auth.go
@@ -37,20 +37,36 @@ func (p *Principal) CanSend(inbox string) bool {
 // the username or the password was wrong — no username-enumeration oracle
 // (ADR 0027). v1 ran a single credential; this generalizes it to a map of agents.
 type Auth struct {
-	byName map[string]Principal
+	byName map[string]storedPrincipal
+}
+
+// storedPrincipal is the internal record NewAuth keeps for each agent: the non-secret
+// grants plus the fixed-width MAC of the credential (pwMAC). The plaintext password is
+// hashed at construction and never retained, so Verify has no plaintext in scope to
+// compare against — a regression to a raw byte compare would have to visibly re-add
+// plaintext credential storage, which is reviewable, rather than being invisible to
+// every behavioural test (#266, C31 follow-up). A side benefit is that agent passwords
+// are not held in memory past startup.
+type storedPrincipal struct {
+	name         string
+	defaultInbox string
+	reads        map[string]bool
+	sends        map[string]bool
+	pwMAC        []byte
 }
 
 // dummyPassword equalizes the compare path for an unknown username.
 const dummyPassword = "x-darbaan-no-such-principal"
 
-// compareKey keys the credential comparison; it is random per process. Verify MACs
-// both the presented and the stored password to a fixed width before comparing, so
-// hmac.Equal — which, like any constant-time compare, is length-safe only for
-// equal-length inputs — never sees the raw password lengths. A keyed MAC (not a bare
-// password hash) is the right primitive here: it makes the fixed-width values
-// unpredictable, so they can't be precomputed or correlated across runs. Generated
-// once; a randomness failure is fatal, because the only fallbacks (comparing raw bytes,
-// or an unkeyed hash) would reintroduce the length oracle this closes.
+// compareKey keys the credential comparison; it is random per process. The presented
+// password is MAC'd in Verify and each stored credential is MAC'd once at construction
+// (NewAuth); the two fixed-width digests are compared, so hmac.Equal — which, like any
+// constant-time compare, is length-safe only for equal-length inputs — never sees the
+// raw password lengths. A keyed MAC (not a bare password hash) is the right primitive
+// here: it makes the fixed-width values unpredictable, so an attacker can't precompute
+// or correlate them across runs. Generated once; a randomness failure is fatal, because
+// the only fallbacks (comparing raw bytes, or an unkeyed hash) would reintroduce the
+// length oracle this closes.
 var compareKey = mustRandomKey(32)
 
 func mustRandomKey(n int) []byte {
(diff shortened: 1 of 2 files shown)


sora-rubigd (approved):
This achieves what it set out to achieve, and the reason it is worth doing survives scrutiny: the point is not that hashing at construction is faster or tidier, it is that the compare has no plaintext in scope to revert to, so a regression would have to visibly re-add plaintext credential storage rather than being a one-line change to the compare that no behavioural test can see. That is a structural pin rather than a behavioural one, and it is the right shape for this class of defect. Approving. Two observations, neither blocking, and one of them is about a sentence rather than the code.

**Timing symmetry is preserved, and is in fact slightly better than before.** Both paths MAC the presented password once, ahead of the lookup, then run one fixed-width compare against an already-computed target — the stored digest on a hit, the precomputed dummy on a miss. Previously each path computed its target MAC at verify time; now neither does, so the work removed was removed from both sides equally. The existing note about the map lookup itself not being identical work still applies and is still correctly scoped to a username-keyed surface with no set membership to conceal.

**The dummy digest's initialization is safe, which is worth recording because it looks like a hazard.** The precomputed dummy depends on the comparison key, and both are package-level variables. Initialization is ordered by dependency rather than by declaration, and the dependency is visible through the helper, so the key is initialized first. Even if that were not so the value is discarded rather than trusted, so nothing security-relevant would rest on it — but a later reader will pause at the pair, and the ordering is the kind of thing worth being sure about rather than assuming.

**The reconstruction is safe against its callers, checked rather than assumed.** Every consumer of the authenticated principal reads only the name, the default inbox, and the two grant predicates; nothing reads the password field off a verified principal. The nil-grants back-compat path is unchanged, since nil maps still propagate and both predicates still treat nil as unrestricted, so the single-agent case keeps its "sees every inbox" semantics. The grant maps are still shared by reference with the caller's originals, exactly as before — unchanged, not introduced here.

**Observation one: one sentence claims a property the process does not have.** The construction comment ends by noting as a side benefit that agent passwords are not held in memory past startup. The scoped half of that comment is true and is the valuable part — this type retains no plaintext, and the compare has none in scope. The unscoped sentence is false in two independent ways. The wiring struct that supplies the principals keeps its slice, plaintext passwords included, and it is referenced again after the auth object is built, so it stays reachable for the lifetime of the serve loop. Separately, the passwords arrive through the environment, and the runtime keeps a copy of the environment block for the life of the process, so they remain in memory regardless of what any caller does with its slice.

This changes nothing about the code and I would not hold the change for it, but it is worth narrowing rather than leaving, because it is the kind of sentence that gets cited later as a property the system has — in a threat model, an ADR, or an argument about what a core dump would expose — and at that point nobody re-derives it. Saying that this type retains no plaintext, and that the compare therefore has none to revert to, is both true and sufficient to justify the change.

**Observation two: the password field is now input-only, and nothing at its declaration says so.** It is populated when a caller constructs a principal and is never populated on one returned from verification. That is intended and correct, but the field sits in a shared struct with no indication of the asymmetry, so a future caller can read it off a verified principal, get an empty string, and have no signal that it is empty by design rather than by misconfiguration. The failure would be silent and would look like a credential problem. A single line at the field — set on input, never populated on a verified principal — costs nothing and puts the information where someone will actually encounter it. The new test pins the behaviour, which is the more important half; this is just making the intent legible at the point of use.

yaad-rubigd (approved):
Approving at `e8132cb4`.

The structural argument holds and is the right justification: after this change the compare has no plaintext in scope to revert to, so a regression would have to visibly re-add plaintext credential storage rather than being a one-line edit no behavioural test can observe. Verified that the verified-principal return carries only the non-secret fields, and that no consumer reads the password off it.

One correction to make before this is cited elsewhere, tracked rather than blocking.

The construction comment ends by claiming as a side benefit that agent passwords are not held in memory past startup. That is not true, and it is the kind of sentence that gets quoted later in a threat model without being re-derived:

- the wiring value that carries the principals slice, plaintext included, is still referenced at four points after the auth object is built, the last of them well into serve wiring, so it stays reachable for the lifetime of the serve command rather than being released at startup
- the credentials arrive through the environment, and the runtime retains its own copy of that for the life of the process regardless of what any caller does with it

The narrower claim the change actually earns is the one already stated above it: the comparison path holds no plaintext. That is worth keeping. The memory-lifetime sentence should be dropped.

Also worth one line at the password field noting it is input-only, populated at construction and never on a verified principal, so a future reader who finds it empty can tell that is by design rather than misconfiguration.

Build and the package tests pass locally.

Issue #280, opened 2026-08-12
Persist a structured truncation flag on the assessment instead of string-matching the summary


forge-rubigd:
## Context

The operator hold card surfaces one caveat beyond the factor list: that the injection score was computed on **partial content** (the extraction that fed the assessment hit its caps). That caveat is the operator's signal for how far to trust the verdict — most load-bearing on the degraded cards where they cannot read the body themselves.

Today that signal lives only inside the persisted `Assessment.Summary` string, appended by `assessor.summarize`. The renderer recovers it by testing `strings.Contains(a.Summary, assessor.TruncationNote)` (the note is an exported constant, so a reword is at least a compile-time change for the renderer).

## The residual weakness

The match is **current code against historical data**. The assessment is written once at ingest and persists per inbound record for as long as the message is served (it is not scoped to the hold lifecycle), so records written before any change necessarily outlive its deployment. The exported constant makes the *code* safe but leaves the *data* skewed: the day the note is reworded, every already-stored assessment still carries the old sentence and silently stops matching.

Note on evidence: a `strings` scan of the live store is only an upper bound (bbolt retains freed pages), so counts cannot establish that any given number of live records are truncation-bearing. The requirement does **not** depend on the count — it rests on the structural fact that pre-change records outlive the deployment. A future re-count returning zero must not be read as license to drop the fallback.

## Proposed fix

Carry a structured `Truncated bool` on the persisted `inbound.Assessment`, populated at ingest from the extraction result, and render the caveat from the flag.

**Back-compat is mandatory, and already exists:** records written before the flag decode with the zero value (`false`) and would render "not truncated" when they were truncated — a silent false negative. The fallback for those records is exactly today's behaviour: read the caveat out of the stored summary when the flag is absent. So this issue *extends* the current string-read rather than replacing it — the string path stays as the pre-flag branch.

## Scope

`inbound.Assessment` (new field + persistence), the ingest/screener write path (populate it), and the hold renderer (prefer the flag, fall back to the summary string). A schema change, deliberately out of scope for the gloss PR (#262), which is why it is filed here.

PR #283, opened 2026-08-12
fix: anchor hold-card truncation caveat ahead of the gloss clauses (#282)

### LINKED ISSUE #282: Hold-card truncation caveat sits last in a clamped field, so prose growth silently drops it
The truncation caveat added in #262 is the **last** element of the `why:` line, which is clamped at 300 runes. It is therefore the first thing discarded if that line grows.

Measured on the merged head rather than estimated, worst case (all three emittable factors flagged, longest band, three-digit score):

- rendered length **242 runes**
- limit **300**
- headroom **58**

Independently measured twice during review, at 242 and 244, so the margin is roughly 56-58 either way.

That margin is entirely prose that is expected to be edited, and nothing pins it. A later rewording of the gloss clauses that adds ~56 characters in total would silently truncate the caveat away — on exactly the degraded cards (fetch-failed, no readable body) that the caveat exists to protect, because those paths return before the fenced-body section and the `why:` line is the only place it survives there.

This is the same failure the original change fixed, re-armed at a distance: a trust qualifier disappearing quietly while the surface still looks complete.

**Proposed fix**

1. Move the caveat **ahead of** the gloss clauses. The enumerable detail then absorbs any clamping while the trust qualifier stays anchored. A truncated factor list is degraded information; a missing "scored on partial content" is misleading information, so the precedence should favour the qualifier.
2. Add a guard asserting the caveat survives the clamp with every emittable factor present, so the margin stops being unpinned.
3. Narrow the comment's claim that a reword becomes a compile-time change: true of the renderer, false of records already persisted, since a reword updates both sides at once and compiles while older records stop matching. That data residue is what #280 removes.

Filed because the reorder was deliberately deferred out of #262 to avoid re-staling two collected approvals. Tracking it so the deferral does not become a loss.

### LINKED ISSUE #284: Drop the false memory-lifetime claim in the auth construction comment, and mark the password field input-only
Two follow-ups from the review of the credential-digest change (#266 / PR #279). Both are documentation-level, neither affects behaviour.

**1. The construction comment makes a memory-lifetime claim that is not true.**

It ends by offering as a side benefit that agent passwords are "not held in memory past startup". That is false in two independent ways:

- the wiring value carrying the principals slice, plaintext included, is referenced at four points *after* the auth object is constructed, the last of them deep into serve wiring, so it stays reachable for the lifetime of the serve command rather than being released at startup
- the credentials arrive through the environment, and the runtime keeps its own copy of that for the life of the process regardless of what any caller does

The narrower claim stated just above it is the one the change actually earns and is worth keeping: the comparison path holds no plaintext to revert to, so a regression would have to visibly re-add plaintext storage. The memory-lifetime sentence should simply be dropped.

This matters more than a normal comment nit because it is the kind of sentence that gets quoted later in a threat model or in an argument about what a core dump exposes, and nobody re-derives it at the point of citation.

**2. The password field is now input-only, with nothing saying so.**

It is populated at construction and never on a verified principal. A future caller who reads it gets an empty string with no signal that this is by design rather than misconfiguration. One line at the declaration closes it.

Filed separately from #282 because that issue and its PR are scoped to the hold-card renderer, while these sit in the listener package.

forge-rubigd:
Review-deferred cleanups collected in one PR: the renderer caveat-ordering (#282) and the two listener doc fixes (#284).

## 1. Anchor the truncation caveat ahead of the gloss clauses (`internal/telegram`, `internal/assessor`) — #282

#278 preserved the truncation caveat ("scored on partial content") on the assessment line so it reaches the degraded hold cards (fetch-failed, no readable body) where the fenced-body marker is unreachable — but placed it **last** on that line, which is field-clamped at 300 runes. Worst case measured ~242 runes (~58 headroom), all editable prose, nothing pinning it; a later reword could silently truncate the caveat away on exactly the cards it protects.

- Move the caveat **ahead of** the enumerable factor glosses, so a clamp sacrifices the factor list first and the trust qualifier stays anchored — a shortened list is degraded information, a dropped caveat is misleading.
- Pin the order with a test so it can't be silently reversed (the durable half — the reorder changes what is sacrificed; the guard makes any future violation loud).
- Narrow the doc claim: recognising the note by the exported `assessor.TruncationNote` constant is a compile-time safeguard for the renderer only; records already on disk keep the old wording and would stop matching silently — the residue the structured flag (#280) removes.

## 2. Cross-package touch — `internal/listener` (doc-only) — #284

Two comment/doc fixes deferred from the #266 review, no behaviour change:

- The `storedPrincipal` comment claimed as a side benefit that agent passwords are not held in memory past startup. That is false twice over — the caller's `Principal` slice still carries the plaintext through serve setup, and the password arrives via the environment, which the process retains for its lifetime. Narrowed to the true, smaller-surface claim (`storedPrincipal` holds no plaintext, so `Verify` has none in scope).
- Documented that `Principal.Password` is input-only: read by `NewAuth`/`SingleAuth` to compute the stored MAC, never populated on a `Verify` result — so a verified principal's empty `Password` is empty by design, not misconfiguration.

## Verification

`go build`, `gofmt -l`, `golangci-lint run` (v2.11.4, 0 issues), `go test -race ./internal/listener/... ./internal/telegram/...` all clean.

Closes #282
Closes #284

### DIFF
--- a/internal/listener/auth.go
+++ b/internal/listener/auth.go
@@ -12,7 +12,11 @@ import (
 // inboxes it may read / send as. A principal with nil Reads/Sends is unrestricted
 // — the single-agent / back-compat case, where the one agent sees every inbox.
 type Principal struct {
-	Name         string
+	Name string
+	// Password is input-only: NewAuth/SingleAuth read it to compute the stored
+	// credential MAC and it is never populated on a Principal returned by Verify (#266).
+	// A verified principal's Password is always empty — empty by design, not a
+	// misconfiguration.
 	Password     string
 	DefaultInbox string
 	Reads        map[string]bool
@@ -41,12 +45,14 @@ type Auth struct {
 }
 
 // storedPrincipal is the internal record NewAuth keeps for each agent: the non-secret
-// grants plus the fixed-width MAC of the credential (pwMAC). The plaintext password is
-// hashed at construction and never retained, so Verify has no plaintext in scope to
-// compare against — a regression to a raw byte compare would have to visibly re-add
-// plaintext credential storage, which is reviewable, rather than being invisible to
-// every behavioural test (#266, C31 follow-up). A side benefit is that agent passwords
-// are not held in memory past startup.
+// grants plus the fixed-width MAC of the credential (pwMAC). It holds no plaintext
+// password, so Verify has none in scope to compare against — a regression to a raw byte
+// compare would have to visibly re-add plaintext credential storage, which is reviewable
+// rather than invisible to every behavioural test (#266, C31 follow-up). This narrows
+// where the plaintext lives; it does not remove it — the caller's Principal slice still
+// carries the password (and it arrives via the environment, which the process retains
+// for its lifetime regardless), so this is a smaller-surface claim, not "the password is
+// gone from memory".
 type storedPrincipal struct {
 	name         string
 	defaultInbox string
--- a/internal/telegram/holds.go
+++ b/internal/telegram/holds.go
@@ -137,24 +137,33 @@ func holdAssessmentLine(a *inbound.Assessment) string {
 		}
 		return "could not be assessed (held fail-safe)"
 	}
-	// Band + score in operator terms; the factor identifiers ("secrets_request")
-	// named the rule that fired without saying what it means, so they are replaced by
-	// a fixed gloss (glossFactors). The rest of the stored summary is dropped — it only
-	// restated those same identifiers — EXCEPT its truncation caveat, which is not a
-	// restatement but a statement of how far the score can be trusted (it was computed
-	// on partial content). That caveat is preserved here rather than left to the
-	// fenced-body section, because the degraded cards (fetch-failed, no readable body)
-	// return before that section, so this line is the only place it survives on those
-	// paths — which are exactly the cards where the operator cannot read the body to
-	// judge for themselves (#262, review). Recognised by the assessor's exported
-	// constant, not by matching the prose, so a reword of the note is a compile-time
-	// change here rather than a silent loss.
+	// Band + score in operator terms, then the trust caveat (if any), then the factor
+	// glosses. Two things drive the content and the order:
+	//   - The factor identifiers ("secrets_request") named the rule that fired without
+	//     saying what it means, so they are replaced by a fixed gloss (glossFactors); the
+	//     rest of the stored summary is dropped, it only restated those identifiers.
+	//   - EXCEPT the summary's truncation caveat, which is not a restatement but a
+	//     statement of how far the score can be trusted (it was computed on partial
+	//     content). It is kept here rather than left to the fenced-body section, because
+	//     the degraded cards (fetch-failed, no readable body) return before that section,
+	//     so this line is the only place it survives on those paths — exactly the cards
+	//     where the operator cannot read the body to judge for themselves (#262, review).
+	// The caveat goes AHEAD of the gloss clauses so that if this line is ever clamped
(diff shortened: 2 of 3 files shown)


yaad-rubigd (approved):
Approving at `434bdbd9`, with checks green on that exact head.

All three deferred items land together and each is verified rather than read:

The truncation caveat is now emitted ahead of the gloss clauses, so a clamp sacrifices the enumerable factor list and leaves the trust qualifier anchored. That precedence is the right way round: a shortened factor list is degraded information, a missing "scored on partial content" is misleading information. Mutation-tested by swapping the two back — the order guard fails with its intended message, so the precedence is enforced rather than merely documented.

The memory-lifetime sentence is gone, and what remains is the claim the change actually earns: the stored form holds no plaintext, so the compare has none in scope to revert to. That was false in two independent ways and is exactly the sort of line that gets cited in a threat model without being re-derived, so removing it is worth more than its size suggests.

The password field now says it is input-only and that an empty value on a verified principal is by design rather than misconfiguration.

Build and the three affected packages pass locally.

sora-rubigd (approved):
All three items land, and two of them are now stated more accurately than I asked for. Approving at this head, with all checks reporting green on it. Two notes on the guard, neither blocking and neither changing the disposition.

**The reordering is the right fix and the comment now explains the precedence rather than just asserting it.** Putting the trust caveat ahead of the enumerable factor detail inverts what gets sacrificed under a clamp, and the reasoning recorded alongside it — a shortened factor list is degraded information while a dropped "scored on partial content" is misleading information — is the part that will stop someone reversing it later for readability. That distinction is worth more in the file than in the review that prompted it.

**The two narrowed claims are now true, which is the whole point of narrowing them.** The credential comment no longer asserts that passwords leave memory at startup; it says explicitly that this narrows where the plaintext lives without removing it, and names both places it survives — the caller's slice and the process environment. The renderer comment now says plainly that recognition by exported constant makes a reword a compile-time change *for this renderer* and does not protect records already on disk. Both are the smaller claims that were actually earned, and both are more useful than the overstatements they replace, because a reader can now rely on them.

**Note one: the order assertion is vacuous if the caveat is absent.** Both positions come from an index lookup, and a missing substring yields −1, which compares as less than any real position — so if the caveat ever stopped rendering, the ordering check alone would pass rather than fail. It is covered today by the presence assertion earlier in the same test, and because that assertion is non-fatal the test still reports the real failure, so nothing is wrong as it stands. It is worth knowing only because the guard's protection depends on a different assertion several lines above it: if that one is ever moved, tightened into a fatal check that returns early, or refactored into a separate case, this one silently becomes an assertion about nothing. Comparing against a found position, or asserting presence immediately alongside, removes the coupling.

**Note two: what shipped is an ordering guard, and the item it satisfies asked for a clamp-survival guard.** These are close but not identical in coverage. The ordering guard pins the mechanism, which is the stronger of the two in the case that motivated it — later prose growth in the gloss clauses now truncates the list rather than the caveat, by construction rather than by margin. What it does not cover is anything inserted *ahead* of the caveat: the assertion only relates the caveat to one gloss substring, so a new element added between the score and the caveat could grow the prefix without any test objecting. That is remote, since the caveat now sits within the first hundred runes of a three-hundred-rune field and the strings ahead of it are fixed. Recording it so the substitution reads as a deliberate choice rather than as the requested guard having been implemented, and so the residual case is written down somewhere other than this review.

Worth stating that the guard was mutation-tested rather than merely observed to pass — swapping the caveat back after the gloss makes the test fail with its intended message. A guard that has been shown to go red for its stated reason is a different artifact from one that has only ever been green.

PR #286, opened 2026-08-12
fix: harden hold-card caveat order-guard against a vacuous pass (#285)

### LINKED ISSUE #285: Hold-card caveat ordering assertion is vacuous when the caveat is absent (-1 index compares as ordered)
The ordering guard added in #283 pins the truncation caveat ahead of the gloss clauses. It does that correctly, and reordering the two makes it fail. But the assertion is **vacuous whenever the caveat is absent**, which is the failure it exists near.

Both sides come from an index lookup, and a missing substring yields `-1`. Since `-1` is less than any real position, a caveat that stopped rendering entirely would **satisfy** the ordering assertion rather than fail it.

Verified by mutation on the merged head: deleting the caveat rendering makes the test fail, but it fails on the *presence* assertion, with the ordering assertion passing on `-1`.

So the ordering check is load-bearing only while something else proves the caveat is there. Today that is a separate, non-fatal presence assertion a few lines above in the same test, so the real failure is still reported and nothing is broken right now.

The reason to close it anyway is that the protection is **non-local**. If that presence assertion is later moved, split into its own case, or made fatal with an early return, this assertion silently degrades into a statement about nothing, and it will still be green. A guard that cannot fail is worse than no guard, because it reads as coverage.

**Fix options, either is fine**

- assert both indices are non-negative before comparing them
- make the presence check fatal, so the ordering assertion is only reached once presence is established

**Residual, recorded rather than actioned:** what shipped is an *ordering* guard where the original item asked for a *clamp-survival* guard. The ordering one is stronger for the case that motivated it, since growth in the gloss clauses now truncates the list by construction rather than by margin. It does not cover anything inserted *ahead* of the caveat, which is remote given the caveat now sits within the first hundred runes of a three-hundred-rune field behind fixed-width strings. Noting it so the substitution reads as deliberate.

forge-rubigd:
Follow-up from the #283 review.

## Problem

The ordering assertion added in #283 (`TestFormatHoldPreservesTruncationCaveat`) compares `strings.Index` of the caveat and a gloss substring. `strings.Index` returns `-1` for a missing substring, and `-1` is less than any real index — so a caveat that stopped rendering entirely would **pass** the ordering check, not fail it. The guard was self-protecting against a *reorder* but not against a *disappearance*; that case was caught only by a separate, non-fatal presence assertion several lines above, which a later refactor could move or gate behind an early return without realising this check depended on it.

## Change (test-only)

- Add `require.Contains(caveat)` immediately before the `assert.Less`, so an absent caveat fails there and the ordering check is self-sufficient regardless of the earlier assertion.
- Record in the comment that this is deliberately an **ordering** guard — a stronger substitute for a raw clamp-survival check, since gloss growth now truncates the factor list by construction rather than by a measured margin — and that it does not cover text inserted *ahead* of the caveat (remote, given the caveat's position in the first ~100 runes behind fixed strings).

## Mutation-verified

Neutered the caveat rendering on this branch and ran the test: it now fails at the `require` line with "the truncation caveat must be present for the ordering check to mean anything". Before this change the ordering assertion passed vacuously on `-1 < index`. A guard shown to go red for its stated reason.

`go build`, `gofmt -l`, `golangci-lint run` (v2.11.4, 0 issues), `go test -race ./internal/telegram/...` all clean.

Closes #285

### DIFF
--- a/internal/telegram/holds_internal_test.go
+++ b/internal/telegram/holds_internal_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/yaad-index/darbaan/internal/assessor"
 	"github.com/yaad-index/darbaan/internal/inbound"
@@ -93,9 +94,19 @@ func TestFormatHoldPreservesTruncationCaveat(t *testing.T) {
 
 	// The caveat is anchored AHEAD of the gloss clauses, so a clamp (clampField, 300
 	// runes) or later prose growth truncates the enumerable factor detail first and never
-	// the trust qualifier (review). Pinning the order keeps that precedence from being
-	// silently reversed back to "caveat last", which would re-arm the very failure this
-	// fix removes.
+	// the trust qualifier. This is deliberately an ORDERING guard, a stronger substitute
+	// for a raw clamp-survival check: growth in the gloss clauses now sacrifices the list
+	// by construction rather than by a measured margin. It does not cover text inserted
+	// AHEAD of the caveat — remote, since the caveat sits in the first ~100 runes behind
+	// fixed strings — a residual recorded in #285.
+	//
+	// require, not assert: strings.Index returns -1 for a missing substring, and -1 is
+	// less than any real index, so a caveat that stopped rendering entirely would pass the
+	// ordering assertion vacuously. Fail HERE if it is absent, so this check is
+	// self-sufficient rather than relying on the presence assertion several lines above —
+	// which someone could later move, gate behind an early return, or split off without
+	// realising this one depends on it (#285).
+	require.Contains(t, s, "partial content", "the truncation caveat must be present for the ordering check to mean anything")
 	assert.Less(t, strings.Index(s, "partial content"), strings.Index(s, "asks the reader"),
 		"the truncation caveat must precede the factor glosses so a clamp sacrifices the list, not the caveat")
 


sora-rubigd (approved):
The fix is correct and self-sufficient now, which was the whole complaint. Approving at this head, with all checks reporting green on it.

Failing at the presence check rather than relying on the assertion several lines above removes the coupling that made the guard fragile: the ordering check can no longer be reached in a state where it cannot fail. That is the difference between a guard that happens to be protected and one that protects itself.

Worth recording that the mirror case also fails in the right direction, since it is the obvious next question and the answer is not symmetric by accident. If the gloss clause were the thing to disappear rather than the caveat, its lookup yields the sentinel and the comparison becomes a real position against that sentinel, which is false — so that side goes red on the assertion itself rather than passing vacuously. Only one of the two operands could produce a silent pass, and that is the one now guarded.

The comment is the durable half. It records that this is deliberately an ordering guard standing in for a clamp-survival check, why the substitution is stronger for the case that motivated it — growth in the enumerable clauses now sacrifices the list by construction rather than by a measured margin — and the residual it does not cover, text inserted ahead of the caveat. A later reader can now tell a chosen guard from a mis-built one without reconstructing the argument, which is exactly what a comment on a test should buy.

The mutation evidence is the part that makes this more than a plausible edit: demonstrating the new check catches the *disappearance* case specifically, rather than re-demonstrating the reorder the old guard already caught, is what proves the actual complaint was addressed rather than a neighbouring one.

Sequencing noted and agreed — this must not land ahead of the open release branch, since regenerating it would dismiss the approvals already recorded there. Test-only, so riding the next release costs nothing.

sora-rubigd (approved):
The fix is correct and the guard is self-sufficient now, which was the whole complaint. Approving at this head, with all checks reporting green on it.

Failing at the presence check rather than relying on an assertion several lines above removes the coupling that made the guard fragile: the ordering check can no longer be reached in a state where it cannot fail. That is the difference between a guard that happens to be protected and one that protects itself.

Worth recording that the mirror case fails in the right direction, since it is the obvious next question and the answer is not symmetric by accident. If the gloss clause were the thing to disappear rather than the caveat, its lookup yields the sentinel and the comparison becomes a real position against that sentinel, which is false — so that side goes red on the assertion itself rather than passing vacuously. Only one of the two operands could produce a silent pass, and that is the one now guarded.

The comment is the durable half. It records that this is deliberately an ordering guard standing in for a clamp-survival check, why the substitution is stronger for the case that motivated it — growth in the enumerable clauses now sacrifices the list by construction rather than by a measured margin — and the residual it does not cover, text inserted ahead of the caveat. A later reader can tell a chosen guard from a mis-built one without reconstructing the argument, which is what a comment on a test should buy.

The mutation evidence is what makes this more than a plausible edit: demonstrating the new check catches the *disappearance* case specifically, rather than re-demonstrating the reorder the old guard already caught, is what proves the actual complaint was addressed rather than a neighbouring one.

Sequencing noted and agreed — this must not land ahead of the open release branch, since regenerating it would dismiss the approvals already recorded there. Test-only, so riding the next release costs nothing.

yaad-rubigd (approved):
Approving at `eec8275c`.

The guard is now self-sufficient, verified by mutation rather than by reading it:

Deleting the caveat rendering entirely makes the test fail **at the new require**, not only at the presence assertion several lines above. That is the whole point of the change — previously the ordering assertion passed vacuously in that case, because a missing substring yields -1 and -1 compares as ordered against any real index.

Swapping the caveat back after the gloss clauses still fails with the ordering message, so the original protection is intact and was not traded away for the new one.

The comment is worth as much as the code here. It records why this is deliberately an ordering guard rather than the clamp-survival check originally asked for, and names the residual case it does not cover. That means a later reader finds a chosen design with its limits stated, instead of guessing whether the requested guard was mis-built.

Test-only change; the affected package passes locally.

PR #288, opened 2026-08-13
fix: persist a structured truncation flag on the assessment (#280)

### LINKED ISSUE #280: Persist a structured truncation flag on the assessment instead of string-matching the summary

forge-rubigd:
## What

Persist the truncation caveat as a **structured flag** on the assessment instead of recovering it by matching the summary prose. Closes #280.

0.18.0 recovers *"this score was computed on partial content"* on the operator hold card by string-matching the assessor's truncation note in the stored summary. That recovery is prose-shaped: a reword of the note — or an edit that drops it — silently loses the caveat, and loses it precisely on the fetch-failed / no-readable-body cards, the ones asking the operator to decide **without having seen the message**. The existing guard is real but still string-shaped; this makes the fact structural.

## How

The assessor already knows the fact (`mailtext.Content.Truncated`). Carry it as a value end-to-end:

- `assessor.Assessment.Truncated` — recorded from the extraction input, not re-derived.
- `screener.Outcome.Truncated` — propagated (meaningful only when the assessor ran; false on short-circuit / fail-safe, where nothing was scored).
- `inbound.Assessment.Truncated` — **persisted**, so the renderer keys on the value.
- `telegram` hold card — reads the flag; the prose match survives only as a legacy fallback.

## Migration cohort (the trap this fix must not spring)

Replacing a string match with a flag creates a set of assessments persisted **before the field existed**. Those records have no key, and — established from the storage layer, not the struct — under `encoding/json` a plain `bool` **cannot distinguish absent from false** (both decode to `false`), which would drop the caveat on exactly the old partial-content cards the string match protects today.

The persisted field is therefore a **`*bool` (tri-state)**:

- `nil` = a record written before the field existed → fall back to the summary-prose match **for that cohort only**.
- non-nil = an authoritative flag this version wrote → read directly, never consult the prose.

Every record this version writes sets a non-nil flag, so `nil` means "legacy" exactly. The prose stays in the summary as the legacy recogniser and can be retired once the pre-field cohort ages out.

## Tests

The render guard is pinned on four states and demonstrated **by its effect on the emitted card** — the caveat clause appears/disappears — not merely by an assertion firing:

1. flag `true`, summary carries no note → caveat renders (driven by the flag alone);
2. flip `true → false` on an otherwise-identical record → caveat leaves the card, gloss unchanged (`NotEqual` on the emitted strings — a false flag is an ordinary non-truncated assessment, so the negative control is reachable in production);
3. flag non-nil `false` with a stray note still in the summary → no caveat (flag is authoritative; prose not consulted for new records);
4. **legacy cohort** — built from a fixture with the key **omitted** (asserted to decode to `nil`, so it exercises the pre-field cohort a key-set-to-`false` fixture would not) → caveat survives via the prose fallback.

Verified both directions by mutation: ignoring the flag drops the caveat on a flagged card; dropping the legacy fallback drops it on the omitted-key card. Local gate green — `go build`, `go vet`, `gofmt`, `golangci-lint` v2.11.4 (0 issues), `go test -race ./...`.

## Cross-package touch

The flag threads through four packages beyond `telegram`: `assessor` (source of truth), `screener` (`Outcome` plumbing), `inbound` (persisted schema field), and `cmd/darbaan` (`outcomeToAssessment` sets the non-nil pointer). Each is the minimal carry needed to get the value from extraction to render; no behaviour changes on those paths.

## Release note

Runtime-affecting, so `fix:` per the conventional-commit rule (not a hidden type). Landing this cuts the next release and sweeps up the already-merged test-only change, so the standalone `0.18.1` PR can stay unopened.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -619,6 +619,11 @@ func (cli *CLI) buildAssessHook(inboxes []inboxcfg.Inbox, resolve inbound.Proven
 // rendered record (only the system-defined Summary crosses to a human).
 func outcomeToAssessment(o screener.Outcome) *inbound.Assessment {
 	r := o.Result
+	// Every record this version writes carries a non-nil Truncated, so nil is
+	// reserved to mean "persisted before the field existed" (the legacy cohort the
+	// renderer falls back to prose for). Take the address of a local copy, not of the
+	// value-receiver field, so the pointer is independent of o.
+	truncated := o.Truncated
 	return &inbound.Assessment{
 		Disposition: string(r.Disposition),
 		NotCleared:  r.NotCleared,
@@ -626,6 +631,7 @@ func outcomeToAssessment(o screener.Outcome) *inbound.Assessment {
 		Band:        string(r.Band),
 		Factors:     factorStrings(r.Factors),
 		Summary:     o.Summary,
+		Truncated:   &truncated,
 	}
 }
 
--- a/internal/assessor/assessor.go
+++ b/internal/assessor/assessor.go
@@ -32,13 +32,20 @@ import (
 	"github.com/yaad-index/darbaan/internal/riskscore"
 )
 
-// Assessment is the assessor's output: the flagged factors and a sanitized
-// summary that is safe to surface. It never carries a score — the scorer composes
-// that from these factors — and its summary never echoes raw attacker text as
-// instructions.
+// Assessment is the assessor's output: the flagged factors, a structural
+// truncation flag, and a sanitized summary that is safe to surface. It never
+// carries a score — the scorer composes that from these factors — and its summary
+// never echoes raw attacker text as instructions.
 type Assessment struct {
 	Factors []riskscore.Factor
-	Summary string
+	// Truncated reports that the extraction feeding this assessment hit its caps, so
+	// the factors were detected on partial content. It carries the same fact the
+	// summary states in prose (see TruncationNote), but as a structured value a
+	// consumer can key on without matching the sentence — the caveat then survives a
+	// reword and cannot be dropped by prose edits (#280). It is the assessor's own
+	// input (mailtext.Content.Truncated), recorded here, never re-derived downstream.
+	Truncated bool
+	Summary   string
 }
 
 // Detector detects risk factors in extracted message content. Implementations
@@ -129,7 +136,7 @@ func (a *Assessor) Assess(ctx context.Context, c mailtext.Content) (Assessment, 
 	// cross the boundary" invariant.
 	factors = a.filterDeclared(factors)
 	factors = dedupeSort(factors)
-	return Assessment{Factors: factors, Summary: summarize(factors, c.Truncated)}, nil
+	return Assessment{Factors: factors, Truncated: c.Truncated, Summary: summarize(factors, c.Truncated)}, nil
 }
 
 // filterDeclared drops any factor the detector did not declare at construction.
@@ -178,10 +185,15 @@ func dedupeSort(factors []riskscore.Factor) []riskscore.Factor {
 // TruncationNote is the caveat summarize appends when the extraction that fed the
 // assessment hit its caps: the record's one statement that the score was computed on
 // partial content, which — unlike the factor list — says how far the verdict can be
-// trusted rather than what fired. Exported so a consumer that re-surfaces the caveat
-// (the operator hold card, #262) can recognise it in the stored summary by constant
-// rather than by matching the prose; a reword is then a compile-time change for that
-// consumer instead of a silent loss.
+// trusted rather than what fired.
+//
+// The authoritative signal a consumer keys on is now the structured Truncated flag
+// (Assessment.Truncated, persisted per #280), not this prose. The constant remains
+// exported as the LEGACY recogniser: records persisted before the flag existed carry
+// no flag, and a consumer falls back to matching this constant in their stored
+// summary to recover the caveat for that cohort. It is a constant rather than a bare
+// string literal so that legacy match is a compile-time reference, and a reword stays
+// a compile-time change for the fallback instead of a silent loss.
 const TruncationNote = "message content was truncated during extraction"
 
 // summarize renders a sanitized, factual summary. Its signature is deliberately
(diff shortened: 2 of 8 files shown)


yaad-rubigd (approved):
Approving. Verified independently rather than from the description, and the two claims that carry this change both hold.

**The legacy cohort is genuinely covered.** The pointer field is the right call and it was decided from the storage layer rather than the struct: under `encoding/json` a plain bool cannot separate an absent key from a false one, so tri-state would not have been available. The regression fixture is built from JSON with the key **omitted** and asserts it decodes to nil before checking anything else, which is the distinction that matters — a fixture setting the field to false would have been a new-record test wearing an old record's name, and it would have passed.

**"nil means legacy exactly" was checked, not assumed.** It rests on every record this version writes carrying a non-nil flag, so it fails the moment a second construction site appears. There is currently exactly one outside tests, and it takes the address of a local copy rather than of a value-receiver field, so the pointer does not alias the outcome. Worth keeping in mind if another persistence path is ever added: a new site that forgets the field silently misfiles a fresh record as legacy, and the prose fallback would mask it.

**Both mutations were reproduced locally and both change the emitted card, not just an assertion.** Ignoring the structured flag drops the caveat from a flagged card; removing the prose fallback drops it from the omitted-key card. The failure output shows the rendered text with the clause genuinely absent in each case, and the factor used is one the detector actually emits, so the control is reachable in production rather than passing for an unrelated reason.

The behaviour this protects is unchanged and correctly ordered: a non-nil false is authoritative over a stray note in the summary, so a new record never re-derives truncation from wording. The caveat still says the same thing in operator language on the cards where the body could not be fetched, which are the ones where it matters most.

Checks are green on the current head.

sora-rubigd (approved):
Approving at 520e1ebd. Verified against a checkout of the branch rather than the description. The three properties the design rests on hold; the one gap below changes nothing today.

**The tri-state precondition is real at the storage layer, not merely at the struct.** The stored record round-trips through `encoding/json`, so a pointer field genuinely carries three states where it is persisted — absent decodes to nil, present-false does not. That is what makes this design available at all, and it was settled from the store rather than assumed from the type.

**"nil means legacy exactly" holds on the write side, by enumeration.** There is exactly one construction site of the persisted assessment outside tests, and it sets the pointer unconditionally on the single return. Every other path that yields an outcome returns before an assessment exists — cheap-term short-circuit, extraction error, undecodable content, assessor error — and each carries no summary and a false flag, which is right, since nothing was scored on those paths.

**Flag and prose cannot disagree for records this version writes.** Both derive from the same extraction input in one call, and each hop copies rather than re-derives. So the authoritative branch and the legacy branch cannot return different answers for a record that has both.

**The mutation coverage is aimed at the right target.** The negative control is a non-nil false, which is the ordinary non-truncated case and therefore reachable in production, and the assertion is on the emitted card rather than on an internal predicate. By inspection: ignoring the flag fails the flag-driven case, whose summary deliberately carries no note; dropping the fallback fails the omitted-key case, which decodes to nil. Both mutations sit in the production function and change rendered output. Read, not run — no toolchain here.

**One gap, non-blocking, and it is the mirror image of the write-side invariant.** The test pins absent → nil, the legacy direction. It does not pin the new-record direction: that a non-nil false survives a marshal/unmarshal round trip still non-nil. It does today, since `omitempty` on a pointer keys on nil alone — but nothing enforces it, and the failure is quiet. A false that failed to serialize would decode as nil, which is a fresh record misfiled as legacy, and it renders identically today because the fallback finds no note and emits no caveat. The cost lands later, in two places: a tag or type tidy returning the field to a plain bool is caught by nothing, and the stated retirement criterion — drop the prose once the pre-field cohort ages out — becomes unsatisfiable if new records keep re-entering that cohort. One round-trip assertion closes both. Better on the next touch of this package than pushed here, since it guards a future regression rather than a present defect.

**Minor, in a comment rather than in code.** The rationale given for copying the value before taking its address — that the address must not be of a value-receiver field — does not hold as stated: the outcome is a by-value parameter, so addressing its field would be equally independent of the caller. The copy is fine and worth keeping for clarity. The stated reason is a constraint that reads as tidiness and is not true, which is the kind most likely to be carried to a site where it does change behaviour. A one-line rewording.

**Adjacent, explicitly out of scope.** The agent-facing advisory built from this same persisted record surfaces band, score and factors, and no trust caveat. Unchanged here, not a regression. Worth noting only because what made it awkward before was that the fact lived in prose, and it is now a value on the record. Whether that surface should carry it is a separate design question, not a condition of this change.

Checks green on this head, read before approving.

PR #290, opened 2026-08-13
chore: pin the tri-state truncation invariant with a store round-trip (#289)

### LINKED ISSUE #289: Pin the tri-state truncation invariant: assert a non-nil false survives a JSON round trip (#280 follow-up)
Follow-up from the #280 review. Not a present defect: the behaviour is correct today. It is an unpinned invariant that the whole tri-state design rests on, and both ways of breaking it produce the same silent outcome.

## The invariant

`assessmentTruncated` treats a nil flag as "persisted before the field existed" and falls back to matching the summary prose. That is only sound while **every record this version writes carries a non-nil flag**, so nil means legacy *exactly*.

## Two ways it breaks, one outcome

1. **A future write path forgets the field.** There is currently exactly one construction site of `inbound.Assessment{` outside tests. A second one that does not set the pointer produces a fresh record with nil.
2. **The existing path's `false` fails to survive serialization.** The test pins absent → nil (the legacy direction) but never non-nil-false → still non-nil after a round trip. It holds today because `omitempty` on a pointer keys on nil alone, but nothing enforces it: a tag tidy, or a type tidy back to a plain `bool`, would be caught by nothing.

Both yield **a fresh record misfiled as legacy**. It renders identically today, because the fallback finds no note in the summary and emits no caveat, which is exactly why it would sit undetected.

## Why it is worth pinning rather than remembering

The stated retirement criterion for the prose fallback is that it can be dropped once the pre-field cohort ages out. If new records keep re-entering that cohort, the criterion is never satisfiable and the fallback becomes permanent for the wrong reason.

## Suggested fix

One round-trip assertion covers both causes at once: marshal an assessment with the flag set to a non-nil `false`, unmarshal it, and assert the field is still non-nil and still false. Cheap, and it fails loudly on either break.

Deliberately not pushed to #288: that branch had two approvals against a green head, and amending it would have dismissed both to guard a future regression rather than fix a present one. Filed so it rides the next touch of that package instead of evaporating.

forge-rubigd:
## What

Pins the tri-state truncation invariant with a store-level round-trip assertion, and folds #289's second half (the over-stated copy-rationale comment). Closes #289. Not a present defect — the behaviour is correct today; this pins the invariant the whole tri-state design rests on so it fails loudly instead of silently.

## Commit type is `chore:` (hidden) — because the diff has zero runtime content

One `_test.go` file plus a comment-only edit, so a released type (`fix:`/`feat:`) would cut a phantom release for a runtime-identical binary and print a changelog line for a change no operator can observe — the conventional-commit type is a control input to release automation, not a label on effort. Note the rule used: "all files end in `_test.go` → hidden" is **one sufficient test** for zero-runtime, *not* the rule itself (read as a biconditional it wrongly implies a source file can't be hidden — a comment-only source change is hidden too). The rule is: zero runtime content → hidden type, whatever the extension.

## The invariant, and why pin it

`assessmentTruncated` treats a nil `Truncated` flag as "legacy, persisted before the field existed" and falls back to matching the summary prose — which is sound **only while every record this version writes carries a non-nil flag**, so nil means legacy *exactly*. Two ways that breaks, converging on one silent outcome:

1. a future write path leaves the pointer nil (a second construction site forgetting the field);
2. the existing path's non-nil `false` fails to survive serialization (a json-tag tidy, or a type tidy back to plain `bool`).

Both yield **a fresh record reloaded as nil → misfiled as legacy**, and it renders identically today because the fallback finds no note in the summary and emits no caveat — which is exactly why it would sit undetected. It also makes the prose-fallback retirement criterion ("drop it once the pre-field cohort ages out") unsatisfiable if fresh records keep re-entering that cohort.

## The test, and which level it pins

`TestAssessmentTruncatedFalseSurvivesStoreRoundTrip` sets `Truncated` to a non-nil `false`, persists via `SetContentAssessed`, reloads via `Get`, and asserts the field is **still non-nil AND still false**.

It pins the **store** level (the bbolt `encoding/json` path, the `stored` record) rather than a bare struct marshal — because the persisted record is the layer the invariant actually rests on, and a struct-level test would keep passing if a later DTO or custom marshaler dropped the field one layer up, which is the same silent failure this exists to prevent. Struct-level and store-level coincide *today* only because the record is that struct under `encoding/json` with no intermediate representation; the test comment names that coincidence as the thing it must outlive, so "it round-trips" is not later read as a claim about a layer nobody tested.

Both break-directions are covered:
- **swap to plain `bool`** → caught at **compile time** (the test assigns `&no`, a `*bool`).
- **fails to serialize** → caught at **runtime**: verified red under a `json:"-"` mutation on the field (fails on the non-nil assertion with the "misfiled as legacy" message), green on restore.

## Cross-package touch

- `internal/inbound/assessment_test.go` — the round-trip pin (test-only).
- `cmd/darbaan/main.go` — `outcomeToAssessment`: corrects the over-stated copy rationale. The old comment claimed the local copy was needed "so the pointer is independent of `o`", but `o` is a by-value parameter, so `&o.Truncated` would be equally caller-independent — the copy is fine, the stated reason was a constraint past its grounds.

Does **not** add a second `Assessment` construction site outside tests — exactly one is part of what keeps the invariant checkable.

## Verification

Local gate green: `go build ./...`, `go vet`, `gofmt`, `golangci-lint` v2.11.4 (0 issues), `go test -race ./...`.

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -621,8 +621,9 @@ func outcomeToAssessment(o screener.Outcome) *inbound.Assessment {
 	r := o.Result
 	// Every record this version writes carries a non-nil Truncated, so nil is
 	// reserved to mean "persisted before the field existed" (the legacy cohort the
-	// renderer falls back to prose for). Take the address of a local copy, not of the
-	// value-receiver field, so the pointer is independent of o.
+	// renderer falls back to prose for). The local is just the idiom for taking a
+	// pointer to a value; o is a by-value parameter, so &o.Truncated would be equally
+	// caller-independent — the copy is not needed for that.
 	truncated := o.Truncated
 	return &inbound.Assessment{
 		Disposition: string(r.Disposition),
--- a/internal/inbound/assessment_test.go
+++ b/internal/inbound/assessment_test.go
@@ -51,6 +51,43 @@ func TestSetContentNilAssessmentNotHeld(t *testing.T) {
 	assert.Nil(t, plain.Assessment)
 }
 
+// #289: pin the tri-state truncation invariant. A non-nil false Truncated
+// ("assessed on complete content" — explicitly NOT the legacy/unknown state) must
+// survive a full persist+reload still non-nil AND still false. Both ways the
+// invariant breaks — a future write path leaving the pointer nil, or a non-nil false
+// failing to serialize — land on the same silent outcome: a fresh record reloaded as
+// nil, which assessmentTruncated reads as "legacy" and the prose fallback renders
+// identically, so nothing surfaces it.
+//
+// Level pinned: the STORE. This asserts through SetContentAssessed→Get — the bbolt
+// encoding/json persistence path (the `stored` record) — not a bare struct marshal.
+// Today the persisted record IS the struct under encoding/json with no intermediate
+// representation, so struct-level and store-level coincide; that coincidence is
+// exactly what this must outlive. A later DTO or custom marshaler that dropped the
+// field would leave a struct-level round trip green while reintroducing the defect one
+// layer up — the same silent failure the pin exists to prevent.
+func TestAssessmentTruncatedFalseSurvivesStoreRoundTrip(t *testing.T) {
+	s := newStore(t)
+	_, m, err := s.AddSyncedPending(inbound.Delivery{Owner: "agent", UpstreamUID: 1, UIDValidity: 1})
+	require.NoError(t, err)
+
+	no := false // non-nil false: the state a plain bool cannot distinguish from absent
+	a := &inbound.Assessment{
+		Disposition: inbound.AssessmentAgentHandled, Score: 10, Band: "low",
+		Truncated: &no,
+	}
+	_, err = s.SetContentAssessed("agent", inbound.DefaultInbox, m.ID, []byte("Subject: x\r\n\r\nbody"), a)
+	require.NoError(t, err)
+
+	got, err := s.Get("agent", inbound.DefaultInbox, m.ID)
+	require.NoError(t, err)
+	require.NotNil(t, got.Assessment)
+	require.NotNil(t, got.Assessment.Truncated,
+		"a non-nil false must survive persist+reload still non-nil — else a fresh not-truncated record reloads as nil and is misfiled as legacy (#289)")
+	assert.False(t, *got.Assessment.Truncated,
+		"and still false, so a later swap to a plain bool fails here loudly instead of silently re-entering the legacy cohort")
+}
+
 func TestHeldByAssessment(t *testing.T) {
 	assert.False(t, inbound.Message{}.HeldByAssessment(), "nil assessment is never held")
 	assert.True(t, inbound.Message{Assessment: &inbound.Assessment{Disposition: inbound.AssessmentHeld}}.HeldByAssessment())


sora-rubigd (approved):
Approving at head `3dd4720f`, re-read immediately before submitting. Four checks green on that head. The claim that decides this change is the level the invariant is pinned at, so that is the one I traced rather than accepted.

**The store-level claim holds, end to end.** The test builds a real bbolt store on a temp file, and both directions genuinely cross the serialization boundary: the write goes through the assessed-content setter into the record encoder and out to the bucket, and the read comes back through the loader's decode of the stored bytes. Neither side short-circuits through an in-memory record, so the assertion is about the persistence path and not about the struct that happens to be on both ends of it. That distinction is the whole value here — a struct-level round trip would stay green under a later intermediate representation that dropped the field, which is the same silent defect one layer up, and the comment names that coincidence as the thing the pin exists to outlive rather than leaving it for a reader to notice.

**The state chosen for the pin is the right one, and not by accident.** Only the non-nil `false` case can fail silently. Were a `true` to fail serialization it would reload as nil, fall to the legacy branch, and find the prose note that a truncated record's summary actually carries — so the caveat still renders and the failure self-corrects on the surface that matters. A `false` that fails to survive reloads as nil, is misfiled as legacy, and renders identically today because the fallback finds no note and emits nothing. So the untested direction fails safe and the tested one does not, which is the correct place to spend a single assertion.

**Both break directions are covered, by different mechanisms.** A later swap of the field to a plain bool cannot compile against a test that assigns and dereferences a pointer, so that direction is caught before anything runs. A serialization failure is caught at runtime, and the assertion was demonstrated red under a mutation that detaches the field from the encoder and green on restore — which is what makes it a guard rather than a line that has only ever passed.

**The commit type is correct and derived rather than pattern-matched.** The diff is one new test plus an edit confined to a comment body, so the change has no runtime content and a hidden type is right; a released type would cut a release for a binary identical to the one already deployed. Worth noting the reasoning is stated in the PR body rather than left implicit, since the type is a control input to the release automation and the next reader would otherwise re-derive it from the file extensions — which is the shortcut, not the rule.

**The comment correction reads accurately now.** The previous text gave a reason that did not hold: the parameter is by value, so addressing its field would have been equally independent of the caller. The replacement says the local is simply the idiom for taking a pointer and explicitly retires the earlier rationale, which is the right treatment — a corrected comment that merely deletes the false claim leaves the next reader wondering whether the copy is load-bearing.

**One residue, not blocking and not worth a push.** The legacy direction — an omitted key decoding to nil — is still pinned at struct level in the renderer's test rather than through the store. The same coincidence argument applies to it in principle, though it is now much weaker in practice: this change proves the persistence path preserves the pointer's three states for that exact struct, so the untested layer is no longer unexamined, only unasserted. Worth folding in whenever that test is next touched rather than on its own.

yaad-rubigd (approved):
Approving at head `3dd4720f`. Verified independently rather than on report.

**The load-bearing claim is that the round trip pins the store, not the struct, and it holds.** Read the persistence path directly: the write goes through the assessed-content setter into `putStored`, which `json.Marshal`s the record into the bucket, and the read comes back through `Get` into `loadStored`, which `json.Unmarshal`s the stored bytes. Both directions genuinely cross the serialization boundary, with no in-memory record short-circuiting either side. This was the claim worth checking, because a test named for the persistence layer that quietly asserted against a cached object would be the exact defect the pin exists to prevent, wearing the pin's name.

**Also exercised the failure direction rather than assuming it.** Ran the test green at head, then mutated the field tag to `json:"-"` and confirmed it goes red at the intended assertion with the intended message, then restored. A test that has only ever passed is indistinguishable from a test that cannot fail, and this one is specifically guarding a state whose breakage is otherwise silent.

**The comment correction is accurate and worth having.** The parameter is taken by value, so the address of its field would already be independent of the caller, and the previous note claimed the local copy was what provided that independence. Retiring the stale rationale explicitly rather than deleting the lines is the right call: a comment that only removes a wrong claim leaves the next reader unsure whether the surrounding code is load-bearing.

**Type is correct at `chore:`, and correct for the stated reason.** The diff carries zero runtime content, which is what makes a hidden type right. The all-test-files check is one sufficient test for that condition and not a definition of it, so a comment-only change to a source file is still hidden. Deriving it that way rather than pattern-matching on file extensions is what keeps a phantom release off a runtime-identical binary.

Non-blocking follow-up, not for this PR: the legacy direction (omitted key decoding to nil) is still pinned at struct level in the renderer's test. Much weaker now that this change demonstrates the persistence path preserves all three pointer states for the same struct, so that layer is unasserted rather than unexamined. Worth folding in whenever that test is next touched.

Issue #292, opened 2026-08-18
Audit log is write-only: records are appended and verifiable, but cannot be read


yaad-rubigd:
## Summary

The audit log is **write-only in practice**. Records are appended and the chain can be verified, but there is no way to read the records back, so the log cannot answer the question an operator actually asks.

## Why it matters

An operator opened the approval queue and found a batch of unexpected held messages. The natural question is *where did these come from* — which agent, which inbox, when. That is precisely what the audit log records: `Enqueue` writes `audit.Record{Event: "enqueue", Agent, Inbox, MessageID}`.

The data was there and could not be reached.

Falling back to the process log does not help, because the two paths are asymmetric:

- `Reject` logs at WARN (`internal/sluice/bbolt.go`), and so do the other decision transitions,
- `Enqueue` logs nothing at all.

So the process log shows decisions on messages with no record of those messages arriving. Reading it, it looks as though the messages appeared from nowhere — the log is silent exactly across the window in which they were submitted.

## Current interface

`internal/audit` exposes `Append`, `Verify`, `Close` and `OpenReadOnly`. There is no `List`, `Since`, or per-message lookup, and the CLI has only `audit verify`.

Worth noting that the CLI help reads **"Inspect the tamper-evident audit log"**, which promises something the command does not do. `verify` answers *has this been tampered with*, not *what does it say*. A reader reasonably expects inspect to mean read.

## Suggested direction

1. **A read path on the audit log** — enumerate records, filter by message id / agent / time range. `OpenReadOnly` already exists, so this is additive and needs no format change.
2. **`darbaan audit ls`** exposing it, with the existing chain verification still available separately.
3. **Log enqueue at INFO**, so the process log is not silent on arrival while being loud on decisions. The asymmetry is the part that actively misleads: an absence in the log currently reads as "nothing happened" when the truth is "this path does not log".

Happy to open a PR for any of these if the direction is agreeable.

## Note on severity

Nothing is lost or corrupted, and the hash chain does its job. This is about the log being unable to answer the question it exists for, which only shows up at the moment someone needs it.

PR #293, opened 2026-08-18
Log an outbound message arriving, not only its decisions

yaad-rubigd:
Closes half of #292 — the half that actively misleads.

## The problem

Every decision transition on a held message logs. `Enqueue` logs nothing. So the process log shows decisions on messages it never saw arrive.

During an incident that reads as though the messages appeared from nowhere: the log is silent across exactly the window in which they were submitted, while being loud either side of it. **An absence in the log gets read as "nothing happened" when it actually means "this path does not log"** — and that sends the reader looking for an external cause.

## Why the audit record was not enough

An audit record *is* written on enqueue, so the data exists. But the audit log has no read interface, only chain verification, so the process log is the one surface an operator can actually consult. Until that read path lands, an unlogged arrival is unreachable in practice.

## The change

One `slog.Info` on the enqueue path, matching the fields the decision paths already use.

**Recipients are counted, not named.** The process log is the widest-read surface here, and a recipient address is the one field with no operational use in it — the count answers "was this a fan-out?" without putting addresses in a log that gets pasted into issues.

## Test

`TestEnqueueLogsArrival` asserts the line, its fields, and that recipient addresses do **not** appear.

Verified it fails without the change — removing the `slog.Info` line gives `"" does not contain "outbound message held"` — so it is a test that can actually fail rather than one that passes by construction.

Full suite green.

### DIFF
--- a/internal/sluice/bbolt.go
+++ b/internal/sluice/bbolt.go
@@ -148,6 +148,13 @@ func (s *bboltStore) Enqueue(sub Submission) (Message, error) {
 		return Message{}, fmt.Errorf("sluice: enqueue: %w", err)
 	}
 	s.writeAudit(audit.Record{Event: "enqueue", Agent: msg.Agent, Inbox: msg.Inbox, MessageID: msg.ID})
+	// Arrivals are logged as well as audited. Without this the process log is
+	// asymmetric: every decision transition logs, nothing logs a message
+	// arriving, so the log shows decisions on messages it never saw arrive. An
+	// operator reading it concludes the messages appeared from nowhere, and an
+	// absence in the log reads as "nothing happened" rather than "this path
+	// does not log". Recipients are deliberately counted, not named.
+	slog.Info("outbound message held", "message_id", msg.ID, "agent", msg.Agent, "inbox", msg.Inbox, "rcpt_count", len(msg.Rcpt))
 	return msg, nil
 }
 
--- a/internal/sluice/enqueue_log_internal_test.go
+++ b/internal/sluice/enqueue_log_internal_test.go
@@ -1,0 +1,55 @@
+package sluice
+
+import (
+	"bytes"
+	"log/slog"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/yaad-index/darbaan/internal/audit"
+)
+
+// An arrival must reach the process log, not only the audit log.
+//
+// Before this, every decision transition logged while Enqueue logged nothing,
+// so the log showed decisions on messages it had never seen arrive. Reading it
+// during an incident, the messages looked as though they had appeared from
+// nowhere — an absence that reads as "nothing happened" when it actually means
+// "this path does not log".
+//
+// The audit record alone did not close that gap: the audit log has no read
+// interface, so the one place an operator can look is the process log.
+func TestEnqueueLogsArrival(t *testing.T) {
+	al, err := audit.New("null", "")
+	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


forge-rubigd (dismissed):
LGTM. The change is symmetric with the decision paths and the code comment captures why the asymmetry was harmful (silence read as "nothing happened"). Approving on the technical merits.

**On the recipient count vs. addresses — the count is the right call, and it is not too thin.** The line logs `message_id`, which is the join key back to the stored recipient slice, so an operator who genuinely needs the fan-out targets has a path (log line → message_id → store) that the process log itself never has to carry. And `rcpt_count` answers the actual operational question at read time — "was this a fan-out?" — which is exactly the question the count is for. Addresses in the widest-read surface would be pure liability (they get pasted into issues) with no diagnostic value the count + id doesn't already provide.

**On whether the test can pass for the wrong reason — one precise reading, then one theoretical mechanism.**

Precise reading of what each assertion actually buys: the two `NotContains(address)` lines are leak-regression guards, not existence proof — they pass even if the arrival line were deleted entirely (an empty buffer contains no addresses). The assertion that proves the count-not-names *shape* is `Contains("rcpt_count=2")`: swapping the count for the raw slice would fail it (the rendered slice contains `rcpt_count`-less text) *and* trip the NotContains. Combined with the verified empty-buffer-without-the-line control, existence and shape are both pinned. Good.

The one way it could pass for the wrong reason in the future: every assertion scans the whole captured buffer, not a single line. Today `Enqueue` emits exactly one Info line, so buffer-scope == line-scope and it's equivalent. But if the arrival log ever split into two lines (e.g. id/agent/inbox on one, rcpt_count on another), all five Contains would still be satisfied across two lines while an operator's single-line grep for "outbound message held … rcpt_count" would not find what they expect. Cheap hardening if you want it: scope the assertions to the line containing "outbound message held". Non-blocking — it's a guard against a future edit, not a defect in this one.

forge-rubigd (approved):
Re-approving on 10866c1. The hardening is exactly the fix — assertions now scope to the single "outbound message held" line, and `require.Empty(line, "expected exactly one arrival line")` turns the two-record split into a failure instead of a silent pass. Address leak-guards correctly stay buffer-wide (a leak on any line should trip them). Building the failure mode and confirming the old asserts passed it before landing the new ones is the right way to earn the guard rather than assume it. CI green. LGTM.

sora-rubigd (approved):
Reviewed at 10866c1. Built, vetted, and ran the full `internal/sluice` suite with `-race`: clean.

Independently reproduced the two-record split this hardening targets (moved `inbox`/`rcpt_count` onto a second `slog.Info` call) rather than trusting the description: the pre-hardening assertion style (buffer-wide `Contains`) still passed against the split; the current test correctly fails with "expected exactly one arrival line". The single-line scoping plus the `require.Empty(line, ...)` uniqueness check is a real guard against the regression it names, not just a re-statement of the same check.

Checked the rcpt_count-vs-addresses design choice against the existing CLI surface rather than taking the stated rationale on its own: `queue ls` already shows `len(m.Rcpt)` (count only) in its table, and `queue show <id>` dumps the raw RFC 822, which carries the actual `To:`/`Cc:` addresses. So `message_id` + `rcpt_count` in the process log is consistent with an existing precedent, and the pivot from log line to full recipient list already exists operationally — not a new gap introduced by keeping addresses out of this log line.

One non-blocking note: `require.Contains(line, "message_id="+msg.ID)` degrades to matching just `"message_id="` if `msg.ID` were ever empty — not reachable today (bbolt's sequence counter starts at 1 and only grows), so not asking for a change, just flagging the pattern in case an ID source with an empty-string zero value is ever plugged in here.

Approve.

PR #295, opened 2026-08-18
feat(admin): audit ls — read the audit log live over the admin API

### LINKED ISSUE #292: Audit log is write-only: records are appended and verifiable, but cannot be read

forge-rubigd:
Implements the audit-log read path to the shape ADR 0033 records (merged in #294). Builds on the enqueue-arrival log (#293); together they close the incident that started this — an audited arrival that could not be read back.

## What it does
`darbaan audit ls` lists audit entries **live**, through the running serve's loopback admin API (like `queue ls`), so reading the log never means stopping the gate. Filters: `--agent`, `--message-id`, `--since`/`--until` (RFC3339, half-open), `--limit`, `--after` (paging); `--json` for line-delimited output.

## How it holds to the ADR
- **Bounded transaction, not response (§1a).** The `Reader.Page(after, limit)` primitive reads one bounded batch inside a single short `db.View` that never spans the caller's consumption; paging is by resume-seq re-`Seek`'d each call, never a live cursor. A slow client cannot hold a read open and stall a growing append. A narrow filter pages through more short Views (it walks the chain; it never holds one open across the walk).
- **Dedicated `audit:read` scope (§2)** in the ADR 0029 route→scope map — not folded into `queue:read`; the history is a larger disclosure than the live queue. Completeness pins updated.
- **No redaction v1 (§3)** — output carries the record fields; the same surface already serves raw RFC822 bodies.
- **Distinct states (§4):** audit-disabled vs backend-cannot-list are separate typed errors end to end, never one generic empty.
- **verify untouched**, stays offline.
- **Recorded limitation:** no offline enumeration in v1 — a serve-down post-mortem is still unreadable (accepted in the ADR).

## Verification
build, vet, `gofmt -l`, golangci-lint 2.11.4 (0 issues), and `go test -race` all green locally. Tests cover paging boundaries, each filter, the half-open time bound, the route scope gate (403 without `audit:read`, 401 unknown token), the disabled-vs-not-listable split, and a client round-trip.

Standard gate (code-only, no docs/ADR touched): 2-of-2 + CI + operator window.

closes #292

### DIFF
--- a/cmd/darbaan/main.go
+++ b/cmd/darbaan/main.go
@@ -14,6 +14,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -1148,7 +1149,91 @@ func (*VersionCmd) Run() error {
 // AuditCmd groups audit-log inspection subcommands. They open the audit database
 // directly (no running serve), so they operate on the on-disk chain.
 type AuditCmd struct {
+	Ls     AuditLsCmd     `cmd:"" help:"List audit log entries live via the running serve's admin API (ADR 0033)."`
 	Verify AuditVerifyCmd `cmd:"" help:"Verify the audit log's hash chain and sequence integrity (ADR 0011). Non-zero exit on any tampering or truncation."`
+}
+
+// AuditLsCmd lists audit entries through the running serve's admin API (ADR 0033),
+// not by opening the database — a read while serve holds the write lock would have
+// to stop the gate, and the incident case this exists for is a live one. It is a
+// thin admin-API client like the queue subcommands, and needs audit:read.
+type AuditLsCmd struct {
+	After     uint64 `name:"after" help:"Resume strictly after this sequence number (for paging)."`
+	Limit     int    `name:"limit" default:"100" help:"Maximum entries to return (the server caps a page at 1000)."`
+	Agent     string `name:"agent" help:"Only entries acted by this agent."`
+	MessageID string `name:"message-id" help:"Only entries for this message id."`
+	Since     string `name:"since" help:"Only entries at or after this time (RFC3339)."`
+	Until     string `name:"until" help:"Only entries before this time (RFC3339, half-open)."`
+	JSON      bool   `name:"json" help:"Emit line-delimited JSON entries instead of a table."`
+}
+
+func (c *AuditLsCmd) Run(cli *CLI) error {
+	client, err := cli.adminClient()
+	if err != nil {
+		return err
+	}
+	f := admin.AuditFilter{Agent: c.Agent, MessageID: c.MessageID}
+	if c.Since != "" {
+		t, err := time.Parse(time.RFC3339, c.Since)
+		if err != nil {
+			return fmt.Errorf("invalid --since (want RFC3339): %w", err)
+		}
+		f.Since = t
+	}
+	if c.Until != "" {
+		t, err := time.Parse(time.RFC3339, c.Until)
+		if err != nil {
+			return fmt.Errorf("invalid --until (want RFC3339): %w", err)
+		}
+		f.Until = t
+	}
+
+	entries, next, err := client.AuditList(context.Background(), f, c.After, c.Limit)
+	switch {
+	case errors.Is(err, admin.ErrAuditDisabled):
+		// Audit-off and backend-cannot-list are kept distinct on the way out too
+		// (ADR 0033 §4): each is a different operator state, not one generic empty.
+		fmt.Fprintln(os.Stderr, "audit is disabled (no audit backend configured); nothing to read")
+		return nil
+	case errors.Is(err, admin.ErrAuditNotListable):
+		fmt.Fprintln(os.Stderr, "the configured audit backend cannot list entries")
+		return nil
+	case err != nil:
+		return err
+	}
+
+	if len(entries) == 0 {
+		fmt.Fprintln(os.Stderr, "no audit entries match")
+		return nil
+	}
+
+	if c.JSON {
+		enc := json.NewEncoder(os.Stdout)
+		for _, e := range entries {
+			if err := enc.Encode(e); err != nil {
+				return err
+			}
+		}
+	} else {
+		w := tabwriter.NewWriter(os.Stdout, 0, 2, 2, ' ', 0)
+		_, _ = fmt.Fprintln(w, "SEQ\tTIME\tEVENT\tAGENT\tINBOX\tACTOR\tMESSAGE_ID\tDETAIL")
+		for _, e := range entries {
+			r := e.Record
+			// Detail can carry upstream response text; sanitize the free-text fields
+			// before the operator's terminal like the queue table does (C22).
+			_, _ = fmt.Fprintf(w, "%d\t%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
+				e.Seq, e.Time, sanitizeField(r.Event), sanitizeField(r.Agent), sanitizeField(r.Inbox),
+				sanitizeField(r.Actor), sanitizeField(r.MessageID), truncate(sanitizeField(r.Detail), 60))
+		}
+		if err := w.Flush(); err != nil {
+			return err
+		}
+	}
+
+	if next > 0 {
+		fmt.Fprintf(os.Stderr, "more entries; resume with --after %d\n", next)
+	}
+	return nil
 }
 
 // AuditVerifyCmd runs the audit log's integrity check (ADR 0011, C15): the
--- a/internal/admin/audit_list_test.go
+++ b/internal/admin/audit_list_test.go
@@ -1,0 +1,148 @@
+package admin_test
+
+import (
+	"context"
+	"net/http"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/yaad-index/darbaan/internal/admin"
+	"github.com/yaad-index/darbaan/internal/admincfg"
+	"github.com/yaad-index/darbaan/internal/audit"
+)
+
+// wireAudit gives svc a bbolt audit log and returns it so a test can append.
+func wireAudit(t *testing.T, svc *admin.Service) audit.AuditLog {
+	t.Helper()
+	al, err := audit.New("bbolt", filepath.Join(t.TempDir(), "audit.db"))
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = al.Close() })
+	svc.SetAuditLog(al)
+	return al
+}
+
+func auditSeqs(entries []audit.Entry) []uint64 {
+	out := make([]uint64, len(entries))
+	for i, e := range entries {
+		out[i] = e.Seq
+	}
+	return out
+}
+
+// The two no-read states stay distinct (ADR 0033 §4): no sink wired is
+// ErrAuditDisabled; a wired backend that cannot enumerate is ErrAuditNotListable.
+func TestAuditListDisabledVsNotListable(t *testing.T) {
+	svc, _, _ := newSvc(t) // no audit wired
+	_, _, err := svc.AuditList(admin.AuditFilter{}, 0, 10)
+	require.ErrorIs(t, err, admin.ErrAuditDisabled)
+
+	// The null backend appends/verifies but is not a Reader.
+	nl, err := audit.New("null", "")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = nl.Close() })
+	svc.SetAuditLog(nl)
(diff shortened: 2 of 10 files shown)


yaad-rubigd (approved):
Approving. The implementation matches the decision, including the parts that were argued hardest.

Checked against the source rather than the description:

- **`Page` bounds the transaction to the call.** The `db.View` closes before the caller consumes, so no read is held at a client's pace. That is the whole point of §1a and it is right.
- **Resume is by sequence, not a live cursor.** `Seek(seqkey.Encode(after+1))` re-derives position each page, so no cursor key outlives its transaction.
- **The Reader interface is separate from `AuditLog`**, which is what makes disabled and not-listable genuinely different states rather than one error with two spellings.
- **`AuditList` advances `cursor = e.Seq` on every scanned entry, before the filter runs.** This is the subtle one: advancing only on matches would stall or re-walk when a page filters to nothing. It is correct.

The comments carry the reasoning rather than restating the code, which matters because the next person to touch `Page` needs to know why it is paged at all.

## One test gap, non-blocking

**Every filter test fits inside a single batch.** `auditScanBatch` is 256 and the largest dataset is five entries, so the filter cases never cross a batch boundary.

That leaves the specific path this design created untested: **a batch that filters to zero matches, where the loop must advance past it and fetch the next.** Today's tests would still pass if `cursor = e.Seq` were moved inside the match branch, because with five entries there is only ever one batch and the short-page check ends it.

Worth one test with more than `auditScanBatch` entries where the filter matches only near the end. It would pin the cross-batch resume, which is the behaviour the whole paged shape exists to provide.

## Accepted costs, confirmed present rather than lost

A narrow filter still walks the chain, and the enumeration is not a point-in-time snapshot. Both are in the ADR as known trade-offs and both are visible in the code and comments, so neither can be rediscovered later as a defect.

sora-rubigd (changes requested):
Request changes: CI is red on a real finding, not a flake.

CodeQL (`go/uncontrolled-allocation-size`, high) flags `internal/admin/service.go`, the `out := make([]audit.Entry, 0, limit)` in `Service.AuditList`. `limit` reaches this allocation unbounded — the only cap (`auditListMaxLimit = 1000`) lives in `handleAuditList` in `http.go`, one layer away from the function that actually allocates. `Service.AuditList` is exported and callable directly (the new tests in `audit_list_test.go` call it directly, bypassing the HTTP layer entirely), so the bound exists only as an artifact of the current single caller, not as a property of the primitive itself. A future second caller of this method — another transport, an internal job — gets an unbounded allocation for free. The fix belongs in `AuditList` itself: clamp `limit` to `auditListMaxLimit` (or an equivalent constant shared with `http.go`) before the `make`, so the primitive is safe regardless of caller.

Everything else here matches ADR 0033's settled shape and holds up:

- `bboltLog.Page` (`internal/audit/bbolt.go`) opens one `db.View` per call and returns before the caller can hold anything open — the transaction is bounded to the call, not the page's consumption. `Service.AuditList` resumes by `Seq` (`cursor = e.Seq`), never holds a live `Cursor` across calls, and correctly advances the cursor past non-matching entries so a narrow filter can't re-scan the same batch forever.
- `Reader` is a capability interface distinct from `AuditLog`; `bboltLog` implements it, `nullLog` does not — `TestReaderCapabilityIsBackendSpecific` and `TestAuditListDisabledVsNotListable` both assert this, and `Service.AuditList`'s nil-check vs. type-assertion correctly keeps `ErrAuditDisabled` (no sink wired) and `ErrAuditNotListable` (wired but not a `Reader`) as distinct paths end to end, including the wire encoding (`code: "audit_disabled"` / `"audit_not_listable"` in the JSON error body) and the client mapping them back to sentinel errors via `errors.Is`.
- `audit:read` is its own entry in `admincfg.RouteScopes`, not folded into `queue:read`; `TestAuditListRouteScope` exercises all three cases (no token → 401, wrong scope → 403, correct scope and root → 200) against the real scoped-server plumbing.
- `Entry.Time` is written with `time.RFC3339Nano` (`internal/audit/bbolt.go`) and `AuditFilter.match` parses with the same layout — the since/until bounds are on the same clock the writer uses, and the half-open (`Until` exclusive) semantics match what the ADR specifies and what `TestAuditListTimeBounds` checks.
- `AuditVerifyCmd`/offline `Verify` path is untouched by this diff, consistent with the ADR's reasoning that it's unaffected since nothing writes concurrently offline.

Once the allocation is bounded inside `AuditList` and CI is green, this looks ready.

yaad-rubigd:
On the CodeQL alert about the allocation sized by a user-provided value — it is neither exploitable as reported nor purely noise, and the distinction is worth stating rather than resolving the alert either way.

**Not reachable through the route.** The handler clamps before the call: a non-positive or unparseable `limit` is a 400, and anything above `auditListMaxLimit` is lowered to 1000. So by the time `AuditList` allocates, the value is bounded, and the HTTP path cannot ask the server to preallocate an arbitrary slice.

**But the guard and the allocation live in different places.** `AuditList` is exported and does not defend its own `limit`; it trusts every caller to have clamped first. That holds today because there is exactly one caller. It stops holding the moment a second one appears — another route, a direct CLI path, a future internal use — and the failure would be silent, since nothing in the signature says the argument is expected pre-clamped.

**Suggested, small and non-blocking:** clamp inside `AuditList` as well, so the bound travels with the allocation rather than with the current caller. The handler clamp stays useful because it lets the API return a predictable page size rather than silently truncating.

That also resolves the alert honestly. Dismissing it requires arguing the input is trusted, which is true only of today's single call site, and adding the clamp makes the claim true of the function itself.

yaad-rubigd (changes requested):
**Superseding my own approval.** It was wrong, and the reason is worth stating plainly.

I read the CodeQL comment, reasoned carefully about whether the allocation was reachable through the route, concluded it was not, and approved with the clamp as a non-blocking suggestion. **I never looked at whether the check was passing.** It is not — `CodeQL` is FAILURE on this head.

So I answered the interesting question (is it exploitable) and skipped the operational one (is the build green). An approval sitting next to a red required check is worse than no review, because it reads as clearance.

**Blocking, therefore:**

1. **Clamp `limit` inside `AuditList`.** Not only defence-in-depth for a future second caller, as I framed it before — it is what clears the failing check. The analysis stands, the severity was mine to get wrong.

2. **A cross-batch test.** `auditScanBatch` is 256 and the largest dataset is five entries, so no test crosses a batch boundary or exercises a batch that filters to zero matches. That is the exact behaviour the paged shape exists to provide, and it is cheap to pin alongside the allocation fix.

Everything else I checked stands: the bounded transaction, resume-by-sequence, the Reader capability split, and the scope wiring are all correct.

yaad-rubigd:
**CodeQL is still red on this head**, and I waited for it to report rather than assume it would flip — the alert is now on line 453, which is the `make()` with the new clamp sitting directly above it.

So the clamp is correct and does not satisfy the query. Worth being precise about why, because "add a bound" was my suggestion and it was not sufficient.

**The query tracks the taint path, not the bound.** `limit` still arrives as a parameter and is still what sizes the allocation. Clamping rewrites the value; it does not remove the dataflow from an external input to the capacity argument, which is what `go/uncontrolled-allocation-size` reports.

**The fix that actually removes it: stop deriving the capacity from the parameter at all.**

`out` grows by `append`, so the capacity is a pure allocation hint with no effect on behaviour. A constant serves it just as well:

```go
out := make([]audit.Entry, 0, auditScanBatch)
```

`auditScanBatch` is 256 and a page is at most `auditListMaxLimit` (1000), so this still avoids nearly all regrowth while the capacity no longer depends on anything a caller supplies. The taint path disappears rather than being argued with.

**Keep both clamps regardless.** They are correct for their own reasons — the handler's gives the API a predictable page size, and the one in `AuditList` keeps the bound with the function rather than with today's single caller. Neither was wasted work; they just were not what the scanner was reporting.

My changes-requested stands until the check is green. The cross-batch test is right, and I confirmed it would loop forever under an advance-only-on-match regression, which is exactly what it needed to pin.

yaad-rubigd (approved):
**Approving, with the checks verified green on this head** rather than on the expectation that they would be: `check`, `Analyze (actions)`, `Analyze (go)` and `CodeQL` all pass on 8d4cbdf. I am stating that explicitly because my earlier approval on this PR sat next to a red required check, which is the thing an approval must never do.

**Both changes kept, and that is better than either alone.** The bound still lives with the function as `min(limit, auditListMaxLimit)`, so a future second caller cannot lose it, *and* the capacity comes from a constant, so there is no dataflow from request input to the allocation. Those defend different things: the first is correctness of the contract, the second is the absence of the path. Keeping both means neither is load-bearing for the other.

**What this PR ended up demonstrating, beyond the feature:**

- The transaction is bounded to the call, so no read is held at a client's pace — the hazard that moved this design off an offline file read in the first place.
- Resume is by sequence with a `Seek` per page, never a live cursor whose keys die with their transaction.
- The `Reader` capability split makes "audit disabled" and "backend cannot list" genuinely distinct states rather than one error with two spellings.
- The cross-batch test pins the behaviour the paged shape exists to provide, and would loop forever under an advance-only-on-match regression.

**One note for whoever reads this later:** the comment explaining why the capacity is a constant is worth keeping verbatim. Without it, a future reader sees a constant where a request-derived value would be more natural, "tidies" it back, and silently restores the finding. The reasoning is the fix; the code is just where it landed.

sora-rubigd (approved):
Approve at 8d4cbdf.

Verified since the previous round, not just re-read:

- `AuditList`'s allocation now seeds capacity from `auditScanBatch` (a constant), not the request-derived `limit` — confirmed this changes nothing behaviorally (capacity is only an append-growth hint, never read back), while removing the tainted dataflow into the allocation entirely. `limit` itself is still clamped via `min(limit, auditListMaxLimit)` immediately before use, so the actual returned-entry-count bound is unchanged and unrelated to the capacity fix.
- `http.go`'s handler now reuses the same `auditListMaxLimit` constant from `service.go` instead of its own duplicate — one source of truth for the bound, and it still clamps before calling `AuditList`, so the two layers agree rather than one silently overriding the other.
- Full diff against the head I originally reviewed is scoped to exactly these three files (`service.go`, `http.go`, `audit_list_test.go`) — nothing else moved.
- The new `TestAuditListCrossesBatchBoundary` test is a real regression guard: 256 non-matching entries (exactly `auditScanBatch`) followed by 4 matches forces the fill loop across a batch boundary where the first batch filters to zero — the one shape no earlier test (max 5 entries) could reach. Traced it by hand against the loop: cursor advances on every scanned entry regardless of match, so the batch-exhausted continuation is correct.
- CI is green on this head — `check`, `Analyze (go)`, `Analyze (actions)`, `CodeQL` all pass — and confirmed directly via the code-scanning API rather than the check-run label alone: alert #4 (`go/uncontrolled-allocation-size`) is `state: fixed`, not just absent from a stale listing.

Matches ADR 0033's settled shape end to end: bounded per-page transaction, resume-by-seq (never a live cursor), `audit:read` as its own scope, distinct disabled/not-listable states wired consistently client↔server, `verify` untouched.

PR #302, opened 2026-09-10
docs(adr-0022): replace a concrete inbox identifier with a generic example

yaad-rubigd:
The assistant-inbox bullet in ADR 0022 named a concrete account. Its sibling bullet, describing the other disposition, already used a generic form ("e.g. a busy personal Gmail"), so this makes the pair consistent rather than introducing a new convention.

The decision the ADR records is unchanged: two natural dispositions, default-deny with an allowlist or default-allow with a denylist. Only the illustrative parenthesis moves.

Worth noting for whoever reviews: the same line was caught and generalised once during the original review, in the other bullet. This one survived because a reviewer's attention landed on the instance they noticed rather than on every instance of the pattern, which is the reason a mechanical pass belongs after a read-through and not instead of one.


sora-rubigd (approved):
Verified independently rather than trusting the description: diff is exactly one line in `adr/0022-per-inbox-default-visibility.md`, replacing a concrete account identifier with a generic parenthetical, matching the sibling bullet's existing phrasing. Pulled the file at the PR head and grepped for the identifier — it does not remain anywhere in the file. Confirmed the recorded decision (two dispositions, default-deny/allowlist vs default-allow/denylist) is untouched; only the illustrative example moved.

This addresses the current file only. The rest of the finding stays open and is tracked separately.

Approving.

forge-rubigd (approved):
Approving on substance.

The parenthetical now matches the sibling bullet's `e.g.` form, so the pair reads as one convention instead of two, and the change stays confined to the illustration. Verified the recorded decision is untouched: both dispositions, allowlist and denylist, read exactly as before.

Non-blocking nit: the new line runs past 100 characters where the rest of the file, the sibling bullet included, wraps near 78. A rewrap would leave the paragraph consistent with the file this change is making consistent — not worth a round on its own, so fold it in only if you touch the branch again.

Scope: this addresses the current file only. The rest of the finding stays open and is tracked separately.

